PR #3, opened 2020-02-05
Add the celery boilerplate

mprahl:


docker-compose.yml
```diff
@@ -0,0 +1,29 @@
...
+  iib-worker:
+    build:
+      context: .
+      dockerfile: ./docker/Dockerfile-workers
+    # Override the default command so that Celery auto-reloads on code changes
+    command:
+    - /bin/bash
+    - -c
+    - >-
+      pip3 install watchdog[watchmedo] &&
```
yashvardhannanavati: shouldn't it just be `pip3 install watchdog`? Am I missing something?

yashvardhannanavati: Ah, got it. nvm :)

mprahl: By doing `watchdog[watchmedo]`, it installs `watchdog` with the extra requirements required to use the `watchmedo` feature.

You can see that this is defined here:
https://github.com/gorakhargosh/watchdog/blob/b752e371792eadf4d1e79cdcc15c40f3539bcb96/setup.cfg#L7-L10

yashvardhannanavati: Ah, I learned something today. thanks!

yashvardhannanavati:
:+1:

PR #4, opened 2020-02-06
Add the database schema with some helper methods

mprahl:
This schema is based on the data that IIB will need to accept and store based on the API design.

This PR also has a separate commit that adds the ability to configure additional loggers.

I tested the migration script on SQLite (manual and in the unit tests) and Postgres 9.6.

Please note that most of the content in `iib/web/migrations` is automatically generated using `FLASK_APP=iib.web.wsgi flask db init`.


.gitignore
```diff
@@ -130,3 +130,6 @@ dmypy.json
 
 # IDE
 .vscode/
+
+# Custom
```
lcarva: Custom what?

iib/web/models.py
```diff
@@ -0,0 +1,219 @@
...
+    """
+    An association table between images and the architectures they were built for.
+
+    This will only be used for images built by IIB.
+    """
+
+    # A primary key is required by SQLAlchemy when using declaritive style tables, so a composite
+    # primary key is used on the two required columns
+    image_id = db.Column(
+        db.Integer, db.ForeignKey('image.id'), autoincrement=False, index=True, primary_key=True
```
lcarva: If both `image_id` and `architecture_id` are primary keys, then won't this enforce a one-to-one relationship between Image and Architecture? 

IIUC, this should be a many-to-many relationship.

lcarva: A unit test to validate this behavior would be nice.

mprahl: The primary key is used to uniquely identify a row. Since this is an association table that maps images to architectures, there would not be more than one row where `image_id` and `architecture_id` are the same.

As per your unit test comment. I think you were asking me to modify the existing unit test so that more than one architecture was added to the image. I updated the test to do this in the latest push.

mprahl:
@lcarva this is ready for another round of review.

iib/web/models.py
lcarva: Thank you for the explanation. The test does help clear things up.

PR #5, opened 2020-02-08
Add auth and `Add` API endpoint

yashvardhannanavati:


yashvardhannanavati:
@mprahl can you PTAL? If this looks good, I'll go ahead and write the tests.

iib/web/models.py
```diff
@@ -190,6 +204,79 @@ def add_state(self, state, state_reason):
         db.session.flush()
         self.request_state_id = request_state.id
 
+    def to_json(self, verbose=True):
+        states = [
+            {
+                'state': RequestStateMapping(state.state).name,
+                'state_reason': state.state_reason,
+                'updated': state.updated.isoformat(),
```
mprahl: One thing we mean to do in Cachito but haven't yet, is to add a "Z" at the end of the timestamp to signify a UTC timestamp.

iib/web/models.py
```diff
@@ -190,6 +204,79 @@ def add_state(self, state, state_reason):
...
+            }
+            for state in self.states
+        ]
+        # Reverse the list since the latest states should be first
+        states = list(reversed(states))
+        latest_state = states[0]
+
+        rv = {
```
mprahl: Could you alphabetize the keys to make them easier to read?

mprahl: This is missing the following keys:
* from_index_resolved
* binary_image_resolved
* index_image
* request_type

Please note that these will be `None` when the request is first created, so you will need to guard against this.

mprahl: In the API document, we said we'd have an `arches` list, which would map to the `index_image` arches.

iib/web/models.py
```diff
@@ -190,6 +204,79 @@ def add_state(self, state, state_reason):
...
+            'id': self.id,
+            'bundles': [bundle.to_json() for bundle in self.bundles],
+            'binary_image': self.binary_image.to_json(),
+            'from_index': self.from_index.to_json(),
```
mprahl: In the API document, we said that all the images would be strings of the pull specification and not a JSON object. That is my preference to make the API simpler. This is especially true since we will not be recording the arches for images other than the built index image, so the empty `arches` list would be misleading.

iib/web/models.py
```diff
@@ -190,6 +204,79 @@ def add_state(self, state, state_reason):
         db.session.flush()
         self.request_state_id = request_state.id
 
+    def to_json(self, verbose=True):
+        states = [
```
mprahl: If `verbose` is not set, use the `state` relationship to get the latest state, otherwise, continue with your approach.

See https://github.com/release-engineering/cachito/pull/87 which fixes this.

iib/web/models.py
```diff
@@ -190,6 +204,79 @@ def add_state(self, state, state_reason):
...
+        return rv
+
+    @classmethod
+    def from_json(cls, kwargs):
+        # Validate all required parameters are present
+        required_params = {'bundles', 'binary_image'}
+        optional_params = {'from_index', 'add_arches', 'cnr_token'}
+        missing_params = required_params - set(kwargs.keys()) - optional_params
+        if missing_params:
+            raise ValidationError('Missing required parameter(s): {}'
```
mprahl: Can you use the following format?
```python
raise ValidationError(
    'Missing required parameter(s): {}'.format(', '.join(missing_params))
)
```

iib/web/models.py
```diff
@@ -190,6 +204,79 @@ def add_state(self, state, state_reason):
...
+
+        # Validate from_index and cnr_token if specified
+        for param in ('from_index', 'cnr_token'):
+            param_value = request_kwargs.get(param)
+            if param_value and not isinstance(param_value, str):
+                raise ValidationError(f'"{param}" must be a string')
+
+        # Validate add_arches are correctly provided
+        add_arches = request_kwargs.get('add_arches', [])
+        if not isinstance(add_arches, list):
```
mprahl: Could you please also make sure it is an array of strings?

iib/web/models.py
```diff
@@ -190,6 +204,79 @@ def add_state(self, state, state_reason):
...
+            raise ValidationError('"add_arches" should be an array')
+
+        # current_user.is_authenticated is only ever False when auth is disabled
+        if not current_user.is_authenticated:
+            raise ValidationError('Cannot set "user" when authentication is disabled')
```
mprahl: This method should just not set `user_id` if authentication is disabled since it allows us to not have authentication setup for the dev environment. An `Unauthorized` exception would have been raised previous to this method being called if authentication was required and not performed.

iib/web/models.py
```diff
@@ -190,6 +204,79 @@ def add_state(self, state, state_reason):
...
+        # Don't allow the user to set arbitrary columns or relationships
+        invalid_params = set(kwargs.keys()) - required_params - optional_params
+        if invalid_params:
+            raise ValidationError(
+                'The following parameters are invalid: {}'.format(', '.join(invalid_params)))
+
+        request_kwargs = deepcopy(kwargs)
+
+        # Validate bundles are correctly provided
+        bundles = request_kwargs.get('bundles')
```
mprahl: `request_kwargs['bundles']` would need to be an array of `Image` objects before it is passed in `request = cls(**request_kwargs)`

iib/web/models.py
```diff
@@ -190,6 +204,79 @@ def add_state(self, state, state_reason):
...
+        if not isinstance(bundles, list) or len(bundles) == 0:
+            raise ValidationError('"bundles" should be a non-empty array')
+
+        # Validate binary_image is correctly provided
+        if not isinstance(request_kwargs.get('binary_image'), str):
```
mprahl: `request_kwargs['binary_image']` would need to be an `Image` object before it is passed in `request = cls(**request_kwargs)`

iib/web/models.py
```diff
@@ -190,6 +204,79 @@ def add_state(self, state, state_reason):
...
+            raise ValidationError('"binary_image" should be a string')
+
+        # Validate from_index and cnr_token if specified
+        for param in ('from_index', 'cnr_token'):
```
mprahl: `request_kwargs['from_index']` would need to be an `Image` object before it is passed in `request = cls(**request_kwargs)`

mprahl: We don't store `cnr_token`, so it should be removed from `request_kwargs` before it is passed in `request = cls(**request_kwargs)`

iib/web/models.py
```diff
@@ -190,6 +204,79 @@ def add_state(self, state, state_reason):
...
+
+        # Validate add_arches are correctly provided
+        add_arches = request_kwargs.get('add_arches', [])
+        if not isinstance(add_arches, list):
+            raise ValidationError('"add_arches" should be an array')
+
+        # current_user.is_authenticated is only ever False when auth is disabled
+        if not current_user.is_authenticated:
+            raise ValidationError('Cannot set "user" when authentication is disabled')
+        request_kwargs['user_id'] = current_user.id
```
mprahl: Use `request_kwargs['user'] = current_user` instead if it works. I'd like to avoid interacting directly with the foreign key columns and let SQLAlchemy figure it out with the relationships we configured in the models.

iib/web/api_v1.py
```diff
@@ -10,3 +14,36 @@
...
+
+
+@api_v1.route('/requests/add', methods=['POST'])
+@login_required
+def add_bundle():
+    """
+    Submit a request to resolve and cache the given source code and its dependencies.
+
+    :param list<str> bundles: each string is a URL to a bundle
+    :param str from_index: index image to which the bundles are to be added. If not specified,
```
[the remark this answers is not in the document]
yashvardhannanavati: D'oh! what was I thinking..sorry about that. I'll fix this

iib/web/models.py
yashvardhannanavati: Regarding the `arches_list`. Shouldn't it be part of the `index_image`? Or should we have the arches in the `index_image` json and have the `arches_list`?

yashvardhannanavati: Ok, I see you prefer to have only the pull_spec for all the images. I'll change it. :)

iib/web/models.py
```diff
@@ -190,6 +192,93 @@ def add_state(self, state, state_reason):
...
+            rv['state_history'] = states
+
+        return rv
+
+    @classmethod
+    def from_json(cls, kwargs):
+        # Validate all required parameters are present
+        required_params = {'bundles', 'binary_image'}
+        optional_params = {'from_index', 'add_arches', 'cnr_token'}
+        missing_params = required_params - set(kwargs.keys()) - optional_params
```
mprahl: Optional: In Python 3, `.keys()` produces an object that can be used in set operations, so you don't need to cast it to a set.

iib/web/models.py
```diff
@@ -190,6 +192,93 @@ def add_state(self, state, state_reason):
...
+                raise ValidationError(f'"{param}" must be a string')
+            if param == 'from_index' and param_value:
+                request_kwargs['from_index'] = Image(pull_specification=param_value)
+
+        # Validate add_arches are correctly provided
+        add_arches = request_kwargs.pop('add_arches', [])
+        if not isinstance(add_arches, list) or not all(isinstance(arch, str) for arch in add_arches):
+            raise ValidationError('"add_arches" should be an array of strings')
+
+        if not request_kwargs.get('from_index') and not request_kwargs.get('add_arches'):
```
mprahl: Move this up in the method so that we don't waste processing time when the request is doomed to fail.

iib/web/models.py
```diff
@@ -190,6 +192,93 @@ def add_state(self, state, state_reason):
...
+        required_params = {'bundles', 'binary_image'}
+        optional_params = {'from_index', 'add_arches', 'cnr_token'}
+        missing_params = required_params - set(kwargs.keys()) - optional_params
+        if missing_params:
+            raise ValidationError(
+                'Missing required parameter(s): {}'.format(', '.join(missing_params))
+            )
+
+        # Don't allow the user to set arbitrary columns or relationships
+        invalid_params = set(kwargs.keys()) - required_params - optional_params
```
mprahl: Optional: Same as above

iib/web/models.py
```diff
@@ -190,6 +192,93 @@ def add_state(self, state, state_reason):
...
+        if invalid_params:
+            raise ValidationError(
+                'The following parameters are invalid: {}'.format(', '.join(invalid_params))
+            )
+
+        request_kwargs = deepcopy(kwargs)
+
+        # Validate bundles are correctly provided
+        bundles = request_kwargs.pop('bundles')
+        if not isinstance(bundles, list) or len(bundles) == 0:
```
mprahl: Could you also validate that this is a list of a strings?

iib/web/models.py
```diff
@@ -190,6 +192,93 @@ def add_state(self, state, state_reason):
...
+            raise ValidationError('"bundles" should be a non-empty array')
+        request_kwargs['bundles'] = [Image(pull_specification=bundle) for bundle in bundles]
+
+        # Validate binary_image is correctly provided
+        binary_image = request_kwargs.pop('binary_image')
+        if not isinstance(binary_image, str):
+            raise ValidationError('"binary_image" should be a non-empty string')
```
mprahl: This is nice, but it's actually not testing if it's empty or not  :)

iib/web/models.py
```diff
@@ -190,6 +192,93 @@ def add_state(self, state, state_reason):
...
+                'The following parameters are invalid: {}'.format(', '.join(invalid_params))
+            )
+
+        request_kwargs = deepcopy(kwargs)
+
+        # Validate bundles are correctly provided
+        bundles = request_kwargs.pop('bundles')
+        if not isinstance(bundles, list) or len(bundles) == 0:
+            raise ValidationError('"bundles" should be a non-empty array')
+        request_kwargs['bundles'] = [Image(pull_specification=bundle) for bundle in bundles]
```
mprahl: You are assuming that there is no image with this pull specification in the database. You will need to perform a "get or create" here instead.

iib/web/models.py
```diff
@@ -190,6 +192,93 @@ def add_state(self, state, state_reason):
...
+
+        # Validate binary_image is correctly provided
+        binary_image = request_kwargs.pop('binary_image')
+        if not isinstance(binary_image, str):
+            raise ValidationError('"binary_image" should be a non-empty string')
+        request_kwargs['binary_image'] = Image(pull_specification=binary_image)
```
mprahl: You are assuming that there is no image with this pull specification in the database. You will need to perform a "get or create" here instead.

iib/web/models.py
```diff
@@ -190,6 +192,93 @@ def add_state(self, state, state_reason):
...
+
+        # Validate from_index and cnr_token if specified
+        for param in ('from_index', 'cnr_token'):
+            param_value = request_kwargs.pop(param)
+            if param_value and not isinstance(param_value, str):
+                raise ValidationError(f'"{param}" must be a string')
+            if param == 'from_index' and param_value:
+                request_kwargs['from_index'] = Image(pull_specification=param_value)
```
mprahl: You are assuming that there is no image with this pull specification in the database. You will need to perform a "get or create" here instead.

iib/web/models.py
```diff
@@ -190,6 +192,93 @@ def add_state(self, state, state_reason):
...
+
+        # Validate binary_image is correctly provided
+        binary_image = request_kwargs.pop('binary_image')
+        if not isinstance(binary_image, str):
+            raise ValidationError('"binary_image" should be a non-empty string')
+        request_kwargs['binary_image'] = Image(pull_specification=binary_image)
+
+        # Validate from_index and cnr_token if specified
+        for param in ('from_index', 'cnr_token'):
+            param_value = request_kwargs.pop(param)
```
mprahl: This will raise a `KeyError` if they are not provided, which should not happen since these are optional parameters.

iib/web/models.py
```diff
@@ -190,6 +192,93 @@ def add_state(self, state, state_reason):
...
+            if param_value and not isinstance(param_value, str):
```
mprahl: This should probably verify it's not an empty string.

iib/web/models.py
yashvardhannanavati: Nice catch! I should've added the default return value. I'll fix it:)

iib/web/models.py
yashvardhannanavati: I'm sorry I didn't get this change request

iib/web/api_v1.py
```diff
@@ -10,3 +14,26 @@
 def get_request():
     ping.delay()
     return 'Test request success!'
+
+
+@api_v1.route('/requests/add', methods=['POST'])
+@login_required
+def add_bundle():
+    """
+    Submit a request to add operator bundles to index images.
```
mprahl:
```suggestion
    Submit a request to add operator bundles to an index image.
```

iib/web/models.py
```diff
@@ -190,6 +209,101 @@ def add_state(self, state, state_reason):
...
+            'id': self.id,
+            'arches': [arch.name for arch in getattr(self.index_image, 'architectures', [])],
+            'binary_image': self.binary_image.pull_specification,
+            'binary_image_resolved': getattr(
+                self.binary_image_resolved, 'pull_specification', None),
+            'bundles': [bundle.pull_specification for bundle in self.bundles],
+            'from_index': getattr(self.from_index, 'pull_specification', None),
+            'from_index_resolved': getattr(self.from_index_resolved, 'pull_specification', None),
+            'index_image': getattr(self.index_image, 'pull_specification', None),
+            'state': _state_to_json(self.state),
```
mprahl: I don't know what SQLAlchemy does under the hood, but it might be nice to do this as part of the else branch of the `if verbose`. The `if verbose` branch can then pick the 0th index of `states` to get the latest state. It might save one query to the database if `verbose` is `True`.

iib/web/models.py
mprahl: With your current logic, an exception will not be raised if `from_index` is an empty string.

```python
>>> param_value = ''
>>> if param_value and not isinstance(param_value, str):
...     print('invalid')
... else:
...     print('not invalid')
... 
not invalid
```

I think you could simplify this code by adding
```python
if param_value is None:
    continue
```

iib/web/models.py
```diff
@@ -190,6 +209,101 @@ def add_state(self, state, state_reason):
...
+            Image.get_or_create(pull_specification=bundle) for bundle in bundles
+        ]
+
+        # Validate binary_image is correctly provided
+        binary_image = request_kwargs.pop('binary_image')
+        if not isinstance(binary_image, str) or not binary_image:
+            raise ValidationError('"binary_image" should be a non-empty string')
+        request_kwargs['binary_image'] = Image.get_or_create(pull_specification=binary_image)
+
+        if not request_kwargs.get('from_index') and not request_kwargs.get('add_arches'):
```
mprahl: Could you move this to right after the `if invalid_params:` block?

iib/web/models.py
```diff
@@ -190,6 +209,101 @@ def add_state(self, state, state_reason):
...
+            if param_value and not isinstance(param_value, str):
+                raise ValidationError(f'"{param}" must be a string')
+            if param == 'from_index' and param_value:
+                request_kwargs['from_index'] = Image.get_or_create(pull_specification=param_value)
+
+        # Validate add_arches are correctly provided
+        add_arches = request_kwargs.pop('add_arches', [])
+        if (
+            not isinstance(add_arches, list) or
+            not all(isinstance(arch, str) for arch in add_arches)
```
```suggestion
            any(not arch or not isinstance(arch, str) for arch in add_arches)
```

iib/web/models.py
```diff
@@ -190,6 +209,101 @@ def add_state(self, state, state_reason):
...
+            )
+
+        request_kwargs = deepcopy(kwargs)
+
+        # Validate bundles are correctly provided
+        bundles = request_kwargs.pop('bundles')
+        if (
+            not isinstance(bundles, list) or
+            len(bundles) == 0 or
+            not all(isinstance(bundle, str) for bundle in bundles)
```
```suggestion
            any(not bundle or not isinstance(bundle, str) for bundle in bundles)
```

iib/web/models.py
```diff
@@ -190,6 +209,101 @@ def add_state(self, state, state_reason):
...
+        ):
+            raise ValidationError('"bundles" should be a non-empty array')
```
```suggestion
            raise ValidationError('"bundles" should be a non-empty array of strings')
```

iib/web/models.py
```diff
@@ -190,6 +209,101 @@ def add_state(self, state, state_reason):
...
+        request_kwargs['binary_image'] = Image.get_or_create(pull_specification=binary_image)
+
+        if not request_kwargs.get('from_index') and not request_kwargs.get('add_arches'):
+            raise ValidationError('One of "from_index" or "add_arches" must be specified')
+
+        # Validate from_index and cnr_token if specified
+        for param in ('from_index', 'cnr_token'):
+            param_value = request_kwargs.pop(param, None)
+            if param_value and not isinstance(param_value, str):
+                raise ValidationError(f'"{param}" must be a string')
```
```suggestion
                raise ValidationError(f'"{param}" must be a non-empty string')
```

iib/web/models.py
```diff
@@ -190,6 +209,101 @@ def add_state(self, state, state_reason):
...
+        if verbose:
+            states = [_state_to_json(state) for state in self.states]
+            # Reverse the list since the latest states should be first
+            states = list(reversed(states))
+            rv['state_history'] = states
+
+        return rv
+
+    @classmethod
+    def from_json(cls, kwargs):
```
mprahl: Optional: I find the validation logic in here quite complex. Perhaps we can consider using JSON schema to validate the input JSON instead. We can pair-program through this together if you'd like.

iib/web/models.py
yashvardhannanavati: I don't think an exception will be raised when `from_index` is an empty string as shown by your example itself. An exception will only be raised when param_value is a non-empty value and not a string. Am I missing something?

iib/web/models.py
yashvardhannanavati: I don't think we should add `non-empty` here because that would imply that it's mandator to specify the param. thoughts?

iib/web/models.py
```diff
@@ -190,6 +209,104 @@ def add_state(self, state, state_reason):
...
+        }
+
+        latest_state = None
+        if verbose:
+            states = [_state_to_json(state) for state in self.states]
+            # Reverse the list since the latest states should be first
+            states = list(reversed(states))
+            rv['state_history'] = states
+            latest_state = states[0]
+        rv['state'] = latest_state or _state_to_json(self.state)
```
mprahl: We want "state", "state_reason" and "updated" to be on the top-level of the JSON object.

```suggestion
        rv.update(latest_state or _state_to_json(self.state))
```

iib/web/models.py
yashvardhannanavati: As discussed on the chat, let's add a SG in phase 2 of this project to use https://github.com/zalando/connexion for validation. Great find Matt :D

iib/web/models.py
yashvardhannanavati: sorry I misread your comment. we want the exception to be raised? I don't see why we should raise an exception when an empty string is specified.

mprahl: We decided that an empty string should be treated as null.

yashvardhannanavati:
Thanks Matt!

PR #6, opened 2020-02-10
Add the opm_index_add Celery task

mprahl:


docker-compose.yml
```diff
@@ -45,17 +61,23 @@ services:
     build:
       context: .
       dockerfile: ./docker/Dockerfile-workers
-    # Override the default command so that Celery auto-reloads on code changes
+    # Override the default command so that Celery auto-reloads on code changes.
+    # This also adds the self-signed CA that was used to sign the Docker registry's certificate
+    # to the trusted CA bundle. This will make podman trust the Docker registry's certificate.
```
yashvardhannanavati: how about "podman trust the Docker registry's certificate." => "podman trust the local Docker registry's certificate."

iib/workers/config.py
```diff
@@ -74,6 +80,19 @@ def validate_celery_config(conf, **kwargs):
...
+    if not conf.get('iib_registry_credentials'):
+        raise ConfigError(
+            'iib_registry_credentials must be set in the format of "username:password"'
+        )
+
+
+def get_worker_config():
+    # Import this here to avoid a circular import
+    import iib.workers.tasks.celery
+
```
yashvardhannanavati: optional: this newline seems unnecessary. we could remove it.

iib/workers/tasks/build.py
```diff
@@ -0,0 +1,171 @@
...
+
+    This is a workaround until https://github.com/operator-framework/operator-registry/pull/173 is
+    released.
+
+    :param str dockerfile_dir: the path to the directory containing the data generated by the
+        opm command.
+    """
+    log.debug('Fixing the opm path in index.Dockerfile')
+    dockerfile_path = os.path.join(dockerfile_dir, 'index.Dockerfile')
+    for line in fileinput.input(dockerfile_path, inplace=True):
```
yashvardhannanavati: Can I please request you to add a comment just above this.

iib/workers/tasks/build.py
```diff
@@ -0,0 +1,171 @@
...
+        bundles_str = ', '.join(bundles)
+        if from_index:
+            log.info(
+                'Generating the database file from %s and adding the following bundle(s): %s',
+                from_index,
+                bundles_str,
+            )
+            cmd.extend(['--from-index', from_index])
+        else:
+            log.info(
```
yashvardhannanavati: Can we log here that from_index wasn't specified?

yashvardhannanavati: we can potentially simplify this if-else block by logging something like `log.info(
                'Generating the database file with the following bundle(s): %s'.,
                bundles_str,
            )` and then adding another log statement with the value of the `index_image` only if it is specified. That way we can eliminate the else block. does that make sense?

iib/workers/config.py
mprahl: I've been using the auto-formatting tool "black" and it did this automatically, so I'll keep it as is.

mprahl:
@yashvardhannanavati thanks for the review. This is ready for another round.

yashvardhannanavati:
:ship:

PR #7, opened 2020-02-12
Add 'Patch' API endpoint

yashvardhannanavati:


yashvardhannanavati:
@mprahl I'll add the tests once you are good with the implementation

iib/web/api_v1.py
```diff
@@ -37,3 +38,94 @@ def add_bundles():
 
     flask.current_app.logger.debug('Successfully scheduled request %d', request.id)
     return flask.jsonify(request.to_json()), 201
+
+
+@api_v1.route('/requests/<int:request_id>', methods=['PATCH'])
```
mprahl:
```suggestion
@api_v1.route('/builds/<int:request_id>', methods=['PATCH'])
```

iib/web/api_v1.py
```diff
@@ -37,3 +38,94 @@ def add_bundles():
...
+
+    payload = flask.request.get_json()
+    if not isinstance(payload, dict):
+        raise ValidationError('The input data must be a JSON object')
+
+    if not payload:
+        raise ValidationError('At least one key must be specified to update the request')
+
+    valid_keys = {
+        'arches', 'bundles', 'binary_image_resolved', 'from_index_resolved', 'index_image',
```
mprahl: So since the bundles are added to the database when the request is created and we aren't resolving those, they should not be modifiable.

iib/web/api_v1.py
```diff
@@ -37,3 +38,94 @@ def add_bundles():
...
+        'state', 'state_reason'
+    }
+    invalid_keys = payload.keys() - valid_keys
+    if invalid_keys:
+        raise ValidationError(
+            'The following keys are not allowed: {}'.format(', '.join(invalid_keys)))
+
+    for key, value in payload.items():
+        if key in ('bundles', 'arches'):
+            if (
```
mprahl: This is duplicating the code in `from_json`. How about creating a static method on the `Architecture` class that validates the `arches` JSON and use that method in `from_json` and here.

iib/web/api_v1.py
```diff
@@ -37,3 +38,94 @@ def add_bundles():
...
+    if 'bundles' in payload:
+        for bundle in payload.get('bundles', []):
+            image_object = Image.get_or_create(bundle)
+            if image_object not in request.bundles:
+                request.bundles.append(image_object)
+
+    for key in ('binary_image_resolved', 'from_index_resolved', 'index_image'):
+        if key not in payload:
+            continue
+        if key == 'index_image' and 'arches' in payload:
```
mprahl: The `arches` payload will be supplied independently from `index_image` because `index_image` is the pull specification to the manifest list (each manifest being an image for a specific architecture), which gets generated after the images for each architecture are built.

This leads me to believe that we should associate architectures directly with the request instead to avoid needing a results backend.

iib/web/api_v1.py
yashvardhannanavati: I thought we decided on keeping it `requests` since it's an internal endpoint used to patch requests.

iib/web/api_v1.py
yashvardhannanavati: :+1:

iib/web/api_v1.py
yashvardhannanavati: Ah, that makes sense. I agree, we should associate architectures directly with the request instead of index_image. I'll do that as a part of this PR

iib/web/api_v1.py
mprahl: To be consistent and "RESTful", we should use the same name.

iib/web/models.py
```diff
@@ -49,18 +49,30 @@ class Architecture(db.Model):
...
 
+    @staticmethod
+    def validate_architecture_json(arches):
+        """
+        Validate the JSON representation of architectures.
 
-class ImageArchitecture(db.Model):
-    """
-    An association table between images and the architectures they were built for.
+        :param list arches: the JSON representation of architectures for an image
```
```suggestion
        :param list arches: the JSON representation of architectures for a build request
```

iib/web/api_v1.py
```diff
@@ -37,3 +38,83 @@ def add_bundles():
...
+
+    payload = flask.request.get_json()
+    if not isinstance(payload, dict):
+        raise ValidationError('The input data must be a JSON object')
+
+    if not payload:
+        raise ValidationError('At least one key must be specified to update the request')
+
+    valid_keys = {
+        'add_arches', 'binary_image_resolved', 'from_index_resolved', 'index_image',
```
mprahl: How about `arches` instead of `add_arches`? `add_arches` is only relevant in the POST request since we are adding to what is currently set on the index image or nothing if there is no index image.

iib/web/api_v1.py
```diff
@@ -37,3 +38,83 @@ def add_bundles():
...
+    invalid_keys = payload.keys() - valid_keys
+    if invalid_keys:
+        raise ValidationError(
+            'The following keys are not allowed: {}'.format(', '.join(invalid_keys)))
+
+    for key, value in payload.items():
+        if key == 'add_arches':
+            Architecture.validate_architecture_json(value)
+
+        if key in ('binary_image_resolved', 'from_index_resolved', 'index_image'):
```
mprahl:
```suggestion
        elif key in ('binary_image_resolved', 'from_index_resolved', 'index_image'):
```

iib/web/api_v1.py
```diff
@@ -37,3 +38,83 @@ def add_bundles():
...
+            flask.current_app.logger.info('Not adding a new state since it matches the last state')
+        else:
+            request.add_state(new_state, new_state_reason)
+
+    for key in ('binary_image_resolved', 'from_index_resolved', 'index_image'):
+        if key not in payload:
+            continue
+        key_value = payload.get(key, None)
+        key_object = Image.get_or_create(key_value)
+        if key_object not in getattr(request, key):
```
mprahl: Since these relationships have `uselist=False`, it should just be set with normal value assignments and not like interacting with a list.

iib/web/api_v1.py
```diff
@@ -37,3 +38,83 @@ def add_bundles():
...
+            getattr(request, key).append(key_object)
+
+    if 'add_arches' in payload:
```
mprahl: This if statement can be removed since your for loop guards against the key not being set.

iib/web/api_v1.py
yashvardhannanavati: sure

iib/web/api_v1.py
```diff
@@ -85,3 +88,82 @@ def add_bundles():
...
+    }
+    invalid_keys = payload.keys() - valid_keys
+    if invalid_keys:
+        raise ValidationError(
+            'The following keys are not allowed: {}'.format(', '.join(invalid_keys)))
+
+    for key, value in payload.items():
+        if key == 'arches':
+            Architecture.validate_architecture_json(value)
+        elif key in ('binary_image_resolved', 'from_index_resolved', 'index_image'):
```
mprahl: All other possible values are strings, so you might as well reuse that logic.
```suggestion
        elif not value or not isinstance(value, str):
```

iib/web/api_v1.py
```diff
@@ -85,3 +88,82 @@ def add_bundles():
...
+            flask.current_app.logger.info('Not adding a new state since it matches the last state')
+        else:
+            request.add_state(new_state, new_state_reason)
+
+    for key in ('binary_image_resolved', 'from_index_resolved', 'index_image'):
+        if key not in payload:
+            continue
+        key_value = payload.get(key, None)
+        key_object = Image.get_or_create(key_value)
+        if key_object is not getattr(request, key, None):
```
mprahl: SQLAlchemy is smart enough to not make any changes to the database if you assign the same value, so you can simplify this by removing the if statement. You can leave a comment explaining this.

iib/web/models.py
```diff
@@ -275,7 +286,7 @@ def _state_to_json(state):
 
         rv = {
             'id': self.id,
-            'arches': [arch.name for arch in getattr(self.index_image, 'architectures', [])],
+            'arches': [arch.name for arch in getattr(self, 'architectures', [])],
```
```suggestion
            'arches': [arch.name for arch in self.architectures],
```

tests/test_web/test_api_v1.py
```diff
@@ -206,13 +206,124 @@ def test_add_bundle_success(db, auth_env, client):
...
+    data = {
+        'arches': arches,
+        'binary_image_resolved': binary_image_resolved,
+        'from_index_resolved': from_index_resolved,
+        'index_image': index_image,
+        'state': state,
+        'state_reason': state_reason
+    }
+
+    if data['state'] == 'no_state':
```
mprahl: Instead of "magic" like this, why not just have a single parameter of `data` which is a dictionary that allows you to set whatever you'd like?

tests/test_web/test_api_v1.py
```diff
@@ -206,13 +206,124 @@ def test_add_bundle_success(db, auth_env, client):
...
+
+def test_patch_request_success(db, worker_auth_env, client):
+    data = {
+        'arches': ['arches'],
+        'state': 'complete',
+        'state_reason': 'All done!',
+        'index_image': 'index:image'
+    }
+
+    response_json = {
```
mprahl: Since we already test the JSON method elsewhere, you could just do:
```python
assert rv_json == db.session.query(Request).get(1).to_json(verbose=True)
```

yashvardhannanavati: I prefer having the entire json in the success test. It helps get a glance of the response without having to navigate out of your editor. If you feel strongly against it I can change it.

tests/test_web/test_api_v1.py
```diff
@@ -207,14 +207,117 @@ def test_add_bundle_success(db, auth_env, client):
...
+                'index_image': 'index:image',
+                'state': 'no_state',
+                'state_reason': 'state_reason',
+            },
+            'The "state" key is required when "state_reason" is supplied',
+        ),
+    ),
+)
+def test_patch_request_invalid_params_format(data, error_msg, db, worker_auth_env, client):
+    data = data
```
mprahl: Could you please remove this line?

tests/test_web/test_api_v1.py
```diff
@@ -207,14 +207,117 @@ def test_add_bundle_success(db, auth_env, client):
...
+            },
+            'The value for "binary_image_resolved" must be a non-empty string',
+        ),
+        (
+            {
+                'arches': ['s390x'],
+                'binary_image_resolved': 'resolved:binary',
+                'from_index_resolved': 'resolved:index',
+                'index_image': 'index:image',
+                'state': 'no_state',
```
mprahl: Instead of this, just remove this key.

PR #9, opened 2020-02-12
Add the GET API endpoints

mprahl:


iib/web/api_v1.py
```diff
@@ -16,6 +17,57 @@ def get_request():
...
+    Retrieve the paginated build requests.
+
+    :rtype: flask.Response
+    """
+    state = flask.request.args.get('state')
+    verbose = str_to_bool(flask.request.args.get('verbose'))
+    max_per_page = flask.current_app.config['IIB_MAX_PER_PAGE']
+
+    query = Request.query.options(*Request.get_query_options(verbose=verbose))
+    if state:
```
yashvardhannanavati: I was wondering if we can move this validation to a static method in `RequestState` so that I can re-use this bit in the patch API too. thoughts?

iib/web/models.py
```diff
@@ -215,6 +220,36 @@ def add_state(self, state, state_reason):
         db.session.flush()
         self.request_state_id = request_state.id
 
+    @staticmethod
+    def get_query_options(verbose=False):
```
yashvardhannanavati: very interesting!

PR #10, opened 2020-02-13
Use black to ensure the code style is consistent

mprahl:
"No colors anymore I want them to turn black" - The Rolling Stones


yashvardhannanavati:
:+1:

PR #12, opened 2020-02-13
Add 'remove' API endpoint

yashvardhannanavati:


iib/web/models.py
```diff
@@ -137,6 +137,33 @@ def get_or_create(cls, pull_specification):
         return image
 
 
+class Operator(db.Model):
+    """
+    An operator that has been handled by IIB.
+    """
+
+    id = db.Column(db.Integer, primary_key=True)
+    name = db.Column(db.String, nullable=False, unique=True)
```
mprahl: There should be an index on this column since we will often filter on it when performing a `get_or_create`.

I realize I forgot to add one on `Image.pull_specification` though. Could you please add one to to it in a separate commit?

iib/web/models.py
```diff
@@ -309,8 +340,21 @@ def _state_to_json(state):
...
-        required_params = {'bundles', 'binary_image'}
+        required_params = {'bundles', 'binary_image', 'operators'}
         optional_params = {'from_index', 'add_arches', 'cnr_token'}
+
+        if 'bundles' in kwargs:
+            request_type = RequestTypeMapping.__members__['add'].value
+            required_params.remove('operators')
+        elif 'operators' in kwargs:
+            request_type = RequestTypeMapping.__members__['rm'].value
+            required_params.remove('operators')
```
```suggestion
            required_params.remove('bundles')
```

iib/web/models.py
```diff
@@ -309,8 +340,21 @@ def _state_to_json(state):
     @classmethod
     def from_json(cls, kwargs):
         # Validate all required parameters are present
-        required_params = {'bundles', 'binary_image'}
+        required_params = {'bundles', 'binary_image', 'operators'}
         optional_params = {'from_index', 'add_arches', 'cnr_token'}
+
+        if 'bundles' in kwargs:
+            request_type = RequestTypeMapping.__members__['add'].value
+            required_params.remove('operators')
```
mprahl: My preference is to add `bundles` to `required_params` rather than remove `operators`. It makes it more intuitive to read.

iib/web/models.py
```diff
@@ -365,12 +401,35 @@ def from_json(cls, kwargs):
...
-
         request = cls(**request_kwargs)
         request.add_state('in_progress', 'The request was initiated')
         return request
 
+    @staticmethod
+    def _handle_request_exclusive_params(request_kwargs):
+        """Handles database operations specific to add/rm requests."""
+
+        # Assume the request is an add request
```
mprahl: My preference is to not assume here. It's possible another request type will be added in the future and we don't want this method to assume `add`.

iib/web/models.py
```diff
@@ -309,8 +340,21 @@ def _state_to_json(state):
     @classmethod
     def from_json(cls, kwargs):
         # Validate all required parameters are present
-        required_params = {'bundles', 'binary_image'}
+        required_params = {'bundles', 'binary_image', 'operators'}
         optional_params = {'from_index', 'add_arches', 'cnr_token'}
+
+        if 'bundles' in kwargs:
```
mprahl: What if I specify `operators` in my JSON payload for the `add` endpoint? IIB will fail in unexpected ways. My preference is to have two methods `from_add_json` and `from_remove_json` that both wrap `from_json`. `from_json` could accept an additional keyword argument such as `additional_required_params=None`.  `from_json` should also be renamed to `_from_json` since it shouldn't be called directly anymore.

iib/web/models.py
```diff
@@ -365,12 +401,35 @@ def from_json(cls, kwargs):
...
+            or any(not item or not isinstance(item, str) for item in value)
+        ):
+            raise ValidationError(f'"{key}" should be a non-empty array of strings')
+
+        if key == 'bundles':
+            request_kwargs['bundles'] = [
+                Image.get_or_create(pull_specification=item) for item in value
+            ]
+        else:
+            request_kwargs['operators'] = [Operator.get_or_create(name=item) for item in value]
```
mprahl: The `get_or_create` method doesn't exist on the `Operator` class. I think you could probably have a base class that has a `get_or_create` method that `Architecture`, `Image`, and `Operator` inherit from. The derived class could wrap the base class `get_or_create` by passing in the column name to filter on to the base class `get_or_create`.

tests/test_web/test_api_v1.py
```diff
@@ -123,51 +123,70 @@ def test_get_builds_invalid_state(app, client, db):
 
 
 @pytest.mark.parametrize(
-    'bundles, from_index, binary_image, add_arches, error_msg',
+    'data, error_msg',
```
mprahl: This is a nice change but it should be in a separate commit or PR.

tests/test_web/test_api_v1.py
```diff
@@ -123,51 +123,70 @@ def test_get_builds_invalid_state(app, client, db):
...
-        'from_index': 'from_index',
-        'binary_image': 'binary:image',
-        'add_arches': ['add_arches'],
-    }
-
+@pytest.mark.parametrize(
+    'data, error_msg',
+    (
+        (
+            {'from_index': 'pull:spec', 'binary_image': 'binary:image', 'add_arches': ['s390x'], },
```
yashvardhannanavati: @mprahl Here's the scenario between flake8 and black I was talking about. `space after the second last comma`

iib/web/models.py
yashvardhannanavati: I did think of it but then we would be wrongly representing the two sets in the beginning. I found this approach to be more clear.

iib/web/models.py
yashvardhannanavati: I agree, yours is a more cleaner solution. I'll change it. Thanks!

iib/web/models.py
yashvardhannanavati: I don't know how I feel about doing this though. Because there's some validation in the Image `get_or_create` method. There's a chance we add something similar in others too.

iib/web/models.py
yashvardhannanavati: Though, it wouldn't fail in unexpected ways if you specify `operators` in the JSON payload for the `add` endpoint. It'll fail with Invalid params validation error.

yashvardhannanavati: The solution you suggested will probably duplicate some of the code but I guess it's worth it for readability.

iib/web/models.py
```diff
@@ -306,51 +357,41 @@ def _state_to_json(state):
 
         return rv
 
-    @classmethod
-    def from_json(cls, kwargs):
+    @staticmethod
+    def _from_json(request_kwargs, additional_required_params=None):
```
mprahl: Could you please add a docstring here?

It's not intuitive at first that `request_kwargs` will be updated in the method.

iib/web/models.py
```diff
@@ -365,12 +406,52 @@ def from_json(cls, kwargs):
...
-        request_kwargs['type'] = RequestTypeMapping.__members__['add'].value
+    @classmethod
+    def from_add_json(cls, kwargs):
+        """ Handles JSON requests for Add API endpoint """
+        request_kwargs = deepcopy(kwargs)
 
+        cls._from_json(request_kwargs, additional_required_params=['bundles'])
+
+        bundles = request_kwargs.pop('bundles')
+        if (
```
mprahl: Could you please move this if statement before `cls._from_json(...` so that if this parameter is missing, it will not cause unnecessary load on the database?

iib/web/models.py
```diff
@@ -30,6 +30,7 @@ class RequestStateMapping(BaseEnum):
     in_progress = 1
     complete = 2
     failed = 3
+    removed = 4
```
mprahl: A new state is not necessary. Remember, this state is about the request and not the index image. So `in_progress` => `complete` or `failed` works just fine with the remove workflow.

tests/test_web/test_api_v1.py
```diff
@@ -117,57 +117,96 @@ def test_get_builds_invalid_state(app, client, db):
...
-):
-    data = {
-        'bundles': bundles,
-        'from_index': from_index,
-        'binary_image': binary_image,
-        'add_arches': add_arches,
-    }
-
-    rv = client.post('/api/v1/builds/add', json=data, environ_base=auth_env)
+def test_add_or_rm_invalid_params_format(data, request_type, error_msg, db, auth_env, client):
```
mprahl: Could you make these separate tests (one for add and one for rm)? By keeping them together, you only save 3 lines of code and it's harder to read.

tests/test_web/test_api_v1.py
```diff
@@ -117,57 +117,96 @@ def test_get_builds_invalid_state(app, client, db):
...
+            'Missing required parameter(s): binary_image',
+        ),
+        (
+            {'from_index': 'pull:spec', 'binary_image': 'binary:image', 'add_arches': ['s390x1']},
+            'rm',
+            'Missing required parameter(s): operators',
+        ),
+    ),
+)
+def test_add_or_rm_missing_required_param(data, request_type, error_msg, db, auth_env, client):
```
mprahl: Could you make these separate tests (one for add and one for rm)? By keeping them together, you only save 3 lines of code and it's harder to read.

iib/web/models.py
```diff
@@ -306,51 +356,43 @@ def _state_to_json(state):
 
         return rv
 
-    @classmethod
-    def from_json(cls, kwargs):
+    @staticmethod
+    def _from_json(request_kwargs, additional_required_params=None):
+        """
+        Validates and processes request agnostic parameters
```
```suggestion
        Validate and process request agnostic parameters.
```

mprahl:
@yashvardhannanavati could you please rebase? There's a conflict after merging #14.

yashvardhannanavati:
Done

iib/web/models.py
```diff
@@ -288,13 +334,17 @@ def _state_to_json(state):
...
             'from_index': getattr(self.from_index, 'pull_specification', None),
             'from_index_resolved': getattr(self.from_index_resolved, 'pull_specification', None),
             'index_image': getattr(self.index_image, 'pull_specification', None),
             'user': getattr(self.user, 'username', None),
         }
 
+        if self.type == RequestTypeMapping.__members__['add'].value:
+            rv['bundles'] = [bundle.pull_specification for bundle in self.bundles]
+        else:
+            rv['operators'] = [operator.name for operator in self.operators]
```
midnightercz: Just a question here. What is the format of operator in relation to bundle? So bundle is `<host>/<space>/<name>:<tag>` and operator in that context is `<space>/<name>:<tag>` ?
I'm thinking if it would be possible to return operators when adding bundles as well. It makes sense to me to gather this logic rather here (mapping bundles -> operators) than in pub

iib/web/models.py
```diff
@@ -306,51 +356,45 @@ def _state_to_json(state):
...
+        :param dict request_kwargs: copy of args provided in API request
+        """
         # Validate all required parameters are present
-        required_params = {'bundles', 'binary_image'}
-        optional_params = {'from_index', 'add_arches', 'cnr_token'}
-        missing_params = required_params - kwargs.keys() - optional_params
+        required_params = {'binary_image'} | set(additional_required_params or [])
+        optional_params = {'add_arches', 'cnr_token'} | set(additional_optional_params or [])
+
+        missing_params = required_params - request_kwargs.keys() - optional_params
```
midnightercz: - optional_params doens't make sense here optionals does not intersect with required

iib/web/models.py
mprahl: The purpose of storing the operators in the database is to audit what the original request is and to replay the request if it fails. IIB will not discover what the relationship is between bundles and operators.

iib/web/models.py
mprahl: Agreed

iib/web/models.py
```diff
@@ -365,8 +409,52 @@ def from_json(cls, kwargs):
...
+
+        bundles = request_kwargs.pop('bundles', [])
+        if (
+            not isinstance(bundles, list)
+            or len(bundles) == 0
+            or any(not item or not isinstance(item, str) for item in bundles)
+        ):
+            raise ValidationError(f'"bundles" should be a non-empty array of strings')
+
+        request_kwargs['bundles'] = [
```
mprahl: If you don't `pop` the `bundles` key, you can perform this after the `cls._from_json`, which means the database isn't touched until all validation has occurred which is nice.

iib/web/models.py
```diff
@@ -365,8 +409,52 @@ def from_json(cls, kwargs):
...
+
+        operators = request_kwargs.pop('operators', [])
+        if (
+            not isinstance(operators, list)
+            or len(operators) == 0
+            or any(not item or not isinstance(item, str) for item in operators)
+        ):
+            raise ValidationError(f'"operators" should be a non-empty array of strings')
+
+        request_kwargs['operators'] = [Operator.get_or_create(name=item) for item in operators]
```
mprahl: If you don't `pop` the `operators` key, you can perform this after the `cls._from_json` call, which means the database isn't touched until all validation has occurred which is nice.

iib/web/models.py
```diff
@@ -306,51 +356,45 @@ def _state_to_json(state):
...
         return rv
 
-    @classmethod
-    def from_json(cls, kwargs):
+    @staticmethod
+    def _from_json(
+        request_kwargs, additional_required_params=None, additional_optional_params=None
+    ):
+        """
+        Validate and process request agnostic parameters
```
mprahl: Could you also document that `request_kwargs` will be modified?

tests/test_web/test_api_v1.py
```diff
@@ -123,51 +123,103 @@ def test_get_builds_invalid_state(app, client, db):
...
+    assert rv.status_code == 400
+    assert '"operators" should be a non-empty array of strings' == rv.json['error']
 
-    rv = client.post('/api/v1/builds/add', json=data, environ_base=auth_env)
+
+@pytest.mark.parametrize(
+    'data, error_msg',
+    (
+        (
+            {'from_index': 'pull:spec', 'binary_image': 'binary:image', 'add_arches': ['s390x1']},
```
mprahl:
```suggestion
            {'from_index': 'pull:spec', 'binary_image': 'binary:image', 'add_arches': ['s390x']},
```

mprahl: This suggestion applies to all the other instances of `s390x1`

iib/web/models.py
yashvardhannanavati: Good idea!

tests/test_web/test_api_v1.py
```diff
@@ -123,51 +123,103 @@ def test_get_builds_invalid_state(app, client, db):
...
     assert rv.status_code == 400
-    assert rv.json['error'] == 'Missing required parameter(s): bundles'
+    assert rv.json['error'] == error_msg
+
+
+@pytest.mark.parametrize(
+    'data, error_msg',
+    (
+        (
+            {'operators': ['some:thing'], 'from_index': 'pull:spec', 'add_arches': ['s390x1']},
```
mprahl: Could you also fix these?

```suggestion
            {'operators': ['some:thing'], 'from_index': 'pull:spec', 'add_arches': ['s390x']},
```

iib/web/models.py
```diff
@@ -306,51 +356,46 @@ def _state_to_json(state):
...
         return rv
 
-    @classmethod
-    def from_json(cls, kwargs):
+    @staticmethod
+    def _from_json(
+        request_kwargs, additional_required_params=None, additional_optional_params=None
+    ):
+        """
+        Validate, modify request_args to represent the internal database structure, and
```
mprahl: How about?
```
Validate and process request type agnostic parameters.

As part of the processing, the input ``request_kwargs`` parameter is updated to reference database objects where appropriate.
```

tests/test_web/test_api_v1.py
yashvardhannanavati: Lol, looks like my typo spread wide..sure I'll fix it

PR #13, opened 2020-02-14
Small fixes related to black and tox

mprahl:


requirements-dev.txt
```diff
@@ -1,4 +1,5 @@
 -rrequirements.txt
+black
```
yashvardhannanavati: Do we want to pin the version here?

mprahl: I was torn about this myself, but since you suggested it, it has swayed me.

yashvardhannanavati:
:+1:

PR #14, opened 2020-02-14
Add a task that coordinates index image builds when adding bundles

mprahl:
This PR has a few extra commits:
* Don't log the registry password if a command that uses it fails
* Create the request_architecture table after the request table
* Set the IIB_TESTING environment variable in tox

### DIFF
--- a/iib/workers/tasks/build.py
+++ b/iib/workers/tasks/build.py
@@ -516,7 +516,13 @@ def _run_cmd(cmd, params=None, exc_msg=None):
     response = subprocess.run(cmd, **params)
 
     if response.returncode != 0:
-        log.error('The command "%s" failed with: %s', ' '.join(cmd), response.stderr)
+        conf = get_worker_config()
+        _, password = conf['iib_registry_credentials'].split(':', 1)
+        sanitized_cmd = copy.copy(cmd)
+        for i, arg in enumerate(cmd):
+            if arg in (conf['iib_registry_credentials'], password):
+                sanitized_cmd[i] = '********'
+        log.error('The command "%s" failed with: %s', ' '.join(sanitized_cmd), response.stderr)
         raise IIBError(exc_msg or 'An unexpected error occurred')
 
     return response.stdout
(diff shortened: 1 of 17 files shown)


iib/workers/tasks/build.py
```diff
@@ -80,6 +200,151 @@ def _get_local_pull_spec(request_id):
...
+    All information that was retrieved and/or calculated for the next steps in the build are
+    returned as a dictionary.
+
+    This function was created so that code didn't need to be duplicated for the ``add`` and ``rm``
+    request types.
+
+    :param str binary_image: the pull specification of the image where the opm binary gets copied
+        from.
+    :param int request_id: the ID of the IIB build request
+    :param str from_index: the pull specification of the image containing the index that the index
```
yashvardhannanavati: could you please add the docstring for `add_arches` here

iib/workers/tasks/build.py
```diff
@@ -80,6 +200,151 @@ def _get_local_pull_spec(request_id):
...
+
+    if not arches.issubset(binary_image_arches):
+        raise IIBError(
+            'The binary image is not available for the following arches: {}'.format(
+                ', '.join(sorted(arches - binary_image_arches))
+            )
+        )
+
+    conf = get_worker_config()
+    if not arches.issubset(conf['iib_arches']):
```
yashvardhannanavati: Since we have this, do you think we should remove s390x from our list for now? Since we won't be supporting it when we first deploy IIB?

iib/workers/tasks/build.py
```diff
@@ -80,6 +200,151 @@ def _get_local_pull_spec(request_id):
...
+        log.info('Sleeping for %s seconds', conf['iib_poll_api_frequency'])
+        time.sleep(conf['iib_poll_api_frequency'])
+        log.info(
+            'Polling the API to see if the image builds for the following arches have '
+            'completed: %s',
+            ', '.join(arches_remaining),
+        )
+
+        request = get_request(request_id)
+        if request['state'] != 'in_progress':
```
yashvardhannanavati: can we compare the `state` to `failed` here?

PR #15, opened 2020-02-17
Pin the dependencies using pip-tools

mprahl:
Since this will be deployed on RHEL 8 and several of the required
dependencies are not packaged as RPMs, let's use pip-tools to
ensure the installed dependencies are deterministic (pinned versions)
and relatively secure (hashes).

This also verifies that the pinned dependencies are not vulnerable using `safety`.


lcarva:
Nice!

PR #20, opened 2020-02-19
Add support for the legacy app rigestry

yashvardhannanavati:


yashvardhannanavati:
@mprahl If the implementation looks good, I'll add few tests. :)

iib/workers/tasks/legacy.py
```diff
@@ -0,0 +1,179 @@
...
+
+def _podman_login():
+    """
+    Login into the registry to be able to access private repos.
+
+    :raises iib.exceptions.IIBError: if the login fails
+    """
+    conf = get_worker_config()
+    log.info('Logging in %s using podman', conf['iib_registry'])
+    username, password = conf['iib_registry_credentials'].split(':')
```
lcarva: This is kind of odd. Can we just mount the secret file on the host? Check out `~/.docker/config.json` in your host.

iib/workers/tasks/legacy.py
```diff
@@ -0,0 +1,179 @@
...
+    package_name = package_dir.split('/download-', 1)[-1]
+    log.info('Verifying package_name %s', package_name)
+    # opm does not fail when the package is missing in the index image, hence we
+    # check the number of generated files. If it's equal to 1, that means only an empty
+    # `package.yaml` file is generated and the package is missing
+    if len(os.listdir(package_dir)) == 1:
+        raise IIBError(f'package {package_name} is missing in index image {from_index}')
+
+    try:
+        output = api.build_and_verify(source_dir=package_dir)
```
lcarva: OMPS will run the same verification command. Could we simply rely on it?

iib/workers/tasks/legacy.py
```diff
@@ -0,0 +1,179 @@
...
+
+            log.info('Generating the back ported operator for package: %s', package)
+
+            run_cmd(
+                cmd,
+                {'cwd': temp_dir},
+                exc_msg=f'Failed to export package {package} for index image {from_index}',
+            )
+
+            package_dir = temp_dir.rstrip('/') + f'/download-{package}'
```
lcarva: `os.path.join` ?

iib/workers/tasks/legacy.py
yashvardhannanavati: I haven't looked at the code of OMPS, to find out if it runs this immediately when the zipfile is submitted. So I added it here just to be sure. But I can check if OMPS does this check before returning the response and remove it from here if you want.

iib/workers/tasks/build.py
```diff
@@ -453,6 +443,27 @@ def handle_add_request(bundles, binary_image, request_id, from_index=None, add_a
...
+        conf = get_worker_config()
+        rebuilt_index_image = conf['iib_image_push_template'].format(
+            registry=conf['iib_registry'], request_id=request_id
+        )
+        opm_index_export(
+            legacy_support_packages, request_id, rebuilt_index_image, cnr_token, organization
+        )
+
+        if from_index:
+            # Re-verify index image after exporting it for legacy app registry
```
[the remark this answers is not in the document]
yashvardhannanavati: Ah, nice catch! I should call the `_push_package_manifest` method after re-checking the from_index. 
The reason I do this is to check if the index image has changed since the IIB build started. This is because we cannot use the resolved image in the opm command. So after exporting the package, if the source index image has changed, IIB will fail the request (doesn't maater if the build and export was successful). Does that make sense?

iib/web/api_v1.py
```diff
@@ -92,6 +92,8 @@ def add_bundles():
             request.id,
             payload.get('from_index'),
             payload.get('add_arches'),
+            payload.get('cnr_token'),
```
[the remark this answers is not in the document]
yashvardhannanavati: Let me do some more digging and get back to you on this.

No, it isn't stored permanently.

iib/workers/tasks/build.py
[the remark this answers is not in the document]
lcarva: It does. Can you make sure to add this explanation to the code?

iib/web/models.py
```diff
@@ -401,7 +401,7 @@ def _from_json(
         request_kwargs['binary_image'] = Image.get_or_create(pull_specification=binary_image)
 
         # Validate from_index and cnr_token if specified
-        for param in ('from_index', 'cnr_token'):
+        for param in ('from_index', 'cnr_token', 'organization'):
```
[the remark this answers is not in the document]
yashvardhannanavati: Well, we could if it's useful somewhere. @mprahl WDYT?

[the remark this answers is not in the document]
yashvardhannanavati: We could return it back in the request response then

iib/workers/tasks/legacy.py
yashvardhannanavati: ok, looks like we can rely on OMPS to check this. I'll remove it.

yashvardhannanavati: https://github.com/release-engineering/operators-manifests-push-service/blob/master/omps/quay.py#L312

yashvardhannanavati: As we discussed, having this extra layer of verification doesn't do harm. So let's keep it. :)

iib/web/api_v1.py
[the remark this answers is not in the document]
mprahl: It could get logged though. Let's use this mechanism to hide it:
https://docs.celeryproject.org/en/stable/userguide/tasks.html#hiding-sensitive-information-in-arguments

iib/web/migrations/versions/274ba38408e8_initial_migration.py
```diff
@@ -53,6 +53,7 @@ def upgrade():
         sa.Column('index_image_id', sa.Integer(), nullable=True),
         sa.Column('type', sa.Integer(), nullable=False),
         sa.Column('user_id', sa.Integer(), nullable=True),
+        sa.Column('organization', sa.String(), nullable=True),
```
mprahl: You're missing the downgrade part

iib/workers/config.py
```diff
@@ -49,6 +49,7 @@ class DevelopmentConfig(Config):
     iib_log_level = 'DEBUG'
     iib_registry = 'registry:8443'
     iib_registry_credentials = 'iib:iibpassword'
+    iib_omps_url = 'https://omps-dev.cloud.paas.upshift.redhat.com/v2/{organization}/zipfile'
```
mprahl: We shouldn't store this in Git

mprahl: If this value is not set and we are handling a legacy operator, then an exception should be raised before the build starts saying that OMPS is not configured and fail the request with that reason.

iib/workers/tasks/build.py
```diff
@@ -428,7 +412,12 @@ def handle_add_request(bundles, binary_image, request_id, from_index=None, add_a
     :param list add_arches: the list of arches to build in addition to the arches ``from_index`` is
         currently built for; if ``from_index`` is ``None``, then this is used as the list of arches
         to build the index image for
-    :raises iib.exceptions.IIBError: if the index image build fails.
+    :param str cnr_token: the token required to push back ported packages to the legacy
```
mprahl:
```suggestion
    :param str cnr_token: the token required to push backported packages to the legacy
```

iib/workers/tasks/build.py
```diff
@@ -428,7 +412,12 @@ def handle_add_request(bundles, binary_image, request_id, from_index=None, add_a
...
+        app registry via OMPS.
+    :param str organization: organization name in the legacy app registry to which the back ported
```
mprahl:
```suggestion
    :param str organization: organization name in the legacy app registry to which the backported
```

iib/workers/tasks/build.py
```diff
@@ -124,19 +124,18 @@ def _create_and_push_manifest_list(request_id, arches):
     return output_pull_spec
 
 
-def _finish_request_post_build(request_id, arches):
+def _finish_request_post_build(output_pull_spec, request_id, arches):
     """
     Finish the request after the architecture builds have completed.
```
mprahl:
```suggestion
    Finish the request after the manifest list has been pushed.
```

iib/workers/tasks/legacy.py
mprahl: As discussed, we will fix this in another PR next week.

iib/workers/tasks/legacy.py
```diff
@@ -0,0 +1,226 @@
...
+from iib.workers.api_utils import get_request, set_request_state, update_request
+from iib.workers.config import get_worker_config
+from iib.workers.tasks.utils import run_cmd, skopeo_inspect
+
+log = logging.getLogger(__name__)
+
+
+def get_legacy_support_packages(bundles):
+    """
+    Checks if legacy support is required for any packages in the bundles.
```
mprahl:
```suggestion
    Check if legacy support is required for any packages in the bundles.
```

iib/workers/tasks/legacy.py
```diff
@@ -0,0 +1,226 @@
...
+    username, password = conf['iib_registry_credentials'].split(':')
+    run_cmd(
+        ['podman', 'login', '-u', username, '-p', password, conf['iib_registry']],
+        exc_msg=f'Failed to login into the registry {conf["iib_registry"]}',
+    )
+
+
+def _verify_package_info(package_dir, from_index):
+    """
+    Verifies if the exported package info is generated correctly
```
mprahl:
```suggestion
    Verify if the exported package info is generated correctly
```

iib/workers/tasks/legacy.py
mprahl: As discussed, let's remove this since OMPS does this for us and we can avoid adding more dependencies.

iib/workers/tasks/legacy.py
```diff
@@ -0,0 +1,226 @@
...
+                f'The exported package {package_name} is invalid: '
+                f'{", ".join(output.validation_dict["errors"])}'
+            )
+    except errors.OpCourierBadBundle as err:
+        raise IIBError(f'The exported package {package_name} is invalid: {err}')
+
+
+def _zip_package(package_dir):
+    """
+    Zip content of exported package to a `manifests.zip` file
```
mprahl:
```suggestion
    Zip content of exported package to a ``manifests.zip`` file
```

iib/workers/tasks/legacy.py
```diff
@@ -0,0 +1,226 @@
...
+
+    :param str package_dir: path to the exported package directory
+    :raises iib.exceptions.IIBError: if unable to zip the exported package
+    """
+    base_dir, package_name = package_dir.split('/download-', 1)
+    try:
+        shutil.make_archive(f'{base_dir}/manifests', 'zip', package_dir)
+    except Exception as err:
+        raise IIBError(f'Unable to zip exported package for {package_name}: {err}')
```
mprahl: Perform a `log.exception` here so that we can see the traceback, but not expose the gnarly error to the user.

mprahl:
```suggestion
        raise IIBError(f'Unable to zip exported package for {package_name}')
```

iib/workers/tasks/legacy.py
```diff
@@ -0,0 +1,226 @@
...
+    """
+    Zip content of exported package to a `manifests.zip` file
+
+    :param str package_dir: path to the exported package directory
+    :raises iib.exceptions.IIBError: if unable to zip the exported package
+    """
+    base_dir, package_name = package_dir.split('/download-', 1)
+    try:
+        shutil.make_archive(f'{base_dir}/manifests', 'zip', package_dir)
+    except Exception as err:
```
mprahl:
```suggestion
    except Exception:
```

iib/workers/tasks/legacy.py
```diff
@@ -0,0 +1,226 @@
...
+        raise IIBError(f'Unable to zip exported package for {package_name}: {err}')
+
+
+def _push_package_manifest(package_dir, cnr_token, organization):
+    """
+    Push `manifests.zip` file created for an exported package to OMPS
```
mprahl:
```suggestion
    Push `manifests.zip` file created for an exported package to OMPS
```
```suggestion
    Push ``manifests.zip`` file created for an exported package to OMPS
```

iib/workers/tasks/legacy.py
```diff
@@ -0,0 +1,226 @@
...
+
+    :param str package_dir: path to the exported package directory
```
mprahl:
```suggestion
    :param str package_dir: path to the exported package directory.
```

iib/workers/tasks/legacy.py
```diff
@@ -0,0 +1,226 @@
...
+    :param str cnr_token: the token required to push back ported packages to the legacy
```
mprahl:
```suggestion
    :param str cnr_token: the token required to push backported packages to the legacy
```

iib/workers/tasks/legacy.py
```diff
@@ -0,0 +1,226 @@
...
+        app registry via OMPS.
+    :param str organization: organization name in the legacy app registry to which the back ported
```
mprahl:
```suggestion
    :param str organization: the organization name in the legacy app registry to which the backported
```

iib/workers/tasks/legacy.py
```diff
@@ -0,0 +1,226 @@
...
+        packages should be pushed to.
+    :raises iib.exceptions.IIBError: if the push is unsucessful
+    """
+    conf = get_worker_config()
+    base_dir = package_dir.split('/download-', 1)[0]
+    # Taken from https://url.corp.redhat.com/pub-code
```
mprahl: Remove this line since it's not accessible to people outside the internal network

iib/workers/tasks/legacy.py
```diff
@@ -0,0 +1,226 @@
...
+        log.info('Files are %s', files)
+        resp = requests.post(
+            conf['iib_omps_url'].format(organization=organization),
+            headers={'Authorization': cnr_token},
+            files=files,
+        )
+        if not resp.ok:
+            log.error('Request to OMPS failed: %s', resp.text)
+            raise IIBError(f'Push to {organization} in the legacy app registry was unsucessful')
+        resp.raise_for_status()
```
mprahl: This is not needed since you raise an exception if it was unsuccessful

iib/workers/tasks/legacy.py
```diff
@@ -0,0 +1,226 @@
...
+
+
+def _get_rebuilt_index_image(request_id):
```
mprahl: Since this duplicates what's done in `_create_and_push_manifest_list`, I think you should either make this a public function that is used by both or just not bother with this private function at all since it's only called one time.

iib/workers/tasks/legacy.py
```diff
@@ -0,0 +1,226 @@
...
+
+
+def handle_legacy_content(
+    bundles, cnr_token, resolved_from_index, from_index, request_id, organization
+):
+    """
+    Handle exporting legacy content if required.
+
+    :param list bundles: a list of strings representing the bundles to be added to the index image.
+    :param str cnr_token: the token required to push back ported packages to the legacy
```
mprahl:
```suggestion
    :param str cnr_token: the token required to push backported packages to the legacy
```

iib/workers/tasks/legacy.py
mprahl:
```suggestion
    Get the packages that must be pushed to the legacy application registry.
```

iib/workers/tasks/legacy.py
```diff
@@ -0,0 +1,226 @@
...
+        user.
+    :param str from_index: the pull specification of the index image given in the request by the
+        user.
+    :param int request_id: the ID of the IIB build request
+    :param str organization: organization name in the legacy app registry to which the back ported
+        packages should be pushed to.
+    :raises iib.exceptions.IIBError: if legacy support is required and necessary params are missing.
+    """
+    legacy_support_packages = get_legacy_support_packages(bundles)
+    if legacy_support_packages and not all([cnr_token, organization]):
```
mprahl: This validation should occur before the new type index image builds are done. The validation should also check to see if the "iib_omps_url" is set.

iib/workers/tasks/legacy.py
```diff
@@ -0,0 +1,226 @@
...
+        packages_str = ', '.join(legacy_support_packages)
+        raise IIBError(
+            f'Legacy support is required for {packages_str};'
+            f' Both cnr_token and organization should be non-empty strings'
```
mprahl:
```suggestion
            ' Both cnr_token and organization should be non-empty strings'
```

iib/workers/tasks/legacy.py
```diff
@@ -0,0 +1,226 @@
...
+        )
+
+    if legacy_support_packages:
+        rebuilt_index_image = _get_rebuilt_index_image(request_id)
+        opm_index_export(
```
mprahl: After the validation is moved, all that is left is this function call. This could be called directly in `handle_add_request`.

iib/workers/tasks/legacy.py
```diff
@@ -0,0 +1,226 @@
...
+            legacy_support_packages, request_id, rebuilt_index_image, cnr_token, organization
+        )
+
+
+def opm_index_export(packages, request_id, rebuilt_index_image, cnr_token, organization):
+    """
+    Export packages to be back ported and push them via OMPS.
```
mprahl:
```suggestion
    Export packages to be backported and push them via OMPS.
```

iib/workers/tasks/legacy.py
```diff
@@ -0,0 +1,226 @@
...
+
+    :param list packages: a list of strings representing the names of the packages to be exported.
+    :param int request_id: the ID of the IIB build request
+    :param str rebuilt_index_image: the pull specification of the index image rebuilt by IIB.
+    :param str cnr_token: the token required to push back ported packages to the legacy
```
mprahl:
```suggestion
    :param str cnr_token: the token required to push backported packages to the legacy
```

iib/workers/tasks/legacy.py
```diff
@@ -0,0 +1,226 @@
...
+        app registry via OMPS.
+    :param str organization: organization name in the legacy app registry to which the back ported
```
mprahl:
```suggestion
    :param str organization: organization name in the legacy app registry to which the backported
```

iib/workers/tasks/legacy.py
```diff
@@ -0,0 +1,226 @@
...
+        packages should be pushed to.
+    :raises iib.exceptions.IIBError: if the export of packages fails.
+    """
+    request = get_request(request_id)
+    if request['state'] == 'failed':
+        set_request_state(
+            request_id, 'failed', f'Not exporting since the request has already failed',
```
mprahl:
```suggestion
            request_id, 'failed', 'Not exporting since the request has already failed',
```

iib/workers/tasks/legacy.py
```diff
@@ -0,0 +1,226 @@
...
+    :param int request_id: the ID of the IIB build request
+    :param str rebuilt_index_image: the pull specification of the index image rebuilt by IIB.
+    :param str cnr_token: the token required to push back ported packages to the legacy
+        app registry via OMPS.
+    :param str organization: organization name in the legacy app registry to which the back ported
+        packages should be pushed to.
+    :raises iib.exceptions.IIBError: if the export of packages fails.
+    """
+    request = get_request(request_id)
+    if request['state'] == 'failed':
```
mprahl: This is not needed since if it failed previous to this, it would have exited already.

iib/workers/tasks/legacy.py
```diff
@@ -0,0 +1,226 @@
...
+        opm_index_export(
+            legacy_support_packages, request_id, rebuilt_index_image, cnr_token, organization
+        )
+
+
+def opm_index_export(packages, request_id, rebuilt_index_image, cnr_token, organization):
+    """
+    Export packages to be back ported and push them via OMPS.
+
+    :param list packages: a list of strings representing the names of the packages to be exported.
```
mprahl:
```suggestion
    :param set packages: a list of strings representing the names of the packages to be exported.
```

iib/workers/tasks/legacy.py
```diff
@@ -0,0 +1,226 @@
...
+            cmd = [
+                'opm',
+                'index',
+                'export',
+                '--index',
+                rebuilt_index_image,
+                '--package',
+                package,
+                '--download-folder',
+                f'download-{package}',
```
mprahl: How about just `package` so you don't have to do string slicing?

iib/workers/tasks/legacy.py
```diff
@@ -0,0 +1,226 @@
...
+            ]
+
+            log.info('Generating the back ported operator for package: %s', package)
```
mprahl:
```suggestion
            log.info('Generating the backported operator for package: %s', package)
```

iib/workers/tasks/legacy.py
```diff
@@ -0,0 +1,226 @@
...
+
+            run_cmd(
+                cmd,
+                {'cwd': temp_dir},
+                exc_msg=f'Failed to export {package} pkg for index image {rebuilt_index_image}',
```
mprahl:
```suggestion
                exc_msg=f'Failed to push {package} to the legacy application registry',
```

iib/workers/tasks/legacy.py
```diff
@@ -0,0 +1,226 @@
...
+            package_dir = temp_dir.rstrip('/') + f'/download-{package}'
+            _verify_package_info(package_dir, rebuilt_index_image)
+            _zip_package(package_dir)
+            _push_package_manifest(package_dir, cnr_token, organization)
+
+    payload = {
+        'state': 'in_progress',
+        'state_reason': 'Back ported packages successfully pushed to OMPS',
+    }
+    update_request(
```
mprahl: Use `set_request_state` instead.

iib/workers/tasks/legacy.py
```diff
@@ -0,0 +1,208 @@
...
+    # opm does not fail when the package is missing in the index image, hence we
+    # check the number of generated files. If it's equal to 1, that means only an empty
+    # `package.yaml` file is generated and the package is missing
+    if len(os.listdir(package_dir)) == 1:
+        raise IIBError(f'package {package_name} is missing in index image {from_index}')
+
+
+def _get_base_dir_and_pkg_name(package_dir):
+    """
+    Get the base directory and package name from package directory
```
mprahl:
```suggestion
    Get the base directory and package name from package directory.
```

iib/workers/tasks/legacy.py
```diff
@@ -0,0 +1,208 @@
...
+    :param str package_dir: path to the exported package directory
+    :return: base_dir, package name
+    :rtype: str, str
+    """
+    return os.path.dirname(package_dir), os.path.basename(package_dir)
+
+
+def _zip_package(package_dir):
+    """
+    Zip content of exported package to a ``manifests.zip`` file
```
mprahl:
```suggestion
    Zip content of exported package to a ``manifests.zip`` file.
```

iib/workers/tasks/legacy.py
```diff
@@ -0,0 +1,208 @@
...
+
+    :param str package_dir: path to the exported package directory
+    :raises iib.exceptions.IIBError: if unable to zip the exported package
+    """
+    base_dir, package_name = _get_base_dir_and_pkg_name(package_dir)
+    try:
+        shutil.make_archive(f'{base_dir}/manifests', 'zip', package_dir)
+    except Exception:
+        log.exception('Unable to zip exported package: %s', err)
```
mprahl:
```suggestion
        log.exception('Unable to zip exported package: %s', package_name)
```

iib/workers/tasks/legacy.py
```diff
@@ -0,0 +1,208 @@
...
+         the backported packages should be pushed to.
+    :raises iib.exceptions.IIBError: if the push is unsucessful
+    """
+    conf = get_worker_config()
+    base_dir, _ = _get_base_dir_and_pkg_name(package_dir)
+    with open(f'{base_dir}/manifests.zip', 'rb') as fobj:
+        files = {'file': (fobj.name, fobj)}
+        log.info('Files are %s', files)
+        resp = requests.post(
+            conf['iib_omps_url'] + f'{organization}/zipfile',
```
mprahl:
```suggestion
            f'{conf["iib_omps_url"]}{organization}/zipfile',
```

iib/workers/tasks/legacy.py
```diff
@@ -0,0 +1,208 @@
...
+
+def validate_legacy_params_and_config(packages, bundles, cnr_token, organization):
+    """
+    Valiate parameters and config variables required for legacy support.
+
+    :param set packages: a set of strings representing the names of the packages to be exported.
+    :param list bundles: a list of strings representing the bundles to be added to the index image.
+    :param str cnr_token: the token required to push backported packages to the legacy
+        app registry via OMPS.
+    :param int request_id: the ID of the IIB build request
```
mprahl: This needs to be removed.

iib/workers/tasks/legacy.py
```diff
@@ -0,0 +1,208 @@
...
+            cmd = [
+                'opm',
+                'index',
+                'export',
+                '--index',
+                rebuilt_index_image,
+                '--package',
+                package,
+                '--download-folder',
+                f'{package}',
```
mprahl:
```suggestion
                package,
```

tests/test_workers/test_tasks/test_build.py
```diff
@@ -60,18 +60,15 @@ def test_create_and_push_manifest_list(mock_run_cmd, mock_td, tmp_path):
...
     output_pull_spec = 'quay.io/namespace/some-image:3'
-    mock_capml.return_value = output_pull_spec
     request_id = 2
     arches = {'amd64'}
-    build._finish_request_post_build(request_id, arches)
+    build._finish_request_post_build(output_pull_spec, request_id, arches)
 
-    mock_srs.assert_called_once()
-    mock_capml.assert_called_once_with(request_id, arches)
+    # mock_srs.assert_called_once()
```
mprahl: Remove this line

PR #21, opened 2020-02-19
Minor fixes

yashvardhannanavati:


mprahl:
Could you change the commit of `Trust rh certs in docker-compose` to `Use the host certificate bundle in the workers created by docker-compose`?

PR #29, opened 2020-02-24
Add the ability to specify required label values for a bundle

mprahl:
This is configured with the configuration `iib_required_labels`, where each key is an expected label and the value is the expected value of the label.

### DIFF
--- a/iib/workers/tasks/build.py
+++ b/iib/workers/tasks/build.py
@@ -418,6 +418,25 @@ def _verify_index_image(resolved_prebuild_from_index, unresolved_from_index):
         )
 
 
+def _verify_labels(bundles):
+    """
+    Verify that the required labels are set on the input bundles.
+
+    :param list bundles: a list of strings representing the pull specifications of the bundles to
+        add to the index image being built.
+    :raises iib.exceptions.IIBError: if one of the bundles does not have the correct label value.
+    """
+    conf = get_worker_config()
+    if not conf['iib_required_labels']:
+        return
+
+    for bundle in bundles:
+        labels = get_image_labels(bundle)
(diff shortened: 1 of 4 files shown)


iib/workers/tasks/build.py
```diff
@@ -418,6 +418,22 @@ def _verify_index_image(resolved_prebuild_from_index, unresolved_from_index):
...
+    """
+    Verify that the required labels are set on the input bundles.
+
+    :param list bundles: a list of strings representing the pull specifications of the bundles to
+        add to the index image being built.
+    :raises iib.exceptions.IIBError: if one of the bundles does not have the correct label value.
+    """
+    conf = get_worker_config()
+    for bundle in bundles:
+        labels = get_image_labels(bundle)
```
yashvardhannanavati: I think we are making more and more requests to the `get_image_labels` method which in-turn keeps calling skopeo inspect. Do you think it makes sense to store the labels for a bundle in the database?

yashvardhannanavati:
@mprahl The PR looks good. Just one comment, if you agree with it. Do you want to include that in this PR or file a new ticket?

iib/workers/tasks/build.py
mprahl: I agree it's a bit wasteful, but storing it in the database is more wasteful since the worker would have to make several API requests (patch and get) and we'd additionally have to store this data when we don't need to. If you'd like we could consider a cache such as the memory backend of dogpile cache:
https://dogpilecache.sqlalchemy.org/en/latest/api.html#memory-backends

I'd like to avoid doing that in this PR though.

mprahl:
@yashvardhannanavati I made a small change to the PR. Could you please review again?

PR #40, opened 2020-02-27
Only allow cnr_token on the add endpoint and verify all required parameters are non-empty

mprahl:


iib/web/models.py
```diff
@@ -440,6 +442,20 @@ def from_add_json(cls, kwargs):
...
 
+        for param in ('cnr_token', 'organization'):
+            if param not in request_kwargs:
+                continue
+
+            if not isinstance(request_kwargs[param], str):
+                raise ValidationError(f'"{param}" must be a string')
+
+            # If the value is empty, just remove it
+            if not request_kwargs[param]:
```
yashvardhannanavati: This check is technically not required

PR #47, opened 2020-03-05
Work around a podman bug that causes v1 manifests to be pushed

mprahl:


iib/workers/tasks/build.py
```diff
@@ -403,6 +403,19 @@ def _push_image(request_id, arch):
         exc_msg=f'Failed to push the index image to {destination} for the arch {arch}',
     )
 
+    log.debug(f'Verifying that {destination} was pushed as a v2 manifest')
```
yashvardhannanavati: We might want to add a comment here explaining why we are doing this with a link to the bugzilla ticket. WDYT?

PR #48, opened 2020-03-06
Allow IIB to specify a different registry to use for the index image

mprahl:
This will allow IIB to use the registry directly, but tell users to use the registry proxy. This makes it so that IIB is not reliant on the registry proxy.

### DIFF
--- a/iib/workers/tasks/build.py
+++ b/iib/workers/tasks/build.py
@@ -134,9 +134,22 @@ def _finish_request_post_build(output_pull_spec, request_id, arches):
     :param set arches: the set of arches that were built as part of this request
     :raises IIBError: if the manifest list couldn't be created and pushed
     """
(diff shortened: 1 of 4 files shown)


iib/workers/tasks/build.py
```diff
@@ -134,9 +134,17 @@ def _finish_request_post_build(output_pull_spec, request_id, arches):
     :param set arches: the set of arches that were built as part of this request
     :raises IIBError: if the manifest list couldn't be created and pushed
     """
+    conf = get_worker_config()
```
yashvardhannanavati: optional: We can move this to a separate helper method.

mprahl: This function is already quite small and the new code won't be used outside of this function, so I think it's okay as is.

PR #49, opened 2020-03-06
Add retry decorator for critical commands

yashvardhannanavati:


iib/workers/config.py
```diff
@@ -29,6 +29,7 @@ class Config(object):
     # Don't allow the worker to fetch more messages than it can handle at a time. This is so that
     # other tasks aren't starved. This will only be useful once more workers are enabled.
     worker_prefetch_multiplier = 1
+    worker_total_retries = 5
```
mprahl: The `worker_`  prefix is reserved for the Celery worker library configuration. Please use the `iib_` prefix and alphabetize it.

iib/workers/tasks/utils.py
```diff
@@ -25,6 +28,35 @@ def get_image_labels(pull_spec):
...
 
+def retry(
+    retries=worker_config.worker_total_retries, wait_on=Exception, logger=None,
+):
+    """Retry a section of code until success or max retries is reached."""
+
+    def wrapper(function):
+        @functools.wraps(function)
+        def inner(*args, **kwargs):
+            max_retries = retries
```
mprahl:
```suggestion
            remaining_attempts = retries
```

iib/workers/tasks/utils.py
```diff
@@ -25,6 +28,35 @@ def get_image_labels(pull_spec):
     return skopeo_inspect(full_pull_spec).get('Labels', {})
 
 
+def retry(
+    retries=worker_config.worker_total_retries, wait_on=Exception, logger=None,
+):
+    """Retry a section of code until success or max retries is reached."""
```
mprahl: Documenting the parameters would be nice.

iib/workers/tasks/utils.py
```diff
@@ -25,6 +28,35 @@ def get_image_labels(pull_spec):
...
+                except wait_on as e:
+                    max_retries -= 1
+                    if max_retries <= 0:
+                        if logger is not None:
+                            logger.exception(
+                                'The maximum number of retries (%s) have failed', retries
+                            )
+                        raise
+                    if logger is not None:
+                        logger.warning('Exception %r raised from %r.  Retrying now', e, function)
```
mprahl:
```suggestion
                        logger.warning('Exception %r raised from %r.  Retrying now', e, f'{function.__module__}.{function.__name__}')
```

iib/workers/tasks/utils.py
```diff
@@ -25,6 +28,35 @@ def get_image_labels(pull_spec):
     return skopeo_inspect(full_pull_spec).get('Labels', {})
 
 
+def retry(
+    retries=worker_config.worker_total_retries, wait_on=Exception, logger=None,
```
mprahl: If `retries` is 5, I would expect the function to be executed a maximum of 6 times (1 attempt + 5 retries). In the code below, it will only execute it 5 times.

Maybe you can change `retries` to `attempts` to make this less confusing.

iib/workers/config.py
mprahl: Could you please document this in the read me?

iib/workers/tasks/utils.py
mprahl: To avoid a global variable, you can do the following. `get_worker_config` will be executed only once when the module is imported this approach.

```suggestion
    retries=get_worker_config().worker_total_retries, wait_on=Exception, logger=None,
```

tests/test_workers/test_tasks/test_build.py
```diff
@@ -260,6 +260,29 @@ def test_push_image(mock_si, mock_run_cmd, mock_glps, schema_version):
         mock_run_cmd.assert_called_once()
 
 
+@mock.patch('iib.workers.tasks.build.run_cmd')
+def test_fix_schema_version_fail_retry_success(mock_run_cmd):
+
```
mprahl:
```suggestion
```

tests/test_workers/test_tasks/test_build.py
```diff
@@ -260,6 +260,29 @@ def test_push_image(mock_si, mock_run_cmd, mock_glps, schema_version):
...
+    build._fix_schema_version(destination)
+
+    skopeo_args = mock_run_cmd.mock_calls[1][1][0]
+    assert skopeo_args == ['skopeo', 'copy', '--format', 'v2s2', destination, destination]
+    assert mock_run_cmd.call_count == 2
+
+
+@mock.patch('iib.workers.tasks.build.run_cmd')
+def test_fix_schema_version_fail_max_retries(mock_run_cmd):
+
```
mprahl:
```suggestion
```

iib/workers/tasks/build.py
```diff
@@ -411,10 +428,7 @@ def _push_image(request_id, arch):
             'fixing it with skopeo.',
             destination,
         )
-        run_cmd(
-            ['skopeo', 'copy', '--format', 'v2s2', destination, destination],
-            exc_msg=f'Failed to fix the manifest schema version on {destination}',
-        )
+        _fix_schema_version(destination)
```
mprahl: Could you please change the `_push_image` test so that it mocks out `_fix_schema_version` and create a new test for `_fix_schema_version`? This makes it so that the test for `_push_image`  is not reliant on the way `_fix_schema_version` and it makes the tests less brittle and more independent.

tests/test_workers/test_tasks/test_build.py
```diff
@@ -260,6 +260,29 @@ def test_push_image(mock_si, mock_run_cmd, mock_glps, schema_version):
...
 
 
+@mock.patch('iib.workers.tasks.build.run_cmd')
+def test_fix_schema_version_fail_retry_success(mock_run_cmd):
+
+    mock_run_cmd.side_effect = [IIBError('Something went wrong'), None]
+    destination = 'some_destination'
+    build._fix_schema_version(destination)
+
+    skopeo_args = mock_run_cmd.mock_calls[1][1][0]
```
mprahl: This line and the one below can be removed since this is already tested elsewhere. You just need to test that `run_cmd` was called twice to test the retry.

tests/test_workers/test_tasks/test_build.py
```diff
@@ -260,6 +260,29 @@ def test_push_image(mock_si, mock_run_cmd, mock_glps, schema_version):
         mock_run_cmd.assert_called_once()
 
 
+@mock.patch('iib.workers.tasks.build.run_cmd')
+def test_fix_schema_version_fail_retry_success(mock_run_cmd):
```
mprahl: My preference is to test the retry decorator on its own rather than test individual functions that use the retry decorator. The reason is that once the `_fix_schema_version` function is removed, this test will get deleted along with it. It also makes the tests more independent.

For example (untested):
```python
def test_retry():
    mock_func = mock.Mock()

    @retry(retries=3, wait_on=IIBError)
    def _func_to_retry():
        mock_func()
        raise IIBError('Some error')

    with pytest.raises(IIBError, match='Some error'):
        _func_to_retry()

    assert mock_func.call_count == 4
```

tests/test_workers/test_tasks/test_build.py
yashvardhannanavati: Yup, but now that we have separate tests, the right place for these lines would be here right?

README.md
```diff
@@ -153,6 +153,8 @@ The custom configuration options for the Celery workers are listed below:
   `{registry}/iib-build:{request_id}`.
 * `iib_log_level` - the Python log level for `iib.workers` logger. This defaults to `INFO`.
 * `iib_registry` - the container registry to push images to (e.g. `quay.io`).
+* `iib_worker_total_attempts` - the total number of attempts to make at trying a function before erroring
```
mprahl:
```suggestion
* `iib_worker_total_attempts` - the total number of attempts to make at trying a function relating to the container registry before erroring
```

iib/workers/config.py
```diff
@@ -15,6 +15,7 @@ class Config(object):
     iib_image_push_template = '{registry}/iib-build:{request_id}'
     iib_log_level = 'INFO'
     iib_required_labels = {}
+    iib_worker_total_attempts = 5
```
mprahl:
```suggestion
    iib_total_attempts = 5
```

iib/workers/tasks/utils.py
```diff
@@ -1,10 +1,12 @@
 # SPDX-License-Identifier: GPL-3.0-or-later
+import functools
 import json
 import logging
 import subprocess
 
-from iib.exceptions import IIBError
```
mprahl: There is an extra new line now
```suggestion
```

iib/workers/tasks/utils.py
```diff
@@ -25,6 +27,46 @@ def get_image_labels(pull_spec):
...
 
 
+def retry(
+    attempts=get_worker_config().iib_worker_total_attempts, wait_on=Exception, logger=None,
+):
+    """
+    Retry a section of code until success or max retries is reached.
+
+    :param int attempts: the total number of attempts to make before erroring out
+    :param Exception wait_on: the exception on encountering which the function will be retired
```
mprahl:
```suggestion
    :param Exception wait_on: the exception on encountering which the function will be retried
```

iib/workers/tasks/utils.py
```diff
@@ -25,6 +27,46 @@ def get_image_labels(pull_spec):
...
+    :param logging logger: logger to log the messages on
```
mprahl:
```suggestion
    :param logging logger: the logger to log the messages on
```

iib/workers/tasks/utils.py
```diff
@@ -25,6 +27,46 @@ def get_image_labels(pull_spec):
...
+    :raises IIBError: if the maximum retries are reached
```
mprahl:
```suggestion
    :raises IIBError: if the maximum attempts are reached
```

tests/test_workers/test_tasks/test_build.py
```diff
@@ -254,12 +255,30 @@ def test_push_image(mock_si, mock_run_cmd, mock_glps, schema_version):
...
     if schema_version == 1:
-        skopeo_args = mock_run_cmd.mock_calls[1][1][0]
-        assert skopeo_args == ['skopeo', 'copy', '--format', 'v2s2', destination, destination]
+        mock_fsv.assert_called_once_with(destination)
     else:
         mock_run_cmd.assert_called_once()
 
 
+@mock.patch('iib.workers.tasks.build.run_cmd')
+def test_fix_schema_version_fail_retry_success(mock_run_cmd):
```
mprahl:
```suggestion
def test_fix_schema_version(mock_run_cmd):
```

tests/test_workers/test_tasks/test_utils.py
```diff
@@ -55,3 +55,17 @@ def test_skopeo_inspect(mock_run_cmd):
     skopeo_args = mock_run_cmd.call_args[0][0]
     expected = ['skopeo', 'inspect', image]
     assert skopeo_args == expected
+
+
+def test_retry():
```
mprahl: Optional: It's nice to keep the ordering of the tests the same as the ordering of the Python module (e.g. alphabetical).

tests/test_workers/test_tasks/test_build.py
```diff
@@ -254,12 +255,30 @@ def test_push_image(mock_si, mock_run_cmd, mock_glps, schema_version):
...
+    destination = 'some_destination'
+    build._fix_schema_version(destination)
+    skopeo_args = mock_run_cmd.mock_calls[0][1][0]
+    assert skopeo_args == ['skopeo', 'copy', '--format', 'v2s2', destination, destination]
+    mock_run_cmd.assert_called_once()
+
+
+@mock.patch('iib.workers.tasks.build.run_cmd')
+def test_fix_schema_version_fail_max_retries(mock_run_cmd):
```
mprahl: Optional: This test can be removed since you have `test_retry`.

yashvardhannanavati: yeah, but we would anyways want a failure test. So I thought might as well try the retry as well. Let's keep it.

tests/test_workers/test_tasks/test_utils.py
yashvardhannanavati: hehe, sure :)

README.md
```diff
@@ -153,6 +153,8 @@ The custom configuration options for the Celery workers are listed below:
   `{registry}/iib-build:{request_id}`.
 * `iib_log_level` - the Python log level for `iib.workers` logger. This defaults to `INFO`.
 * `iib_registry` - the container registry to push images to (e.g. `quay.io`).
+* `iib_worker_total_attempts` - the total number of attempts to make at trying a function relating to the
```
mprahl: Remember to alphabetize after renaming :smile: 
```suggestion
* `iib_total_attempts` - the total number of attempts to make at trying a function relating to the
```

PR #52, opened 2020-03-09
Add the opm fatal error message to a request's state_reason

mprahl:


iib/workers/tasks/utils.py
```diff
@@ -109,6 +111,15 @@ def run_cmd(cmd, params=None, exc_msg=None):
...
         log.error('The command "%s" failed with: %s', ' '.join(cmd), response.stderr)
-        raise IIBError(exc_msg or 'An unexpected error occurred')
+        if cmd[0] == 'opm':
+            # Caputre the fatal error
+            regex = r'^(?:.+level=fatal .+)(?:error=")(.+)(?:")$'
+            # Start from the last log message since that is often where the failure occurs
+            for msg in reversed(response.stderr.splitlines()):
+                match = re.match(regex, msg)
+                if match:
+                    raise IIBError(f'{exc_msg.rstrip(".")}. The error was: {match.groups()[0]}')
```
yashvardhannanavati: I think you can simply replace `The error was:` with `:`

mprahl: Good suggestion. I made the change.

PR #53, opened 2020-03-09
Surface legacy app registry error msg in state_reason

yashvardhannanavati:


iib/workers/tasks/legacy.py
```diff
@@ -111,7 +112,10 @@ def _push_package_manifest(package_dir, cnr_token, organization):
         )
         if not resp.ok:
             log.error('Request to OMPS failed: %s', resp.text)
-            raise IIBError(f'Push to {organization} in the legacy app registry was unsucessful')
+            error_msg = json.loads(resp.text).get('message', 'An unknown error occured')
```
mprahl: I don't think we should assume this is always JSON. I've seen many JSON APIs that return something else if the error is unexpected.

How about something like this?
```suggestion
            try:
                error_msg = resp.json().get('message', 'An unknown error occured')
            except json.JSONDecodeError:
                error_msg = resp.text
```

PR #56, opened 2020-03-26
Add the overwrite_from_index parameter

mprahl:


iib/web/api_v1.py
```diff
@@ -87,6 +112,7 @@ def add_bundles():
         payload.get('add_arches'),
         payload.get('cnr_token'),
         payload.get('organization'),
+        _should_force_overwrite() or payload.get('overwrite_from_index'),
```
lcarva: This seems odd. Won't this allow any authorized users to overwrite the from index image? This seems pretty dangerous.

lcarva: Nevermind. I see that this is done somewhere else.

PR #67, opened 2020-04-22
Add regenerate bundle request type

lcarva:
Introduce the new API endpoint /api/v1/builds/regenerate-bundle.
This is used to rebuild an operator bundle image.
In the future, modifications will be applied to the bundle image as part
of the build process.

Ticket-ID: CLOUDBLD-520

Signed-off-by: Luiz Carvalho <lucarval@redhat.com>

PR #68, opened 2020-04-24
Add the concept of batching requests

mprahl:
This introduces the concept of batches in the database and REST API. This paves the way for a future change that will allow a user to submit multiple requests at the same time and keep track of them as a batch.

All existing requests will be individually assigned to a new batch as part of the database migration. All new requests will be individually assigned to a new batch.

A separate commit also fixes a deprecation warning in one of the migration scripts.

### DIFF
--- a/iib/web/api_v1.py
+++ b/iib/web/api_v1.py
@@ -10,6 +10,7 @@ from iib.exceptions import ValidationError
 from iib.web import db
 from iib.web.models import (
     Architecture,
+    Batch,
     Image,
     Operator,
     Request,
@@ -53,6 +54,7 @@ def get_builds():
 
     :rtype: flask.Response
     """
+    batch_id = flask.request.args.get('batch')
     state = flask.request.args.get('state')
     verbose = str_to_bool(flask.request.args.get('verbose'))
     max_per_page = flask.current_app.config['IIB_MAX_PER_PAGE']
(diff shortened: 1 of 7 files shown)


mprahl:
Please note that I will rebase and fix the conflicts after #67 is merged.

yashvardhannanavati:
@mprahl could you please rebase this PR?

mprahl:
@yashvardhannanavati rebased.

yashvardhannanavati:
@mprahl the changes look great! Can I ask what would stop me from accidentally messing up someone else's batches? And like from a user perspective, how do I know what batch number can I use because I see it is a part of the request_kwargs right...Am I missing something?

mprahl:
@yashvardhannanavati the user can't influence the batch in anyway as part of this PR. The build request is just automatically assigned to a new batch.

In the next PR for this, there will be an API endpoint that allows you to submit multiple requests at a time. In this case, IIB would assign all these requests as part of a new batch.

lcarva:
> @mprahl the changes look great! Can I ask what would stop me from accidentally messing up someone else's batches? And like from a user perspective, how do I know what batch number can I use because I see it is a part of the request_kwargs right...Am I missing something?

To add to what @mprahl has said, the user can provide a batch ID, but that's only used for filtering results on the `/builds` API endpoint. The assignment of a batch number is completely done by IIB.

yashvardhannanavati:
Ah, now I get it. Sorry for the confusion. Thanks guys!

PR #70, opened 2020-04-27
Pin to digest during regenerate bundle images

lcarva:
Ticket-ID: CLOUDBLD-523

### DIFF
--- a/iib/workers/tasks/build.py
+++ b/iib/workers/tasks/build.py
@@ -4,6 +4,8 @@ import os
 import tempfile
 import textwrap
 
+from operator_manifest.operator import ImageName, OperatorManifest
+
 from iib.exceptions import IIBError
 from iib.workers.api_utils import set_request_state, update_request
 from iib.workers.config import get_worker_config
@@ -677,14 +679,16 @@ def handle_regenerate_bundle_request(from_bundle_image, organization, request_id
     update_request(request_id, payload, exc_msg=exc_msg)
 
     with tempfile.TemporaryDirectory(prefix='iib-') as temp_dir:
-        # TODO: Pin related images to digests
-        # TODO: Apply organization specific modifications
+        manifests_path = os.path.join(temp_dir, 'manifests')
+        _copy_files_from_image(from_bundle_image_resolved, '/manifests', manifests_path)
+        _adjust_operator_manifests(manifests_path)
 
         with open(os.path.join(temp_dir, 'Dockerfile'), 'w') as dockerfile:
             dockerfile.write(
                 textwrap.dedent(
                     f"""\
                         FROM {from_bundle_image_resolved}
+                        COPY ./manifests /manifests
                     """
                 )
             )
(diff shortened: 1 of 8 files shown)


lcarva:
Hmm travis doesn't like the direct links for the dependencies. I might need to publish operator-metadata to pypi.

iib/workers/tasks/build.py
```diff
@@ -703,3 +707,68 @@ def handle_regenerate_bundle_request(from_bundle_image, organization, request_id
...
+        )
+    finally:
+        try:
+            run_cmd(
+                ['podman', 'rm', container_id],
+                exc_msg=f'Failed to remove the container {container_id} for image {image}',
+            )
+        except IIBError as e:
+            # Failure to remove the temporary container shouldn't cause the IIB request to fail.
+            log.warning(e)
```
mprahl, line 745: Optional: Perhaps `log.exception` would be better to get the traceback

iib/workers/tasks/build.py
```diff
@@ -703,3 +707,68 @@ def handle_regenerate_bundle_request(from_bundle_image, organization, request_id
         'state_reason': 'The request completed successfully',
     }
     update_request(request_id, payload, exc_msg='Failed setting the bundle image on the request')
+
+
+def _copy_files_from_image(image, src_path, dest_path):
```
yashvardhannanavati, line 712: Do we want to add the `@retry` decorator here?

yashvardhannanavati: That might mean splitting the method though

setup.py
```diff
@@ -1,6 +1,7 @@
 # SPDX-License-Identifier: GPL-3.0-or-later
 from setuptools import setup, find_packages
 
+
```
yashvardhannanavati, line 4: Can I ask if this was intentional?

lcarva: I thought it would be required by black/flake8. It doesn't seem to care. I'll revert this.

iib/workers/tasks/build.py
lcarva: It might be tricky to distinguish whether or not retry should be performed. For instance, if the image provided doesn't have the `/manifests` directory, it would fail but we shouldn't retry.

Can we leave it as is and address this in the future if it does become a problem?

iib/workers/tasks/build.py
lcarva: Good idea. I changed it to `log.exception`. I left `e` as a parameter because that already has the error message we care about.

lcarva:
@yashvardhannanavati, I have addressed your comments.

iib/workers/tasks/build.py
mprahl: Since the image is not pulled down at this point, it's possible that this would fail, especially if the image is pulled through the registry proxy. Perhaps you could just pull down the image beforehand using retries and then keep this function as is?

iib/workers/tasks/build.py
```diff
@@ -703,3 +707,68 @@ def handle_regenerate_bundle_request(from_bundle_image, organization, request_id
...
+        )
+    finally:
+        try:
+            run_cmd(
+                ['podman', 'rm', container_id],
+                exc_msg=f'Failed to remove the container {container_id} for image {image}',
+            )
+        except IIBError as e:
+            # Failure to remove the temporary container shouldn't cause the IIB request to fail.
+            log.exception(e)
```
mprahl, line 745:
```suggestion
            log.exception('Failed to remove the container %s', container_id)
```

iib/workers/tasks/build.py
yashvardhannanavati: Cool, I can live with that. :)

iib/workers/tasks/build.py
lcarva: The exception already has this.

mprahl: Okay, then just do `log.exception()` then.

PR #74, opened 2020-05-01
Fill in spec.relatedImages when regenerating bundles

mprahl:
This also keeps IIB from pinning pull specifications when spec.relatedImages is already set. A request will fail if spec.relatedImages and the RELATED_IMAGE_* environment variables are set in any of containers in the ClusterServiceVersion file.

Resolves CLOUDBLD-524.

### DIFF
--- a/README.md
+++ b/README.md
@@ -216,8 +216,15 @@ The custom configuration options for the Celery workers are listed below:
 In addition to building operator index images, IIB can also be used to regenerate operator bundle
 images. This is useful for applying modifications to the manifests embedded in the bundle image.
 IIB uses the [operator-manifest](https://github.com/containerbuildsystem/operator-manifest) library
-to assist in these modifications. Currently, IIB will pin any container image pull specification
-to its corresponding digest. See the different
+to assist in these modifications.
+
+Currently, IIB will not perform any modifications on a ClusterServiceVersion file if
+[spec.relatedImages](https://access.redhat.com/documentation/en-us/openshift_container_platform/4.3/html-single/operators/index#olm-enabling-operator-for-restricted-network_osdk-generating-csvs)
+is set.
+
+If it's not set, IIB will pin any container image pull specification and set
+[spec.relatedImages](https://access.redhat.com/documentation/en-us/openshift_container_platform/4.3/html-single/operators/index#olm-enabling-operator-for-restricted-network_osdk-generating-csvs).
+See the different
 [pull specifications](https://github.com/containerbuildsystem/operator-manifest#pull-specifications)
 to which this process applies to.
 
--- a/iib/workers/tasks/build.py
+++ b/iib/workers/tasks/build.py
@@ -755,25 +755,54 @@ def _adjust_operator_manifests(manifests_path):
     For any container image pull spec found in the Operator CSV files, replace floating
     tags with pinned digests, e.g. `image:latest` becomes `image@sha256:...`.
 
+    If spec.relatedImages is not set, it will be set with the pinned digests. If it is set but
+    there are also RELATED_IMAGE_* environment variables set, an exception will be raised.
+
     This method relies on the OperatorManifest class to properly identify and apply the
     modifications as needed.
 
     :param str manifests_path: the full path to the directory containing the operator manifests.
+    :raises IIBError: if the operator manifest has invalid entries
     """
     operator_manifest = OperatorManifest.from_directory(manifests_path)
     found_pullspecs = set()
+    operator_csvs = []
     for operator_csv in operator_manifest.files:
+        if operator_csv.has_related_images():
+            csv_file_name = os.path.basename(operator_csv.path)
+            if operator_csv.has_related_image_envs():
+                raise IIBError(
+                    f'The ClusterServiceVersion file {csv_file_name} has entries in '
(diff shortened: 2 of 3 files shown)


iib/workers/tasks/build.py
```diff
@@ -755,25 +755,46 @@ def _adjust_operator_manifests(manifests_path):
...
-        pullspec: ImageName.parse(_get_resolved_image(pullspec)) for pullspec in found_pullspecs
-    }
+    replacement_pullspecs = {}
+    for pullspec in found_pullspecs:
+        # Skip images that are already pinned
+        if ':' not in ImageName.parse(pullspec).tag:
+            replacement_pullspecs[pullspec] = ImageName.parse(_get_resolved_image(pullspec))
 
     # Apply modifications to the operator bundle image metadata
     for operator_csv in operator_manifest.files:
```
lcarva, line 791: If it has relatedImages block populated, pinning should also not happen. This method should be a noop in that case. This wasn't well defined at the time the code was written.

iib/workers/tasks/build.py
```diff
@@ -755,23 +755,55 @@ def _adjust_operator_manifests(manifests_path):
...
+            replacement_pullspecs[pullspec] = ImageName.parse(_get_resolved_image(pullspec))
 
     # Apply modifications to the operator bundle image metadata
-    for operator_csv in operator_manifest.files:
+    for operator_csv in operator_csvs:
+        csv_file_name = os.path.basename(operator_csv.path)
+        log.info('Pinning the pull specifications on %s', csv_file_name)
         operator_csv.replace_pullspecs_everywhere(replacement_pullspecs)
+
+        if not operator_csv.has_related_images():
```
lcarva, line 805: Is this necessary? If there are related images, the the csv won't be added to `operator_csvs`, right? 
This conditional sort of indicates that pinning and setting of related images may happen independently which is definitely something we're trying to avoid.

README.md
```diff
@@ -216,10 +216,16 @@ The custom configuration options for the Celery workers are listed below:
 In addition to building operator index images, IIB can also be used to regenerate operator bundle
 images. This is useful for applying modifications to the manifests embedded in the bundle image.
 IIB uses the [operator-manifest](https://github.com/containerbuildsystem/operator-manifest) library
-to assist in these modifications. Currently, IIB will pin any container image pull specification
-to its corresponding digest. See the different
+to assist in these modifications.
+
+Currently, IIB will pin any container image pull specification
+to its corresponding digest if
```
yashvardhannanavati, line 222: can we please reword this sentence. It caused a bit of confusion when read together with `IIB also sets
[spec.relatedImages](https://access.redhat.com/documentation/en-us/openshift_container_platform/4.3/html-single/operators/index#olm-enabling-operator-for-restricted-network_osdk-generating-csvs)
when it isn't already set.`

yashvardhannanavati: Or did you mean to end this statement with `is not set` => `is set`?

iib/workers/tasks/build.py
```diff
@@ -755,25 +755,55 @@ def _adjust_operator_manifests(manifests_path):
...
     """
     operator_manifest = OperatorManifest.from_directory(manifests_path)
     found_pullspecs = set()
+    operator_csvs = []
     for operator_csv in operator_manifest.files:
+        if operator_csv.has_related_images():
+            csv_file_name = os.path.basename(operator_csv.path)
+            if operator_csv.has_related_image_envs():
+                raise IIBError(
+                    f'The ClusterServiceVersion file {csv_file_name} has entries in '
```
yashvardhannanavati, line 775: Just out of curiosity, can I ask why is that we raise this error?

iib/workers/tasks/build.py
```diff
@@ -755,25 +755,55 @@ def _adjust_operator_manifests(manifests_path):
...
+                    'spec.relatedImages and one or more containers have RELATED_IMAGE_* '
+                    'environment variables set. This is not allowed for bundles regenerated with '
+                    'IIB.'
+                )
+            log.debug(
+                'Skipping pinning since the ClusterServiceVersion file %s has entries in '
```
yashvardhannanavati, line 781: mprahl++ for using the entire name of the CSV file everywhere. It is very confusing with just the acronym! :D

yashvardhannanavati: Should this log statement state that you are skipping pinning now and will do it later for these CSV's if they are not pinned to their digest? Am I missing something?

README.md
mprahl: Well IIB only pins the pull specifications if `spec.relatedImages` isn't set and it only sets `spec.relatedImages` when it's not already set. I'll make this clearer.

iib/workers/tasks/build.py
mprahl: It was a policy decision to limit confusing behavior of setting it in multiple ways.

iib/workers/tasks/build.py
mprahl: Correct. I forgot to update this part of the code after my last force push. Thanks for catching it.

iib/workers/tasks/build.py
mprahl: Based on what @lcarva said, if `spec.relatedImages` is set, then don't do any changes to the CSV file.

PR #75, opened 2020-05-01
Allow overwrite for unprivileged users

lcarva:


iib/workers/tasks/build.py
```diff
@@ -484,6 +489,10 @@ def _skopeo_copy(source, destination, copy_all=False, exc_msg=None):
     ]
     if copy_all:
         args.append('--all')
+    if dest_token:
+        log.debug('Using user-provided token to copy the container image')
+        args.append('--dest-creds')
+        args.append(dest_token)
```
mprahl, line 495: `run_cmd` will need to be adjusted to not log the credentials in case it fails. It'll be similar to this:
https://github.com/release-engineering/iib/commit/41d101c08edd132a9a096857016049f0e0f61a93#diff-1c6ad2472cabab850bfc0c408419ccaaL74-L80

iib/workers/tasks/build.py
```diff
@@ -626,6 +650,9 @@ def handle_rm_request(
         currently built for.
     :param bool overwrite_from_index: if True, overwrite the input ``from_index`` with the built
         index image.
+    :param str overwrite_from_index_token: the token used for overwriting the input
+        from_index image. This is required for non-privileged users to use overwrite_from_index.
```
mprahl, line 654:
```suggestion
        ``from_index`` image. This is required for non-privileged users to use ``overwrite_from_index``.
```

iib/workers/tasks/build.py
```diff
@@ -439,24 +445,20 @@ def _push_image(request_id, arch):
...
+    _skopeo_copy(
+        source,
+        destination,
         exc_msg=f'Failed to push the container image to {destination} for the arch {arch}',
     )
 
-    log.debug(f'Verifying that {destination} was pushed as a v2 manifest due to RHBZ#1810768')
+    log.debug(f'Verifying that {destination} was pushed as a v2 manifest')
     skopeo_raw = skopeo_inspect(destination, '--raw')
     if skopeo_raw['schemaVersion'] != 2:
```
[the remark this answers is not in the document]
lcarva: I don't think it will fallback if we specify the format. I think the check is needed to help us debug unexpected issues. If it's not schema version 2, then it'll fail later when creating the manifest list but it'll be much harder to debug it then. I'll add a comment explaining this.

iib/workers/config.py
```diff
@@ -16,7 +16,7 @@ class Config(object):
     iib_index_image_output_registry = None
     iib_log_level = 'INFO'
     iib_required_labels = {}
-    iib_skopeo_timeout = '30s'
+    iib_skopeo_timeout = '60s'
```
yashvardhannanavati, line 19: Is there a specific reason to do this? If yes, how about bumping this up to 300s based on the discussion in CLOUDDST-472

iib/workers/tasks/build.py
```diff
@@ -439,34 +448,32 @@ def _push_image(request_id, arch):
...
-def _skopeo_copy(source, destination, copy_all=False, exc_msg=None):
+def _skopeo_copy(source, destination, copy_all=False, dest_token=None, exc_msg=None):
     """
     Wrap the ``skopeo copy`` command.
 
     :param str source: the source to copy
     :param str destination: the destination to copy the source to
     :param bool copy_all: if True, it passes ``--all`` to the command
+    :param str dest_token: the token to pass to the ``--dest-token` parameter of the command.
+        If not provided, ``--dest-token`` parameter is also not provided.
```
yashvardhannanavati, line 476: Should this be ``--dest-token`` parameter is also not provided. ==> ``--dest-authfile`` parameter is also not provided.

iib/workers/tasks/build.py
```diff
@@ -482,6 +489,10 @@ def _skopeo_copy(source, destination, copy_all=False, exc_msg=None):
     ]
     if copy_all:
         args.append('--all')
+    if dest_token:
```
yashvardhannanavati, line 492: Should this section also add a dummy `--dest-authfile`?

iib/workers/config.py
lcarva: I did this because skopeo is now used to push the image which can take longer. I think 300s is a valid default. I'll change to use it.

iib/workers/tasks/build.py
lcarva: Great catch! I implemented something similar to celery's argsrepr.

iib/workers/config.py
lcarva: Done.

iib/workers/tasks/build.py
lcarva: The latest released version of skopeo doesn't have support for `--dest-authfile` :disappointed:

iib/workers/tasks/build.py
```diff
@@ -439,40 +448,41 @@ def _push_image(request_id, arch):
...
+    # Use "skopeo copy" because "podman push" doesn't support setting the format
+    _skopeo_copy(
+        source,
+        destination,
         exc_msg=f'Failed to push the container image to {destination} for the arch {arch}',
     )
 
-    log.debug(f'Verifying that {destination} was pushed as a v2 manifest due to RHBZ#1810768')
+    # Skopeo should've copied the image as v2 schema 2. Let's verify that this actually
+    # happened to avoid ambiguous errors later in the process that may be diffucult to
```
mprahl, line 462:
```suggestion
    # happened to avoid ambiguous errors later in the process that may be difficult to
```

iib/workers/tasks/utils.py
```diff
@@ -101,13 +101,16 @@ def podman_pull(*args):
...
-def run_cmd(cmd, params=None, exc_msg=None):
+def run_cmd(cmd, params=None, exc_msg=None, cmd_repr=None):
     """
     Run the given command with the provided parameters.
 
     :param iter cmd: iterable representing the command to be executed
     :param dict params: keyword parameters for command execution
     :param str exc_msg: an optional exception message when the command fails
+    :param iter cmd_repr: an optional representation of the command to be executed. If not
+        provided, this is derived from the cmd parameter. This is useful when the command
```
mprahl, line 112:
```suggestion
        provided, this is derived from the ``cmd`` parameter. This is useful when the command
```

iib/workers/tasks/utils.py
```diff
@@ -120,11 +123,13 @@ def run_cmd(cmd, params=None, exc_msg=None):
     params.setdefault('stderr', subprocess.PIPE)
     params.setdefault('stdout', subprocess.PIPE)
 
-    log.debug('Running the command "%s"', cmd)
+    cmd_repr = cmd_repr or cmd
```
mprahl, line 126: Could you move this to the beginning of the function? I like to avoid reassigning argument values, but if it's needed, then I like it at the top of the function to make it clear that the arguments don't change after that point.

tests/test_web/test_api_v1.py
```diff
@@ -1,5 +1,6 @@
 # SPDX-License-Identifier: GPL-3.0-or-later
 from unittest import mock
+import json
```
mprahl, line 3:
```suggestion
import json
from unittest import mock
```

tests/test_workers/test_tasks/test_utils.py
```diff
@@ -1,5 +1,6 @@
 # SPDX-License-Identifier: GPL-3.0-or-later
 from unittest import mock
+import logging
```
mprahl, line 3:
```suggestion
import logging
from unittest import mock
```

tests/test_workers/test_tasks/test_utils.py
```diff
@@ -39,6 +40,18 @@ def test_run_cmd(mock_sub_run):
...
+@mock.patch('iib.workers.tasks.utils.subprocess.run')
+def test_run_cmd_with_cmd_repr(mock_sub_run, caplog):
+    mock_sub_run.return_value = mock.Mock(returncode=0)
+
+    secret = 'top-secret'
+    with caplog.at_level(logging.DEBUG):
+        utils.run_cmd(['echo', secret], cmd_repr=['echo', '*****'])
+
+    mock_sub_run.assert_called_once()
+    assert secret not in caplog.text
```
mprahl, line 52: Please make sure the logger is enabled:
```suggestion
    assert '*****' in caplog.text
    assert secret not in caplog.text
```

In the past, I had issues where Flask was modifying the loggers as part of the test. I needed to do this ugly workaround to make it work:
```python
    for logger in ('iib.workers', 'iib.workers.tasks', 'iib.workers.tasks.utils'):
        logging.getLogger(logger).disabled = False
```

tests/test_web/test_api_v1.py
lcarva: Oh `import` comes first? I always thought it was the other way around.

tests/test_workers/test_tasks/test_utils.py
lcarva: You're right. The additional assert highlights that this test is not reliable. It works when running on its own, but the logs are empty when running the whole test suite. Flask must be interfering as you mentioned. I tried your suggestion, but that doesn't seem to do it. I also added `"iib"` to the tuple of loggers. I'll keep digging.

lcarva: Ah! The trick is to also set the level. It looks like doing this at only the more specific logger does the trick.

iib/workers/tasks/build.py
```diff
@@ -567,6 +581,9 @@ def handle_add_request(
         packages should be pushed to.
     :param bool overwrite_from_index: if True, overwrite the input ``from_index`` with the built
         index image.
+    :param str overwrite_from_index_token: the token used for overwriting the input
+        from_index image. This is required for non-privileged users to use overwrite_from_index.
```
mprahl, line 585:
```suggestion
        ``from_index image``. This is required for non-privileged users to use ``overwrite_from_index``.
```

iib/workers/tasks/build.py
```diff
@@ -624,6 +652,9 @@ def handle_rm_request(
         currently built for.
     :param bool overwrite_from_index: if True, overwrite the input ``from_index`` with the built
         index image.
+    :param str overwrite_from_index_token: the token used for overwriting the input
+        ``from_index`` image. This is required for non-privileged users to use overwrite_from_index.
```
mprahl, line 656:
```suggestion
        ``from_index`` image. This is required for non-privileged users to use ``overwrite_from_index``.
```

tests/test_web/test_api_v1.py
mprahl: We just keep it alphabetized. We don't have a convention of putting `from` or `import` in a separate order.

tests/test_workers/test_tasks/test_utils.py
mprahl: @lcarva did use that for loop before the test?

lcarva:
I think I have addressed all the comment. I have also re-tested the add and rm operations on my dev environment with this functionality. All seems to be working as expected.

PR #76, opened 2020-05-04
Overwrite from_index before the OMPS push [CLOUDDST-114]

chandwanitulsi:
Earlier if overwrite_from_index option was specified,
the index image would be overridden after the OMPS push.
Fix this as overriding should happen before and
the OMPS push should get the latest pull spec of the index image.
Also so that the OMPS push can be ignored if index override
raised any errors

### DIFF
--- a/iib/workers/tasks/build.py
+++ b/iib/workers/tasks/build.py
@@ -135,7 +135,7 @@ def _create_and_push_manifest_list(request_id, arches):
     return output_pull_spec
 
 
-def _finish_request_post_build(
+def _update_index_image_pull_spec(
     output_pull_spec,
     request_id,
     arches,
@@ -144,7 +144,7 @@ def _finish_request_post_build(
     overwrite_from_index_token=None,
 ):
     """
-    Finish the request after the manifest list has been pushed.
+    Update the request with the modified index image.
 
     This function was created so that code didn't need to be duplicated for the ``add`` and ``rm``
     request types.
@@ -162,10 +162,13 @@ def _finish_request_post_build(
     """
     conf = get_worker_config()
     if from_index and overwrite_from_index:
-        log.info(f'Ovewriting the index image {from_index} with {output_pull_spec}')
+        output_message = f'Ovewriting the index image {from_index} with {output_pull_spec}'
+        log.info(output_message)
         index_image = from_index
         exc_msg = f'Failed to overwrite the input from_index container image of {index_image}'
         args = [f'docker://{output_pull_spec}', f'docker://{index_image}']
+        state_reason = output_message
+        set_request_state(request_id, 'in_progress', state_reason)
         _skopeo_copy(*args, copy_all=True, dest_token=overwrite_from_index_token, exc_msg=exc_msg)
     elif conf['iib_index_image_output_registry']:
         index_image = output_pull_spec.replace(
@@ -182,8 +185,6 @@ def _finish_request_post_build(
     payload = {
         'arches': list(arches),
         'index_image': index_image,
-        'state': 'complete',
-        'state_reason': 'The request completed successfully',
     }
     update_request(request_id, payload, exc_msg='Failed setting the index image on the request')
 
@@ -670,19 +671,20 @@ def handle_add_request(
 
     set_request_state(request_id, 'in_progress', 'Creating the manifest list')
     output_pull_spec = _create_and_push_manifest_list(request_id, arches)
-
-    if legacy_support_packages:
-        export_legacy_packages(
-            legacy_support_packages, request_id, output_pull_spec, cnr_token, organization
-        )
-
-    _finish_request_post_build(
+    _update_index_image_pull_spec(
         output_pull_spec,
         request_id,
         arches,
         from_index,
         overwrite_from_index,
         overwrite_from_index_token,
+    )
+    if legacy_support_packages:
+        export_legacy_packages(
+            legacy_support_packages, request_id, output_pull_spec, cnr_token, organization
+        )
+    set_request_state(
+        request_id, 'complete', 'The operator bundle(s) were successfully added to the index image',
     )
 
 
@@ -731,13 +733,16 @@ def handle_rm_request(
     set_request_state(request_id, 'in_progress', 'Creating the manifest list')
     output_pull_spec = _create_and_push_manifest_list(request_id, arches)
 
-    _finish_request_post_build(
+    _update_index_image_pull_spec(
         output_pull_spec,
         request_id,
         arches,
         from_index,
         overwrite_from_index,
         overwrite_from_index_token,
+    )
+    set_request_state(
+        request_id, 'complete', 'The operator(s) were successfully removed from the index image',
     )
 
 
--- a/iib/workers/tasks/build.py
+++ b/iib/workers/tasks/build.py
@@ -162,7 +162,7 @@ def _update_index_image_pull_spec(
     """
     conf = get_worker_config()
     if from_index and overwrite_from_index:
-        output_message = f'Ovewriting the index image {from_index} with {output_pull_spec}'
+        output_message = f'Overwriting the index image {from_index} with {output_pull_spec}'
         log.info(output_message)
         index_image = from_index
         exc_msg = f'Failed to overwrite the input from_index container image of {index_image}'
--- a/tests/test_workers/test_tasks/test_build.py
+++ b/tests/test_workers/test_tasks/test_build.py
@@ -82,7 +82,9 @@ def test_create_and_push_manifest_list(mock_run_cmd, mock_td, tmp_path):
 @mock.patch('iib.workers.tasks.build.get_worker_config')
 @mock.patch('iib.workers.tasks.build._skopeo_copy')
 @mock.patch('iib.workers.tasks.build.update_request')
-def test_finish_request_post_build(
+@mock.patch('iib.workers.tasks.build.set_request_state')
+def test_update_index_image_pull_spec(
+    mock_srs,
     mock_ur,
     mock_sc,
     mock_gwc,
@@ -101,13 +103,13 @@ def test_finish_request_post_build(
         'iib_index_image_output_registry': iib_index_image_output_registry,
         'iib_registry': 'quay.io',
     }
-    build._finish_request_post_build(
+    build._update_index_image_pull_spec(
         default, request_id, arches, from_index, overwrite, overwrite_token
     )
 
     mock_ur.assert_called_once()
     update_request_payload = mock_ur.call_args[0][1]
-    assert update_request_payload.keys() == {'arches', 'index_image', 'state', 'state_reason'}
+    assert update_request_payload.keys() == {'arches', 'index_image'}
     assert update_request_payload['index_image'] == expected_pull_spec
     if overwrite_token:
         mock_sc.assert_called_once_with(
@@ -125,8 +127,10 @@ def test_finish_request_post_build(
             dest_token=None,
             exc_msg=mock.ANY,
         )
+        mock_srs.assert_called_once()
     else:
         mock_sc.assert_not_called()
+        mock_srs.assert_not_called()
 
 
 @pytest.mark.parametrize('request_id, arch', ((1, 'amd64'), (5, 's390x')))
@@ -421,7 +425,7 @@ def test_skopeo_copy_fail_max_retries(mock_run_cmd):
 @mock.patch('iib.workers.tasks.build._build_image')
 @mock.patch('iib.workers.tasks.build._push_image')
 @mock.patch('iib.workers.tasks.build._verify_index_image')
-@mock.patch('iib.workers.tasks.build._finish_request_post_build')
+@mock.patch('iib.workers.tasks.build._update_index_image_pull_spec')
 @mock.patch('iib.workers.tasks.build.export_legacy_packages')
 @mock.patch('iib.workers.tasks.build.set_request_state')
 @mock.patch('iib.workers.tasks.build._create_and_push_manifest_list')
@@ -437,7 +441,7 @@ def test_handle_add_request(
     mock_capml,
     mock_srs,
     mock_elp,
-    mock_frpb,
+    mock_uiips,
     mock_vii,
     mock_pi,
     mock_bi,
@@ -493,10 +497,10 @@ def test_handle_add_request(
     assert cnr_token in export_args
     assert organization in export_args
 
-    mock_frpb.assert_called_once()
+    mock_uiips.assert_called_once()
     mock_vii.assert_called_once()
     mock_capml.assert_called_once()
-    assert mock_srs.call_count == 2
+    assert mock_srs.call_count == 3
 
 
 @mock.patch('iib.workers.tasks.build.set_request_state')
@@ -563,9 +567,9 @@ def test_handle_add_request_bundle_resolution_failure(mock_grb, mock_srs):
 @mock.patch('iib.workers.tasks.build._verify_index_image')
 @mock.patch('iib.workers.tasks.build.set_request_state')
 @mock.patch('iib.workers.tasks.build._create_and_push_manifest_list')
-@mock.patch('iib.workers.tasks.build._finish_request_post_build')
+@mock.patch('iib.workers.tasks.build._update_index_image_pull_spec')
 def test_handle_rm_request(
-    mock_frpb, mock_capml, mock_srs, mock_vii, mock_pi, mock_bi, mock_oir, mock_prfb, mock_cleanup
+    mock_uiips, mock_capml, mock_srs, mock_vii, mock_pi, mock_bi, mock_oir, mock_prfb, mock_cleanup
 ):
     arches = {'amd64', 's390x'}
     mock_prfb.return_value = {
@@ -581,9 +585,10 @@ def test_handle_rm_request(
     assert mock_bi.call_count == len(arches)
     assert mock_pi.call_count == len(arches)
     mock_vii.assert_called_once()
-    mock_srs.assert_called_once()
+    assert mock_srs.call_count == 2
     mock_capml.assert_called_once()
-    mock_frpb.assert_called_once()
+    mock_uiips.assert_called_once()
+    assert mock_srs.call_args[0][1] == 'complete'
 
 
 @mock.patch('iib.workers.tasks.build._get_resolved_image')


iib/workers/tasks/legacy.py
```diff
@@ -36,8 +36,13 @@ def export_legacy_packages(packages, request_id, rebuilt_index_image, cnr_token,
             _verify_package_info(package_dir, rebuilt_index_image)
             _zip_package(package_dir)
             _push_package_manifest(package_dir, cnr_token, organization)
-
-    set_request_state(request_id, 'in_progress', 'Back ported packages successfully pushed to OMPS')
+    payload = {
+        'index_image': rebuilt_index_image,
+        'packages': packages,
+        'state': 'completed',
```
mprahl, line 42: This is an invalid state. It should be `complete` and not `completed`. The valid states are defined in `iib.web.models::RequestStateMapping`.

iib/workers/tasks/legacy.py
```diff
@@ -36,8 +36,13 @@ def export_legacy_packages(packages, request_id, rebuilt_index_image, cnr_token,
             _verify_package_info(package_dir, rebuilt_index_image)
             _zip_package(package_dir)
             _push_package_manifest(package_dir, cnr_token, organization)
-
-    set_request_state(request_id, 'in_progress', 'Back ported packages successfully pushed to OMPS')
+    payload = {
```
mprahl, line 39: This only gets called if `legacy_support_packages` is `True` in an `add` request and this is never called in an `rm` request, therefore, with your current PR, the request will never get to the `complete` state in many situations.

iib/workers/tasks/legacy.py
chandwanitulsi: Oops, yes I realized that my dev-env didn't test this part to raise errors

iib/workers/tasks/legacy.py
chandwanitulsi: You are right, I missed that. Let me modify my approach and update

iib/workers/tasks/build.py
```diff
@@ -123,7 +123,7 @@ def _create_and_push_manifest_list(request_id, arches):
     return output_pull_spec
 
 
-def _finish_request_post_build(
+def _check_to_modify_index_image(
```
lcarva, line 126: Docstring should also reflect this change, more specifically, it shouldn't say "Finish the request". It should also indicate that it has a return value.

How about calling it something like `update_index_image_in_request` to better reflect its intent?

iib/workers/tasks/build.py
```diff
@@ -167,11 +167,13 @@ def _finish_request_post_build(
     payload = {
         'arches': list(arches),
         'index_image': index_image,
-        'state': 'complete',
-        'state_reason': 'The request completed successfully',
+        'state': 'in_progress',
+        'state_reason': 'Checking and modifying index image',
```
lcarva, line 171: We should use the past tense here, e.g. "Updated the index image"

iib/workers/tasks/build.py
```diff
@@ -584,16 +586,14 @@ def handle_add_request(
 
     set_request_state(request_id, 'in_progress', 'Creating the manifest list')
     output_pull_spec = _create_and_push_manifest_list(request_id, arches)
-
+    modified_from_index = _check_to_modify_index_image(
```
lcarva, line 589: I think `from_index` terminology shouldn't be used here. This is the pull spec of the `index_image` built by IIB. Consider `modified_index_image`.

iib/workers/tasks/build.py
chandwanitulsi: Yup, let me add those in the next commit. TBH I wasn't happy with '_check_to_modify_index_image' either

iib/workers/tasks/build.py
```diff
@@ -156,6 +156,7 @@ def _finish_request_post_build(
         index image.
     :param str overwrite_from_index_token: the token used for overwriting the input
         ``from_index`` image.
+    :rtype: str
```
yashvardhannanavati, line 159: Could you please add `:return: index_image` in the doc string too?

iib/workers/tasks/build.py
```diff
@@ -133,7 +133,7 @@ def _create_and_push_manifest_list(request_id, arches):
     return output_pull_spec
 
 
-def _finish_request_post_build(
+def update_index_image_in_request(
```
yashvardhannanavati, line 136: How about just `update_index_image_pull_spec`? Thoughts?

iib/workers/tasks/legacy.py
```diff
@@ -36,8 +36,13 @@ def export_legacy_packages(packages, request_id, rebuilt_index_image, cnr_token,
...
+        'index_image': rebuilt_index_image,
```
mprahl, line 40: `update_index_image_in_request` already sets this so this can be removed.

iib/workers/tasks/legacy.py
```diff
@@ -36,8 +36,13 @@ def export_legacy_packages(packages, request_id, rebuilt_index_image, cnr_token,
...
+        'packages': packages,
```
mprahl, line 41: What is this for?

chandwanitulsi: Updated in latest commit

iib/workers/tasks/build.py
```diff
@@ -180,10 +182,12 @@ def _finish_request_post_build(
     payload = {
         'arches': list(arches),
         'index_image': index_image,
-        'state': 'complete',
-        'state_reason': 'The request completed successfully',
+        'state': 'in_progress',
+        'state_reason': 'Updated the index image',
```
yashvardhannanavati, line 186: Can we please make this `state_reason` a bit more verbose? How about creating a variable called `state_reason` in the beginning and updating it depending upon which branch gets executed?

@mprahl Thoughts?

iib/workers/tasks/build.py
```diff
@@ -620,20 +624,19 @@ def handle_add_request(
...
         arches,
         from_index,
         overwrite_from_index,
         overwrite_from_index_token,
     )
+    if legacy_support_packages:
+        export_legacy_packages(
+            legacy_support_packages, request_id, modified_index_image, cnr_token, organization
+        )
+    set_request_state(request_id, 'complete', 'Operator bundle added to index image')
```
yashvardhannanavati, line 639: How about adding `The request has been completed successfully.` to the state_reason here.

iib/workers/tasks/build.py
```diff
@@ -681,14 +684,15 @@ def handle_rm_request(
...
-    _finish_request_post_build(
+    update_index_image_pull_spec(
         output_pull_spec,
         request_id,
         arches,
         from_index,
         overwrite_from_index,
         overwrite_from_index_token,
     )
+    set_request_state(request_id, 'complete', 'Operator bundle removed from index image')
```
yashvardhannanavati, line 695: Same as the above comment.

iib/workers/tasks/build.py
yashvardhannanavati: There can be multiple bundles here. Could we please change `bundle` to `bundles(s)`

iib/workers/tasks/build.py
```diff
@@ -133,7 +133,7 @@ def _create_and_push_manifest_list(request_id, arches):
     return output_pull_spec
 
 
-def _finish_request_post_build(
+def update_index_image_pull_spec(
```
mprahl, line 136: Could you please keep this as a private function?

iib/workers/tasks/build.py
mprahl: I like that idea.

chandwanitulsi: @yashvardhannanavati Could you please elaborate on what you mean by "which branch gets executed"?

iib/workers/tasks/build.py
chandwanitulsi: I agree about the bundle(s), I changed the completion message assuming that stating what the request did finally might be more meaningful then the generic completion message. However, I might be wrong about it and I wouldn't mind using the above-mentioned message.

iib/workers/tasks/build.py
chandwanitulsi: oh yes..

iib/workers/tasks/build.py
yashvardhannanavati: yup, you're right. Adding more specific info is more helpful. I meant appending the string `The request has been completed successfully.` to the existing message that you have.

iib/workers/tasks/build.py
yashvardhannanavati: So if the first `if` branch gets executed, maybe you can add the message in `log.info` to the state_reason. Similar if the `elif` block gets executed. Does that make sense?

chandwanitulsi: Got that, earlier I thought you meant changing the state_reason based on what part of add request was executing and I got confused by that. Thanks for clarifying

chandwanitulsi: Got that, earlier I thought you meant changing the state_reason based on what part of add request was executing and I got confused by that. Thanks for clarifying

iib/workers/tasks/build.py
```diff
@@ -156,34 +156,40 @@ def _finish_request_post_build(
...
     """
     conf = get_worker_config()
+    state_reason = 'Updated the index image'
     if from_index and overwrite_from_index:
         log.info(f'Ovewriting the index image {from_index} with {output_pull_spec}')
         index_image = from_index
         exc_msg = f'Failed to overwrite the input from_index container image of {index_image}'
         args = [f'docker://{output_pull_spec}', f'docker://{index_image}']
         _skopeo_copy(*args, copy_all=True, dest_token=overwrite_from_index_token, exc_msg=exc_msg)
+        state_reason = "Overwrote the index image {from_index} with {output_pull_spec}"
```
yashvardhannanavati, line 171: Before Matt catches this, could you please be consistent and use single quotes throughout. :)

chandwanitulsi: New patch already coming :p

yashvardhannanavati: And there should be an `f` before the string for this to work `f'Overwrote the index image {from_index} with {output_pull_spec}'`

iib/workers/tasks/build.py
```diff
@@ -620,20 +626,24 @@ def handle_add_request(
...
     )
+    if legacy_support_packages:
+        export_legacy_packages(
+            legacy_support_packages, request_id, modified_index_image, cnr_token, organization
+        )
+    set_request_state(
+        request_id,
+        'complete',
+        'Operator bundle(s) added to index image, \
+                                               completing the request successfully',
```
yashvardhannanavati, line 645: Wait! Python black isn't complaining to this formatting? That's strange. Does `tox` work successfully?

chandwanitulsi: It was a suggestion from black, unexpected for me too

iib/workers/tasks/build.py
chandwanitulsi: missing in syntax in figuring out the right words already

iib/workers/tasks/build.py
```diff
@@ -156,34 +156,40 @@ def _finish_request_post_build(
         index image.
     :param str overwrite_from_index_token: the token used for overwriting the input
         ``from_index`` image.
+    :return: pull spec of the index_image
+    :rtype: str
     :raises IIBError: if the manifest list couldn't be created and pushed
     """
     conf = get_worker_config()
+    state_reason = 'Updated the index image'
```
chandwanitulsi, line 164: I would appreciate any better suggestions here, is that state_reason okay for the else branch?

iib/workers/tasks/build.py
chandwanitulsi: I guess I mixed up 2 suggestions

iib/workers/tasks/build.py
chandwanitulsi: done

iib/workers/tasks/build.py
chandwanitulsi: updated

iib/workers/tasks/build.py
yashvardhannanavati: It would be nice to add the `output_pull_spec` to the state_reason for the `else` clause.

iib/workers/tasks/build.py
```diff
@@ -158,34 +158,41 @@ def _finish_request_post_build(
...
     payload = {
         'arches': list(arches),
         'index_image': index_image,
-        'state': 'complete',
-        'state_reason': 'The request completed successfully',
+        'state': 'in_progress',
+        'state_reason': state_reason,
     }
-    update_request(request_id, payload, exc_msg='Failed setting the index image on the request')
+    update_request(request_id, payload, exc_msg='Failed updating the index image on the request')
```
mprahl, line 193: Why was this state reason changed?

mprahl, line 193: Why was the exception message changed?

iib/workers/tasks/build.py
```diff
@@ -158,34 +158,41 @@ def _finish_request_post_build(
         index image.
     :param str overwrite_from_index_token: the token used for overwriting the input
         ``from_index`` image.
+    :return: pull spec of the index_image
+    :rtype: str
     :raises IIBError: if the manifest list couldn't be created and pushed
     """
     conf = get_worker_config()
+    state_reason = ''
```
mprahl, line 166: I'd prefer if this was removed since I'd rather get an exception in the tests if a code path does not define `state_reason`.

iib/workers/tasks/build.py
```diff
@@ -158,34 +158,41 @@ def _finish_request_post_build(
...
     if from_index and overwrite_from_index:
         log.info(f'Ovewriting the index image {from_index} with {output_pull_spec}')
         index_image = from_index
         exc_msg = f'Failed to overwrite the input from_index container image of {index_image}'
         args = [f'docker://{output_pull_spec}', f'docker://{index_image}']
         _skopeo_copy(*args, copy_all=True, dest_token=overwrite_from_index_token, exc_msg=exc_msg)
+        state_reason = f'Overwrote the index image {from_index} with {output_pull_spec}'
```
mprahl, line 173: Could you reuse this as the log message like you did in the `elif` branch?

iib/workers/tasks/build.py
```diff
@@ -670,20 +677,23 @@ def handle_add_request(
...
         output_pull_spec,
         request_id,
         arches,
         from_index,
         overwrite_from_index,
         overwrite_from_index_token,
     )
+    if legacy_support_packages:
+        export_legacy_packages(
+            legacy_support_packages, request_id, modified_index_image, cnr_token, organization
```
mprahl, line 690: I'd rather continue using `output_pull_spec` in this method since it will be using `quay.io` directly rather than the registry proxy, which is unreliable. The registry proxy is only provided as the `index_image` value in the REST API for convenience. IIB should use quay.io directly when possible.

iib/workers/tasks/legacy.py
```diff
@@ -36,8 +36,11 @@ def export_legacy_packages(packages, request_id, rebuilt_index_image, cnr_token,
             _verify_package_info(package_dir, rebuilt_index_image)
             _zip_package(package_dir)
             _push_package_manifest(package_dir, cnr_token, organization)
-
-    set_request_state(request_id, 'in_progress', 'Back ported packages successfully pushed to OMPS')
+    payload = {
+        'state': 'in_progress',
+        'state_reason': 'Back ported packages successfully pushed to OMPS',
+    }
+    update_request(request_id, payload, exc_msg='Failed to push package to OMPS')
```
mprahl, line 43: `exc_msg` is incorrect in this case since if an exception is raised from `update_request`, then that means updating the request using the IIB PATCH API endpoint failed, not the interaction with OMPS.

iib/workers/tasks/legacy.py
```diff
@@ -36,8 +36,11 @@ def export_legacy_packages(packages, request_id, rebuilt_index_image, cnr_token,
             _verify_package_info(package_dir, rebuilt_index_image)
             _zip_package(package_dir)
             _push_package_manifest(package_dir, cnr_token, organization)
-
-    set_request_state(request_id, 'in_progress', 'Back ported packages successfully pushed to OMPS')
```
mprahl, line 40: Please revert back your change in this method so that the method continues using `set_request_state`. It's a wrapper around `update_request` so that you don't have to specify the payload like you did below.

iib/workers/tasks/build.py
```diff
@@ -731,14 +741,19 @@ def handle_rm_request(
...
         request_id,
         arches,
         from_index,
         overwrite_from_index,
         overwrite_from_index_token,
     )
+    set_request_state(
+        request_id,
+        'complete',
+        'Operator bundle(s) removed from index image, completing the request successfully',
```
mprahl, line 755: In this case, it's not individual bundles that are removed, it's all bundles associated with an operator. Also, I don't think there is any value in adding `, completing the request successfully`. The word `completing` insinuates there is more that will occur but actually, the request is complete when the state is set to `complete`.
```suggestion
        'The operators were successfully removed from index image',
```

iib/workers/tasks/build.py
```diff
@@ -670,20 +677,23 @@ def handle_add_request(
...
         overwrite_from_index_token,
     )
+    if legacy_support_packages:
+        export_legacy_packages(
+            legacy_support_packages, request_id, modified_index_image, cnr_token, organization
+        )
+    set_request_state(
+        request_id,
+        'complete',
+        'Operator bundle(s) added to index image, completing the request successfully',
```
mprahl, line 695: Same as my comment in the rm endpoint.
```suggestion
        'The operator bundle(s) were successfully added to the index image',
```

iib/workers/tasks/build.py
chandwanitulsi: yup, makes sense

iib/workers/tasks/build.py
chandwanitulsi: yeah, the reason I added again here was that I was trying to keep consistencies with the past tenses for state_reason, used in this function. Let me change it to use log message

iib/workers/tasks/build.py
chandwanitulsi: Aah, this was from my initial commit, changing it back

iib/workers/tasks/build.py
chandwanitulsi: so that means if the index image is overridden by using overwrite_from_index, the push to OMPS should not consider it?

iib/workers/tasks/build.py
chandwanitulsi: I added it to consider another feedback assuming "the request completed successfully" was being more meaningful

iib/workers/tasks/build.py
chandwanitulsi: same as earlier, I actually liked this way of message

iib/workers/tasks/legacy.py
chandwanitulsi: yup, let me change in next patch

iib/workers/tasks/legacy.py
chandwanitulsi: yeah, I used update_request since earlier I added something in payload, but with feedback I realized that wasn't needed, now I can use the set_request_state() directly, changing in next patch

iib/workers/tasks/build.py
mprahl: No, I'm just saying that the push to OMPS should use the unaltered pull specification of the index image.

iib/workers/tasks/build.py
mprahl: I'm sorry, but I don't understand what you mean. Could you please elaborate or rephrase your comment?

chandwanitulsi: In one of my earlier patches, the message was 'Operator bundle(s) added to index image', but later to consider the suggested feedback that 'The request has been completed successfully' should be added,  I changed it 'Operator bundle(s) added to index image, completing the request successfully' and looks like adding it this way isn't any better
 I like this suggestion made by you and I think it's better than both the earlier messages I used, so I will add 'The operator bundle(s) were successfully added to the index image' in next patch.

yashvardhannanavati: This is the feedback Tulsi is talking about. https://github.com/release-engineering/iib/pull/76#discussion_r425423401

iib/workers/tasks/build.py
```diff
@@ -162,28 +162,31 @@ def _finish_request_post_build(
...
-            output_pull_spec,
-            index_image,
+        output_message = (
+            f'Changed the index_image pull specification from {output_pull_spec} to {index_image}'
         )
+        log.info(output_message)
+        state_reason = output_message
     else:
         index_image = output_pull_spec
+        state_reason = f'Updated the index image with {output_pull_spec}'
```
mprahl, line 183: This state reason is too verbose. I think it'd be nice to omit this from being a state reason update.

iib/workers/tasks/build.py
```diff
@@ -162,28 +162,31 @@ def _finish_request_post_build(
...
     elif conf['iib_index_image_output_registry']:
         index_image = output_pull_spec.replace(
             conf['iib_registry'], conf['iib_index_image_output_registry'], 1
         )
-        log.info(
-            'Changed the index_image pull specification from %s to %s',
-            output_pull_spec,
-            index_image,
+        output_message = (
+            f'Changed the index_image pull specification from {output_pull_spec} to {index_image}'
```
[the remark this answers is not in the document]
chandwanitulsi: I am not sure if I get it right, Did you mean I should change the state_reason for this branch (not keep it consistent with the log message) or not a have a different state_reason for each if-else branch?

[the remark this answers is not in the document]
mprahl: Sorry about the confusion, I just mean that the branch under `if from_index and overwrite_from_index:` should update the state reason but the other two branches shouldn't update the state reason.

iib/workers/tasks/build.py
```diff
@@ -162,28 +162,31 @@ def _finish_request_post_build(
     """
     conf = get_worker_config()
     if from_index and overwrite_from_index:
-        log.info(f'Ovewriting the index image {from_index} with {output_pull_spec}')
+        output_message = f'Ovewriting the index image {from_index} with {output_pull_spec}'
```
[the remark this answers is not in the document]
chandwanitulsi: How about just moving the log after _skopeo_copy. It will add past tense which will be applicable to both state_reason and logging?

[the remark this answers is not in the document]
mprahl: I like the log message before it since it provides context for why the `_skopeo_copy` occurred and this is helpful if it fails.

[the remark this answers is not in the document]
mprahl: My preference would be to actually update the state reason with this in the present tense before `_skopeo_copy`. Then in the other branches in the if block would would not update the state reason, due to my comment in https://github.com/release-engineering/iib/pull/76#discussion_r433456758.

iib/workers/tasks/build.py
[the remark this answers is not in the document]
chandwanitulsi: Got it, thanks for clarifying, we would need some default state_reason to pass it to the payload though, isn't it?
How about using a default state_reason to "Updated the index image with {output_pull_spec}" and just updating it for the if from_index and overwrite_from_index: branch, Would that be too verbose as well?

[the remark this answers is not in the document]
mprahl: You can just remove the `state` and `state_reason` keys from `payload` and before `_skopeo_copy`, you can call `set_request_state` separately to update the state then.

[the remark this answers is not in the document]
chandwanitulsi: okay, thanks, updating in next patch

iib/workers/tasks/build.py
```diff
@@ -731,14 +735,17 @@ def handle_rm_request(
...
+    _update_index_image_pull_spec(
         output_pull_spec,
         request_id,
         arches,
         from_index,
         overwrite_from_index,
         overwrite_from_index_token,
     )
+    set_request_state(
+        request_id, 'complete', 'The operator(s) were successfully removed from index image',
```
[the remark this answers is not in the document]
chandwanitulsi: updated in next patch

iib/workers/tasks/build.py
```diff
@@ -670,20 +673,21 @@ def handle_add_request(
...
         from_index,
         overwrite_from_index,
         overwrite_from_index_token,
     )
+    if legacy_support_packages:
+        export_legacy_packages(
+            legacy_support_packages, request_id, output_pull_spec, cnr_token, organization
+        )
+    set_request_state(
+        request_id, 'complete', 'The operator bundle(s) were successfully added to index image',
```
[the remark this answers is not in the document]
chandwanitulsi: updated in next patch

iib/workers/tasks/build.py
```diff
@@ -162,10 +162,13 @@ def _finish_request_post_build(
     """
     conf = get_worker_config()
     if from_index and overwrite_from_index:
-        log.info(f'Ovewriting the index image {from_index} with {output_pull_spec}')
+        output_message = f'Ovewriting the index image {from_index} with {output_pull_spec}'
```
mprahl, line 165:
```suggestion
        output_message = f'Overwriting the index image {from_index} with {output_pull_spec}'
```

chandwanitulsi: Oops..

mprahl:
@chandwanitulsi could you please change the commit message on the second commit to `Fix a typo in build.py`?

chandwanitulsi:
Commit message updated

PR #77, opened 2020-05-05
Add the regenerate-bundle-batch API endpoint

mprahl:
This allows a user to submit multiple regenerate-bundle requests as part of a single batch.

Additionally, this fixes the OpenAPI spec for regenerate-bundle and changes messaging.py to use the correct logger.

### DIFF
--- a/iib/web/api_v1.py
+++ b/iib/web/api_v1.py
@@ -338,3 +338,60 @@ def regenerate_bundle():
     messaging.send_message_for_state_change(request, new_batch_msg=True)
     flask.current_app.logger.debug('Successfully scheduled request %d', request.id)
     return flask.jsonify(request.to_json()), 201
+
+
+@api_v1.route('/builds/regenerate-bundle-batch', methods=['POST'])
+@login_required
+def regenerate_bundle_batch():
+    """
+    Submit a batch of requests to regenerate operator bundle images.
+
+    :rtype: flask.Response
+    :raise ValidationError: if required parameters are not supplied
+    """
+    payloads = flask.request.get_json()
+    if not isinstance(payloads, list) or not payloads:
+        raise ValidationError('The input data must be a non-empty JSON array')
+
+    batch = Batch()
+    db.session.add(batch)
+    requests = []
+    # Iterate through all the payloads and verify that the requests are valid before committing them
+    # and scheduling the tasks
+    for payload in payloads:
+        try:
+            request = RequestRegenerateBundle.from_json(payload, batch)
+        except ValidationError as e:
+            # Rollback the transaction if any of the payloads are invalid
+            db.session.rollback()
+            raise ValidationError(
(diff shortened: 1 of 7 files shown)


iib/web/messaging.py
```diff
@@ -164,53 +238,46 @@ def send_message_for_state_change(request, new_batch_msg=False):
...
+    envelopes = []
+
+    for request in requests:
+        request_envelope = _get_request_state_change_envelope(request)
+        if request_envelope:
+            envelopes.append(request_envelope)
+        else:
+            # If no request envelope was generated, then that means IIB is not configured to
+            # send request state change messages and the rest of the loop can be skipped
+            break
```
lcarva, line 274: This looks like a minor optimization that may cause some unexpected errors in the future due to unrelated code changes.

iib/web/messaging.py
```diff
@@ -164,53 +238,46 @@ def send_message_for_state_change(request, new_batch_msg=False):
...
+
+    # Just use the last request's batch since the batch is the same for all of them
+    batch = request.batch
```
lcarva, line 277: I find that using a variable, that has been defined in a for-loop statement, outside the for-loop to increase the difficulty in understanding and debugging the code. Consider taking the batch ID as a function parameter. You could even consider raising an error if any of the requests have a different batch ID.

iib/web/messaging.py
```diff
@@ -164,53 +238,46 @@ def send_message_for_state_change(request, new_batch_msg=False):
...
+
+def send_messages_for_new_batch_of_requests(requests):
+    """
+    Send the appropriate message(s) based on a new batch of build requests.
+
+    If IIB is not configured to send messages, this function will do nothing.
+
+    :param list requests: the requests that were created as part of the batch request
+    """
+    if not requests:
```
lcarva, line 262: Missing unit test coverage here.

iib/web/messaging.py
mprahl: I'll just use `batch = requests[0].batch` instead.

iib/web/api_v1.py
```diff
@@ -338,3 +338,60 @@ def regenerate_bundle():
...
+    :raise ValidationError: if required parameters are not supplied
+    """
+    payloads = flask.request.get_json()
+    if not isinstance(payloads, list) or not payloads:
+        raise ValidationError('The input data must be a non-empty JSON array')
+
+    batch = Batch()
+    db.session.add(batch)
+    requests = []
+    # Iterate through all the requests and verify that they are valid before committing them and
```
yashvardhannanavati, line 359: `# Iterate through all the requests` => `# Iterate through all the payloads`

mprahl:
@yashvardhannanavati and @lcarva it occurred to me that I should send the state change messages before the tasks are scheduled to avoid a race condition where the workers could have started on the tasks and updated the state before all the state change messages are sent out.

I just moved it up in the last push.

PR #80, opened 2020-05-07
Support adding annotations to a batch when using the regenerate-bundle-batch API endpoint

mprahl:
The first commit changes the regenerate-bundle-batch API endpoint to accept a JSON object instead of an array.

The second commit adds support for setting annotations on a batch when using the regenerate-bundle-batch API endpoint.

Resolves CLOUDBLD-924

### DIFF
--- a/iib/web/models.py
+++ b/iib/web/models.py
@@ -1,6 +1,7 @@
 # SPDX-License-Identifier: GPL-3.0-or-later
 from copy import deepcopy
 from enum import Enum
+import json
 
 from flask import current_app
 from flask_login import UserMixin, current_user
@@ -366,6 +367,7 @@ class Request(db.Model):
 
         latest_state = None
         if verbose:
+            rv['batch_annotations'] = self.batch.annotations
             states = [_state_to_json(state) for state in self.states]
             # Reverse the list since the latest states should be first
             states = list(reversed(states))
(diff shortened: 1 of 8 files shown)


iib/web/models.py
```diff
@@ -389,9 +392,49 @@ class Batch(db.Model):
     """A batch associated with one or more requests."""
 
     id = db.Column(db.Integer, primary_key=True)
+    _annotations = db.Column('annotations', db.Text, nullable=True)
 
     requests = db.relationship('Request', foreign_keys=[Request.batch_id], back_populates='batch')
 
+    @hybrid_property
```
lcarva, line 399: This is new to me. From what I can read in the [docs](https://docs.sqlalchemy.org/en/13/orm/extensions/hybrid.html) this is meant to translate the function body into SQL statements. Why is that needed here instead of a plain `@property`?

iib/web/models.py
```diff
@@ -389,9 +392,49 @@ class Batch(db.Model):
...
+            or not isinstance(payload.get('build_requests'), list)
+            or not payload['build_requests']
+        ):
+            raise ValidationError(
+                'The input data must be a JSON object with the "build_requests" key\'s value as a '
+                'non-empty array'
+            )
+
+        if not isinstance(payload.get('annotations', {}), dict):
+            raise ValidationError('The "annotations" key\'s value must be a JSON object')
```
yashvardhannanavati, line 436: How about changing this to `Value of "annotations" must be a JSON object`?

iib/web/models.py
```diff
@@ -389,9 +392,49 @@ class Batch(db.Model):
...
+
+        :raises ValidationError: if the payload is invalid
+        """
+        if (
+            not isinstance(payload, dict)
+            or not isinstance(payload.get('build_requests'), list)
+            or not payload['build_requests']
+        ):
+            raise ValidationError(
+                'The input data must be a JSON object with the "build_requests" key\'s value as a '
```
yashvardhannanavati, line 431: How about `...and the value of "builds_requests" must be a non-empty array`?

mprahl:
@yashvardhannanavati just a gentle nudge to remind you to review this PR.  :smile:

PR #81, opened 2020-05-11
Resolve manifest list for bundle image references to arch specific digest

yashvardhannanavati:
Refers to CLOUDDST-585

### DIFF
--- a/iib/workers/tasks/build.py
+++ b/iib/workers/tasks/build.py
@@ -255,6 +255,50 @@ def get_rebuilt_image_pull_spec(request_id):
     )
 
 
+def _get_resolved_bundles(bundles):
+    """
+    Get the pull specification of the bundle images using their digests.
+
+    Determine if the pull spec refers to a manifest list.
+    If so, simply use the digest of the first item in the manifest list.
+    If not a manifest list, it must be a v2s2 image manifest and should be used as it is.
+
+    :param list bundles: the list of bundle images to be resolved.
+    :return: the list of bundle images resolved to their digests.
+    :rtype: list
+    :raises IIBError: if unable to resolve a bundle image.
+    """
+    log.info('Resolving bundles %s', ', '.join(bundles))
+    resolved_bundles = set()
+    for bundle_pull_spec in bundles:
+        skopeo_raw = skopeo_inspect(f'docker://{bundle_pull_spec}', '--raw')
+        if (
+            skopeo_raw.get('mediaType')
+            == 'application/vnd.docker.distribution.manifest.list.v2+json'
+        ):
+            # Get the digest of the first item in the manifest list
+            digest = skopeo_raw['manifests'][0]['digest']
+            if '@' in bundle_pull_spec:
+                repo = bundle_pull_spec.split('@', 1)[0]
+            else:
+                repo = bundle_pull_spec.rsplit(':', 1)[0]
+            resolved_bundles.add(f'{repo}@{digest}')
+        elif (
+            skopeo_raw.get('mediaType') == 'application/vnd.docker.distribution.manifest.v2+json'
+            and skopeo_raw.get('schemaVersion') == 2
+        ):
+            resolved_bundles.add(_get_resolved_image(bundle_pull_spec))
+        else:
+            error_msg = (
+                f'The pull specification of {bundle_pull_spec} is neither '
+                f'a v2 manifest list nor a v2s2 manifest. Type {skopeo_raw.get("mediaType")}'
+                f' and schema version {skopeo_raw.get("schemaVersion")} is not supported by IIB.'
+            )
+            raise IIBError(error_msg)
+
+    return list(resolved_bundles)
+
+
 def _get_resolved_image(pull_spec):
     """
     Get the pull specification of the container image using its digest.
@@ -373,7 +417,7 @@ def _prepare_request_for_build(
         currently built for; if ``from_index`` is ``None``, then this is used as the list of arches
         to build the index image for
     :param list bundles: the list of bundles to create the bundle mapping on the request
-    :return: a dictionary with the keys: arches, binary_image_resolved, and fom_index_resolved.
+    :return: a dictionary with the keys: arches, binary_image_resolved, and from_index_resolved.
     :raises IIBError: if the container image resolution fails or the architectures couldn't be
         detected.
     """
@@ -587,16 +631,21 @@ def handle_add_request(
     :raises IIBError: if the index image build fails or legacy support is required and one of
         ``cnr_token`` or ``organization`` is not specified.
     """
-    _verify_labels(bundles)
+    # Resolve bundles to their digests
+    resolved_bundles = _get_resolved_bundles(bundles)
+
+    _verify_labels(resolved_bundles)
 
     # Check if Gating passes for all the bundles
     if greenwave_config:
-        gate_bundles(bundles, greenwave_config)
+        gate_bundles(resolved_bundles, greenwave_config)
 
     log.info('Checking if interacting with the legacy app registry is required')
-    legacy_support_packages = get_legacy_support_packages(bundles)
+    legacy_support_packages = get_legacy_support_packages(resolved_bundles)
     if legacy_support_packages:
-        validate_legacy_params_and_config(legacy_support_packages, bundles, cnr_token, organization)
+        validate_legacy_params_and_config(
+            legacy_support_packages, resolved_bundles, cnr_token, organization
+        )
 
     _cleanup()
     prebuild_info = _prepare_request_for_build(
@@ -604,7 +653,9 @@ def handle_add_request(
     )
 
     with tempfile.TemporaryDirectory(prefix='iib-') as temp_dir:
-        _opm_index_add(temp_dir, bundles, prebuild_info['binary_image_resolved'], from_index)
+        _opm_index_add(
+            temp_dir, resolved_bundles, prebuild_info['binary_image_resolved'], from_index,
+        )
 
         arches = prebuild_info['arches']
         for arch in sorted(arches):
--- a/tests/test_workers/test_tasks/test_build.py
+++ b/tests/test_workers/test_tasks/test_build.py
@@ -175,6 +175,54 @@ def test_get_resolved_image(mock_si):
     assert rv == 'some-image@sha256:abcdefg'
 
 
+@pytest.mark.parametrize(
+    'skopeo_inspect_rv, expected_response',
+    (
+        (
+            {
+                'mediaType': 'application/vnd.docker.distribution.manifest.list.v2+json',
+                'manifests': [
+                    {'platform': {'architecture': 'amd64'}, 'digest': 'arch_digest'},
+                    {'platform': {'architecture': 's390x'}, 'digest': 'different_arch_digest'},
+                ],
+            },
+            ['some_bundle@arch_digest'],
+        ),
+        (
+            {
+                'mediaType': 'application/vnd.docker.distribution.manifest.v2+json',
+                'schemaVersion': 2,
+            },
+            ['some_bundle@manifest_digest'],
+        ),
+    ),
+)
+@mock.patch('iib.workers.tasks.build._get_resolved_image')
+@mock.patch('iib.workers.tasks.build.skopeo_inspect')
+def test_get_resolved_bundles_success(mock_si, mock_gri, skopeo_inspect_rv, expected_response):
+    mock_si.return_value = skopeo_inspect_rv
+    mock_gri.return_value = 'some_bundle@manifest_digest'
+    response = build._get_resolved_bundles(['some_bundle:1.2'])
+    if skopeo_inspect_rv['mediaType'] == 'application/vnd.docker.distribution.manifest.v2+json':
+        mock_gri.assert_called_once()
+    else:
+        mock_gri.assert_not_called()
+    assert response == expected_response
+
+
+@mock.patch('iib.workers.tasks.build.skopeo_inspect')
+def test_get_resolved_bundles_failure(mock_si):
+    skopeo_inspect_rv = {
+        'mediaType': 'application/vnd.docker.distribution.notmanifest.v2+json',
+        'schemaVersion': 1,
+    }
+    mock_si.return_value = skopeo_inspect_rv
+    with pytest.raises(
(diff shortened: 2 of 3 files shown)


iib/workers/tasks/build.py
```diff
@@ -260,6 +260,56 @@ def get_rebuilt_image_pull_spec(request_id):
...
+    Get the pull specification of the bundle images using their digests.
+
+    Determine if the pull spec refers to a manifest list.
+    If so, simply use the digest of the amd64 arch in the manifest list.
+    If not a manifest list, it must be a v2s2 image manifest and should be used as it is.
+
+    :param list bundles: the list of bundle images to be resolved.
+    :return: the list of bundle images resolved to their digests.
+    :rtype: list
+    :raises IIBError: if unable to resolve the bundle image.
```
mprahl, line 274:
```suggestion
    :raises IIBError: if unable to resolve a bundle image.
```

iib/workers/tasks/build.py
```diff
@@ -260,6 +260,56 @@ def get_rebuilt_image_pull_spec(request_id):
...
+    """
+    log.debug('Resolving bundles %s', ', '.join(bundles))
```
mprahl, line 276:
```suggestion
    log.info('Resolving bundles %s', ', '.join(bundles))
```

iib/workers/tasks/build.py
```diff
@@ -260,6 +260,56 @@ def get_rebuilt_image_pull_spec(request_id):
...
+                    # to the _get_resolved_image to make sure we have the pull_spec with digest.
+                    resolved_pull_spec = _get_resolved_image(bundle_pull_spec)
+                    resolved_bundles.add(
+                        f'{resolved_pull_spec.rsplit("@", 1)[0]}@{manifest["digest"]}'
+                    )
+                    break
+            if not arch_found:
+                raise IIBError(
+                    f'Cannot resolve bundle for {bundle_pull_spec}. Arch "amd64" is not available.'
+                )
```
mprahl, line 298:
```suggestion
            for manifest in skopeo_raw['manifests']:
                if manifest['platform']['architecture'] == 'amd64':
                    # The user may specify pull_spec with either tag or digest. Hence, make a call
                    # to the _get_resolved_image to make sure we have the pull_spec with digest.
                    resolved_pull_spec = _get_resolved_image(bundle_pull_spec)
                    resolved_bundles.add(
                        f'{resolved_pull_spec.rsplit("@", 1)[0]}@{manifest["digest"]}'
                    )
                    break
            else:
                raise IIBError(
                    f'Cannot resolve bundle for {bundle_pull_spec}. Arch "amd64" is not available.'
                )
```

iib/workers/tasks/build.py
```diff
@@ -260,6 +260,56 @@ def get_rebuilt_image_pull_spec(request_id):
...
+        ):
+            arch_found = False
+            for manifest in skopeo_raw['manifests']:
+                if manifest['platform']['architecture'] == 'amd64':
+                    arch_found = True
+                    # The user may specify pull_spec with either tag or digest. Hence, make a call
+                    # to the _get_resolved_image to make sure we have the pull_spec with digest.
+                    resolved_pull_spec = _get_resolved_image(bundle_pull_spec)
+                    resolved_bundles.add(
+                        f'{resolved_pull_spec.rsplit("@", 1)[0]}@{manifest["digest"]}'
```
mprahl, line 292: What's wrong with using `resolved_pull_spec` directly?

iib/workers/tasks/build.py
```diff
@@ -260,6 +260,56 @@ def get_rebuilt_image_pull_spec(request_id):
...
+                    )
+                    break
+            if not arch_found:
+                raise IIBError(
+                    f'Cannot resolve bundle for {bundle_pull_spec}. Arch "amd64" is not available.'
```
mprahl, line 297:
```suggestion
                    f'Cannot resolve the bundle {bundle_pull_spec} since it is not available on arch "amd64"'
```

iib/workers/tasks/build.py
```diff
@@ -378,12 +428,16 @@ def _prepare_request_for_build(
...
         to build the index image for
     :param list bundles: the list of bundles to create the bundle mapping on the request
-    :return: a dictionary with the keys: arches, binary_image_resolved, and fom_index_resolved.
+    :return: a dictionary with the keys: arches, binary_image_resolved, bundles_resolved,
+        and from_index_resolved.
     :raises IIBError: if the container image resolution fails or the architectures couldn't be
         detected.
     """
     if bundles is None:
         bundles = []
```
mprahl, line 437: You should use `resolved_bundles` instead of `bundles` when getting the bundle mapping below.

mprahl:
Also, `get_legacy_support_packages` should probably use the resolved bundles as well.

iib/workers/tasks/build.py
yashvardhannanavati: If it is a manifest list, and we use that digest (which resolved_pull_spec uses), opm is unable to resolve those to arch specific digest. Hence, we are providing it with the things it needs.

yashvardhannanavati:
> Also, `get_legacy_support_packages` should probably use the resolved bundles as well.

I don't think it makes a difference for `get_legacy_support_packages` since all it does is get labels for the bundle. I think we can keep it as it is. Thoughts?

iib/workers/tasks/build.py
mprahl: Right. I misread the code.

mprahl:
> > Also, `get_legacy_support_packages` should probably use the resolved bundles as well.
> 
> I don't think it makes a difference for `get_legacy_support_packages` since all it does is get package name from the labels of the bundle. I think we can keep it as it is. Thoughts?

I'm not entirely sure how `skopeo inspect` works under the hood, but since the labels are on the manifests and not manifest lists, I think `skopeo inspect` picks one of the manifests and gets the labels for it. It'd be nice to use the same manifest throughout so that it is consistent.

Here is how I think skopeo does it:
```bash
TOKEN=$(curl -s "https://auth.docker.io/token?service=registry.docker.io&scope=repository:library/fedora:pull" | jq -r .token)
AMD64_DIGEST=$(curl -s -H "Authorization: Bearer $TOKEN" -H "Accept: application/vnd.docker.distribution.manifest.list.v2+json" https://index.docker.io/v2/library/fedora/manifests/31 | jq -r .manifests[0].digest)
CONFIG_DIGEST=$(curl -s -H "Authorization: Bearer $TOKEN" https://index.docker.io/v2/library/fedora/manifests/$AMD64_DIGEST | jq -r .config.digest)
curl -L -s -H "Authorization: Bearer $TOKEN" -H "Accept: application/vnd.docker.distribution.manifest.list.v2+json" https://index.docker.io/v2/library/fedora/blobs/$CONFIG_DIGEST | jq .config.Labels
```

The output is the following:
```
{
  "maintainer": "Clement Verna <cverna@fedoraproject.org>"
}
```

iib/workers/tasks/build.py
```diff
@@ -260,6 +260,55 @@ def get_rebuilt_image_pull_spec(request_id):
...
+                    'since it is not available on arch "amd64".'
+                )
+        elif (
+            skopeo_raw.get('mediaType') == 'application/vnd.docker.distribution.manifest.v2+json'
+            and skopeo_raw.get('schemaVersion') == 2
+        ):
+            resolved_bundles.add(_get_resolved_image(bundle_pull_spec))
+        else:
+            raise IIBError(
+                f'The pull specification of {bundle_pull_spec} is neither'
```
lcarva, line 305: It might be helpful to include the mediaType here.

iib/workers/tasks/build.py
```diff
@@ -260,6 +260,55 @@ def get_rebuilt_image_pull_spec(request_id):
...
+        skopeo_raw = skopeo_inspect(f'docker://{bundle_pull_spec}', '--raw')
+        if (
+            skopeo_raw.get('mediaType')
+            == 'application/vnd.docker.distribution.manifest.list.v2+json'
+        ):
+            for manifest in skopeo_raw['manifests']:
+                if manifest['platform']['architecture'] == 'amd64':
+                    # The user may specify pull_spec with either tag or digest. Hence, make a call
+                    # to the _get_resolved_image to make sure we have the pull_spec with digest.
+                    resolved_pull_spec = _get_resolved_image(bundle_pull_spec)
```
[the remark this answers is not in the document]
yashvardhannanavati: Sure, I just thought fetching the name using `skopeo inspect` is cleaner. But I guess it's not worth the extra call.

iib/workers/tasks/build.py
```diff
@@ -255,6 +255,50 @@ def get_rebuilt_image_pull_spec(request_id):
...
+    for bundle_pull_spec in bundles:
+        skopeo_raw = skopeo_inspect(f'docker://{bundle_pull_spec}', '--raw')
+        if (
+            skopeo_raw.get('mediaType')
+            == 'application/vnd.docker.distribution.manifest.list.v2+json'
+        ):
+            # Get the digest of the first item in the manifest list
+            digest = skopeo_raw['manifests'][0]['digest']
+            if '@' in bundle_pull_spec:
+                repo = bundle_pull_spec.split('@', 1)[0]
```
midnightercz, line 282: why do you get only first item? If it's expected to get amd64 (but question is still relevant, why only amd64), isn't safer to query via platform.architecture? What if there's going to be aarch64 in the future?

yashvardhannanavati: As Luiz mentioned, it does not matter which arch manifest you get as long is it is not the manifest list digest. Hence, only the first item is taken.

tests/test_workers/test_tasks/test_build.py
```diff
@@ -175,6 +175,52 @@ def test_get_resolved_image(mock_si):
...
 
+@pytest.mark.parametrize(
+    'skopeo_inspect_rv, expected_response',
+    (
+        (
+            {
+                'mediaType': 'application/vnd.docker.distribution.manifest.list.v2+json',
+                'manifests': [
+                    {'platform': {'architecture': 'amd64'}, 'digest': 'arch_digest'},
+                    {'platform': {'architecture': 's390x'}, 'digest': 'arch_digest'},
```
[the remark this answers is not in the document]
yashvardhannanavati: Same as the comment below :)

iib/workers/tasks/build.py
```diff
@@ -589,22 +633,29 @@ def handle_add_request(
     """
     _verify_labels(bundles)
 
+    # Resolve bundles to their digests
+    resolved_bundles = _get_resolved_bundles(bundles)
```
mprahl, line 637: Could you consolidate `_verify_labels` and `_get_resolved_bundles` since you are essentially doing `skopeo inspect` on each bundle two times and that is a very expensive operation?

mprahl: Actually, it looks like `get_legacy_support_packages` calls skopeo inspect on every bundle a third time. Bonus points if you can consolidate that as well.

tests/test_workers/test_tasks/test_build.py
```diff
@@ -175,6 +175,52 @@ def test_get_resolved_image(mock_si):
...
+    ),
+)
+@mock.patch('iib.workers.tasks.build._get_resolved_image')
+@mock.patch('iib.workers.tasks.build.skopeo_inspect')
+def test_get_resolved_bundles_success(mock_si, mock_gri, skopeo_inspect_rv, expected_response):
+    mock_si.return_value = skopeo_inspect_rv
+    mock_gri.return_value = 'some_bundle@manifest_digest'
+    response = build._get_resolved_bundles(['some_bundle:1.2'])
+    if skopeo_inspect_rv['mediaType'] == 'application/vnd.docker.distribution.manifest.v2+json':
+        mock_gri.assert_called_once()
```
mprahl, line 207:
```suggestion
    if skopeo_inspect_rv['mediaType'] == 'application/vnd.docker.distribution.manifest.v2+json':
        mock_gri.assert_called_once()
    else:
        mock_gri.assert_not_called()
```

tests/test_workers/test_tasks/test_build.py
```diff
@@ -445,10 +494,37 @@ def test_handle_add_request(
...
+            None,
+            greenwave_config,
+        )
+    mock_vl.assert_called_once()
+    mock_gb.assert_called_once_with(['some-bundle@sha'], greenwave_config)
 
+
+@mock.patch('iib.workers.tasks.build._verify_labels')
+@mock.patch('iib.workers.tasks.build._get_resolved_bundles')
+def test_handle_add_request_bundle_resolution_failure(mock_grb, mock_vl):
```
mprahl, line 525: I don't see any value in this test. I think it could be removed but I'd like opinions from others review this PR.

iib/workers/tasks/build.py
yashvardhannanavati: I thought `_verify_labels` does a non-raw `skopeo inspect` while `_get_resolved_bundles` does a raw `skopeo inspect`. Hence, I did not think of combining the two. 

It does make sense to combine `_verify_labels` and `get_legacy_support_packages` though. But that is outside of the scope of this PR.

Please let me know if I am missing something?

tests/test_workers/test_tasks/test_build.py
yashvardhannanavati: I think there's no harm in being extra cautious but if everyone feels it can be removed, I will be happy to remove it. :)

iib/workers/tasks/build.py
mprahl: Ah you're right. Can you resolve the bundles before verifying the labels? That way the labels are verified on the same manifests that were selected.

iib/workers/tasks/build.py
lcarva: Yeah. The OLM team has explicitly requested this behavior. Operator bundle images are just data containers and are built for a particular arch only out of necessity.

tests/test_workers/test_tasks/test_build.py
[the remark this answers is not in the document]
lcarva: Actually, if we're claiming that IIB always picks the first manifest in the manifest list, we need the digest to be different in the test case in order to verify it's doing that.

mprahl:
As mentioned in chat, we will wait to merge this until a new version of opm is released and opm should be updated in the container image in this PR.

PR #84, opened 2020-05-11
Modify the regenerated bundle based on the organization

mprahl:
* Add a suffix to the operator name based on an organization configuration
* Replace the registry on pull specs in a bundle based on an organization configuration
* Annotate ClusterServiceVersion objects based on an organization configuration

Resolves CLOUDBLD-525


iib/workers/tasks/build.py
```diff
@@ -833,20 +923,94 @@ def _adjust_operator_manifests(manifests_path):
...
+            replacement_needed = True
+
+        if registry_replacements.get(resolved_image.registry):
+            replacement_needed = True
+            log.debug(
+                'Replacing the registry on %s with %s',
+                pullspec,
+                registry_replacements[resolved_image.registry],
+            )
+            resolved_image.registry = registry_replacements[resolved_image.registry]
```
mprahl, line 950: @lcarva in the case of redhat-marketplace, the "rhm" namespace is added. Would I need to prefix the existing namespace with `rhm/` or is it fine to set the registry value to have `/rhm` at the end?

lcarva: That's a good question. It looks like `ImageName` will allow you to do either. Let's go with the easiest approach for now.

iib/workers/tasks/build.py
```diff
@@ -792,7 +799,83 @@ def _copy_files_from_image(image, src_path, dest_path):
...
+
+    with open(annotations_yaml_path, 'r') as f:
+        try:
+            annotations_yaml = yaml.safe_load(f)
+        except yaml.YAMLError:
+            error = 'metadata/annotations/yaml is not valid YAML'
+            log.exception(error)
+            raise IIBError(error)
+
+    package_label = 'operators.operatorframework.io.bundle.package.v1'
```
lcarva, line 830: This seems out of place. Should we move it to just above the line where it gets first used?

iib/workers/tasks/build.py
```diff
@@ -792,7 +799,83 @@ def _copy_files_from_image(image, src_path, dest_path):
...
+    if not isinstance(annotations_yaml.get('annotations', {}), dict):
```
lcarva, line 831: I find it difficult to follow what the code is trying to do when there's so much validation around it. Utility methods can help abstract some of the implementation details and help readers focus on what the code is trying to achieve.

```python
package_name = get_package_name_from_annotations(annotations_yaml)
package_name_with_suffix = apply_package_name_suffix(package_name)
```

There's nothing wrong with the approach you have taken. The code works and it's maintainable. This is just a suggestion to make it more readable which always has a subjective aspect to it.

iib/workers/tasks/build.py
```diff
@@ -792,7 +799,83 @@ def _copy_files_from_image(image, src_path, dest_path):
...
+        return package_annotation, {}
+
+    if package_annotation.endswith(package_name_suffix):
+        log.debug('No modifications are needed on %s in metadata/annotations.yaml', package_label)
+        return package_annotation, {}
+
+    annotations_yaml['annotations'][package_label] = f'{package_annotation}{package_name_suffix}'
+
+    with open(annotations_yaml_path, 'w') as f:
+        yaml.safe_dump(annotations_yaml, f)
```
lcarva, line 863: Consider using `ruamel` as it retains comments in the yaml. We used that in OMPS for this reason. It avoids surprises to users.

iib/workers/tasks/build.py
```diff
@@ -806,8 +889,15 @@ def _adjust_operator_manifests(manifests_path):
...
 
     :param str manifests_path: the full path to the directory containing the operator manifests.
+    :param str metadata_path: the full path to the directory containing the bundle metadata files.
+    :param str organization: the organization this bundle is for. If no organization is provided,
+        no custom behavior will be applied.
     :raises IIBError: if the operator manifest has invalid entries
+    :return: a dictionary of labels to set on the bundle
+    :rtype: dict
     """
+    package_name, labels = _add_package_name_suffix(metadata_path, organization)
```
lcarva, line 899: Maybe `_apply_package_name_suffix` is more descriptive?

iib/workers/tasks/build.py
```diff
@@ -833,20 +923,94 @@ def _adjust_operator_manifests(manifests_path):
...
+
     # Resolve pull specs to container image digests
     replacement_pullspecs = {}
     for pullspec in found_pullspecs:
-        # Skip images that are already pinned
-        if ':' not in ImageName.parse(pullspec).tag:
-            replacement_pullspecs[pullspec] = ImageName.parse(_get_resolved_image(pullspec))
+        replacement_needed = False
+        if ':' in ImageName.parse(pullspec).tag:
+            resolved_image = ImageName.parse(pullspec)
```
lcarva, line 938: I wonder if it makes sense to always use `_get_resolved_image` to ensure the pull spec is valid?

iib/workers/tasks/build.py
```diff
@@ -833,20 +923,94 @@ def _adjust_operator_manifests(manifests_path):
...
+        else:
+            resolved_image = ImageName.parse(_get_resolved_image(pullspec))
+            replacement_needed = True
+
+        if registry_replacements.get(resolved_image.registry):
+            replacement_needed = True
+            log.debug(
```
lcarva, line 945: Could we move this log message to after the registry replacement operation and log the before and after pull spec? The method `ImageName.to_str()` should give you the full pull spec.

iib/workers/tasks/build.py
```diff
@@ -833,20 +923,94 @@ def _adjust_operator_manifests(manifests_path):
...
+    """
+    conf = get_worker_config()
+    org_csv_annotations = (
+        conf['iib_organization_customizations'].get(organization, {}).get('csv_annotations')
+    )
+    if not org_csv_annotations:
+        log.debug('The organization %s does not have CSV annotations configured', organization)
+        return
+
+    glob_pattern = os.path.join(manifests_path, '*.clusterserviceversion.yaml')
```
lcarva, line 989: There's nothing that enforces this naming format. We must read each yaml file, `.yaml` or `.yml`, and determine which ones are CSVs. The operator-manifest library should be used here.

```
operator_manifest = OperatorManifest.from_directory(manifests_path)
for operator_csv in operator_manifest.files:     
  ...                  
```

`_adjust_operator_manifests` is already doing this.

iib/workers/config.py
```diff
@@ -104,6 +121,49 @@ def validate_celery_config(conf, **kwargs):
     if not isinstance(conf['iib_required_labels'], dict):
         raise ConfigError('iib_required_labels must be a dictionary')
 
+    if not isinstance(conf['iib_organization_customizations'], dict):
```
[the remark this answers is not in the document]
mprahl: I tried using `jsonschema` and it doesn't raise a validation error when the keys are not strings. The error messages are not as nice either.

I'll skip this change.

iib/workers/tasks/build.py
mprahl: I think it's fine the way it is. If you feel strongly about this, we can discuss this further.

iib/workers/tasks/build.py
mprahl: I avoided it because `skopeo inspect` without `--raw` is quite expensive but I added it based on your recommendation in a separate commit.

mprahl:
@lcarva thanks! This is ready for review again.

README.md
```diff
@@ -214,13 +214,18 @@ The custom configuration options for the Celery workers are listed below:
...
-  and `registry_replacements`. The `csv_annotations` value is a dictionary where each key is
-  the annotation to set on the ClusterServiceVersion files, and the value is a Python template
-  string of the value to be set. IIB only substitutes `{package_name}` in the template string. The
-  `package_name_suffix` value is a string of a suffix to add to the package name of the operator.
-  The `registry_replacements` value is a dictionary where the keys are the old registries to replace
-  and the values are the registries to replace the old registries with. This replaces the registry
-  in all the ClusterServiceVersion files. Here is an example that ties this all together:
+  and `registry_replacements`.
+
+  * The `csv_annotations` value is a dictionary where each key is the annotation to set on the
```
yashvardhannanavati, line 219: Would it make sense to add here if any of the keys are optional?

iib/workers/tasks/build.py
```diff
@@ -957,4 +957,36 @@ def _adjust_operator_bundle(manifests_path, metadata_path, organization=None):
...
+        _adjust_csv_annotations(operator_manifest.files, package_name, organization)
+
     return labels
+
+
+def _adjust_csv_annotations(operator_csvs, package_name, organization):
+    """
+    Annotate ClusterServiceVersion objects based on an organization configuration.
+
+    :param list operator_csvs: the list of ``OperatorCSV`` objects to examine
```
yashvardhannanavati, line 970: Missing `.` in the end. :)

docs/requirements.txt
```diff
@@ -8,6 +8,7 @@ python-qpid-proton
 psycopg2-binary
 recommonmark
 requests
+ruamel.yaml
```
yashvardhannanavati, line 11: Based on the recommendation here (https://yaml.readthedocs.io/en/latest/basicuse.html) Should this be pinned to a version <0.15?

requirements-test.txt
```diff
@@ -314,7 +314,7 @@ ruamel.yaml.clib==0.2.0 \
 ruamel.yaml==0.16.10 \
```
yashvardhannanavati, line 314: Same as the comment about the recommendation in their docs

docs/requirements.txt
mprahl: The version less than 0.15 is 3 years old. Also, the operator-manifests library uses 0.16.x which does most of the modifications. Also, when installing that old version, I get a traceback on Python 3.7. Lastly, the version in Fedora 31 is 0.16.5, so I'll assume that it's safe to use.

requirements-test.txt
mprahl: Same as above.

PR #87, opened 2020-05-12
Revert "Use skopeo copy for pushing image"

lcarva:


mprahl:
This was needed because `skopeo` would fail when pushing the image to quay.io with the error `invalid manifest`.

PR #89, opened 2020-05-13
Fix the multiple URL support when connecting to an AMQP 1.0 broker

mprahl:
The corresponding changes were pushed to https://github.com/apache/qpid-proton/pull/243. See the diff in that PR to see what's actually new code.

### DIFF
--- a/iib/web/messaging.py
+++ b/iib/web/messaging.py
@@ -2,6 +2,7 @@
 from collections import namedtuple
 import json
 import os
+import time
 import uuid
 
 from flask import current_app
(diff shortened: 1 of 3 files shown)


iib/web/messaging.py
```diff
@@ -59,6 +63,56 @@ def __init__(
...
+        :type msg: ``str``
+        """
+        if timeout is False:
+            timeout = self.timeout
+        if timeout is None:
+            while not condition() and not self.disconnected:
+                self.container.process()
+        else:
+            container_timeout = self.container.timeout
+            self.container.timeout = timeout
```
yashvardhannanavati, line 86: Should we check for the type of ‘timeout’ ? What if someone passes in a string for instance?

mprahl: @yashvardhannanavati I didn't write this part of the code. The only code I added was https://github.com/apache/qpid-proton/pull/243.

I'm trying to keep the code as close to upstream as possible.

lcarva: +1 for minimizing the changes.

yashvardhannanavati: Agreed! But we are anyways never going to be exactly at par with the upstream code. FWIW, I feel it might be okay to add a couple of lines. But if Luiz and you feel differently, I am happy to let it go.

mprahl: @yashvardhannanavati the idea is that if the PR upstream gets merged, then we would just use the library and make any customizations in IIB.

PR #93, opened 2020-05-18
Add organizations and request types to the batch state change messages

mprahl:
This also removes the obsolete `request_ids` array in the batch state change messages.

### DIFF
--- a/iib/web/messaging.py
+++ b/iib/web/messaging.py
@@ -160,6 +160,14 @@ def _get_batch_state_change_envelope(batch, new_batch=False):
         content = {
             'batch': batch.id,
             'annotations': batch.annotations,
+            'requests': [
+                {
+                    'id': request.id,
+                    'organization': getattr(request, 'organization', None),
+                    'type': request.type_name,
+                }
+                for request in batch.requests
+            ],
             'request_ids': sorted(batch.request_ids),
             'state': batch_state,
             'user': batch_username,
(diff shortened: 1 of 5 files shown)


iib/web/messaging.py
```diff
@@ -160,7 +160,9 @@ def _get_batch_state_change_envelope(batch, new_batch=False):
         content = {
             'batch': batch.id,
             'annotations': batch.annotations,
+            'organizations': sorted(batch.organizations),
             'request_ids': sorted(batch.request_ids),
```
lcarva, line 164: Can we turn `request_ids` into a list of objects that only contain cherry-picked attributes? Otherwise the data for each request is split across 3 different fields without any sort of mapping between them:

```
"requests": [
 {"id": 1, "organization": "spam-operators", "request_type": "add"},
 {"id": 1, "organization": null, "request_type": "rm"},
 {"id": 1, "organization": "foo-operators", "request_type": "regenerate-bundle"},
]
```

Either approach should be confirmed with the requester of this feature.

mprahl: What is the use-case where it's useful to know which request ID maps to which organization and request type?

lcarva: If the batch contains multiple types of requests and we only care about certain types, we have to query IIB. What's the reasoning for splitting them up?

mprahl: Well there's actually only one kind of request type per batch, but I made it a list in the event we need to have a batch contain multiple request types.

PR #96, opened 2020-05-26
Create log files per request

lcarva:

### DIFF
--- a/README.md
+++ b/README.md
@@ -153,6 +153,10 @@ The custom configuration options for the REST API are listed below:
   This defaults to `20`.
 * `IIB_PRIVILEGED_USERNAMES` - the list of users that can perform privileged actions such
   as overwriting the input index image with the built index image. This defaults to `[]`.
+* `IIB_REQUEST_LOGS_DIR` - the directory to load the request specific log files. If `None`, per
+  request log files information will not appear in the API response. This defaults to `None`.
+* `IIB_REQUEST_LOGS_DAYS_TO_LIVE` - the amount of days after which per request logs are considered
+  to be expired and may be removed. This defaults to `3`.
 * `IIB_USER_TO_QUEUE` - the mapping, `dict(<str>: <str>)`, of usernames to celery task queues.
   This is useful in isolating the workload from certain users. The default queue is used for tasks
   from users not found in the mapping. This defaults to `{}`.
@@ -247,6 +251,12 @@ The custom configuration options for the Celery workers are listed below:
   }
   ```
 
+* `iib_request_logs_dir` - the directory to write the request specific log files. If `None`, per
+  request log files are not created. This defaults to `None`.
+* `iib_request_logs_format` - the format for the log messages of the request specific log files.
+  This defaults to `%(asctime)s %(name)s %(levelname)s %(module)s.%(funcName)s %(message)s`.
+* `iib_request_logs_level` - the log level for the request specific log files. This defaults to
+  `DEBUG`.
 * `iib_registry` - the container registry to push images to (e.g. `quay.io`).
 * `iib_skopeo_timeout` - the command timeout for skopeo commands run by IIB. This defaults to
   `30s` (30 seconds).
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -57,6 +57,7 @@ services:
     volumes:
       - ./:/src:z
       - ./docker/message_broker/certs:/broker-certs:ro,z
+      - request-logs-volume:/var/log/iib/requests:z
     ports:
       - 8080:8080
     depends_on:
@@ -90,6 +91,7 @@ services:
       - ./iib_data/worker:/var/lib/containers/storage:z
       - ./docker/registry/certs:/registry-certs:ro,z
       - /etc/pki/tls/certs/ca-bundle.crt:/host-ca-bundle.crt:ro
+      - request-logs-volume:/var/log/iib/requests:z
     depends_on:
       - rabbitmq
       - registry
@@ -109,3 +111,4 @@ services:
 
 volumes:
   message-broker-volume:
+  request-logs-volume:
--- a/iib/web/api_v1.py
+++ b/iib/web/api_v1.py
@@ -1,11 +1,13 @@
 # SPDX-License-Identifier: GPL-3.0-or-later
 import copy
+import os
+from datetime import datetime
 
 import flask
 from flask_login import current_user, login_required
 from sqlalchemy.orm import with_polymorphic
 from sqlalchemy.sql import text
-from werkzeug.exceptions import Forbidden
+from werkzeug.exceptions import Forbidden, Gone, NotFound
 
 from iib.exceptions import IIBError, ValidationError
 from iib.web import db, messaging
@@ -46,6 +48,36 @@ def get_build(request_id):
     poly_request = with_polymorphic(Request, '*')
     query = poly_request.query.options(*get_request_query_options(verbose=True))
     return flask.jsonify(query.get_or_404(request_id).to_json())
+
+
+@api_v1.route('/builds/<int:request_id>/logs')
+def get_build_logs(request_id):
+    """
+    Retrieve the logs for the build request.
+
+    :param int request_id: the request ID that was passed in through the URL.
+    :rtype: flask.Response
+    :raise NotFound: if the request is not found or there are no logs for the request
+    :raise Gone: if the logs for the build request have been removed due to expiration
+    """
+    request_log_dir = flask.current_app.config['IIB_REQUEST_LOGS_DIR']
+    if not request_log_dir:
+        raise NotFound()
+
+    request = Request.query.get_or_404(request_id)
+    log_file_path = os.path.join(request_log_dir, f'{request_id}.log')
+    if not os.path.exists(log_file_path):
+        expired = request.logs_expiration < datetime.utcnow()
+        if expired:
+            raise Gone(f'The logs for the build request {request_id} no longer exist')
+        finalized = request.state.state_name in RequestStateMapping.get_final_states()
+        if finalized:
+            raise NotFound()
+        # The request may not have been initiated yet. Return empty logs until it's processed.
+        return flask.Response('', mimetype='text/plain')
+
+    with open(log_file_path) as f:
+        return flask.Response(f.read(), mimetype='text/plain')
 
 
 @api_v1.route('/builds')
--- a/iib/web/models.py
+++ b/iib/web/models.py
@@ -1,9 +1,10 @@
 # SPDX-License-Identifier: GPL-3.0-or-later
 from copy import deepcopy
+from datetime import timedelta
 from enum import Enum
 import json
 
-from flask import current_app
+from flask import current_app, url_for
 from flask_login import UserMixin, current_user
 import sqlalchemy
 from sqlalchemy.ext.declarative import declared_attr
@@ -373,6 +374,11 @@ class Request(db.Model):
             states = list(reversed(states))
             rv['state_history'] = states
             latest_state = states[0]
+            if current_app.config['IIB_REQUEST_LOGS_DIR']:
+                rv['logs'] = {
+                    'expiration': self.logs_expiration.isoformat() + 'Z',
+                    'url': url_for('.get_build_logs', request_id=self.id, _external=True),
+                }
         rv.update(latest_state or _state_to_json(self.state))
 
         return rv
@@ -395,6 +401,17 @@ class Request(db.Model):
         :rtype: str
         """
         return RequestTypeMapping.pretty(self.type)
+
+    @property
+    def logs_expiration(self):
+        """
+        Return the timestamp of when logs are considered expired.
+
+        :return: logs expiration timestamp
+        :rtype: str
+        """
+        logs_lifetime = timedelta(days=current_app.config['IIB_REQUEST_LOGS_DAYS_TO_LIVE'])
+        return self.state.updated + logs_lifetime
 
 
 class Batch(db.Model):
--- a/iib/web/static/api_v1.yaml
+++ b/iib/web/static/api_v1.yaml
@@ -207,6 +207,45 @@ paths:
                   error:
                     type: string
                     example: Database health check failed
+  '/builds/{id}/logs':
+    get:
+      description: Return the logs for a specific build request
+      parameters:
+        - name: id
+          in: path
+          required: true
+          description: The ID of the build request to retrieve the logs for
+          schema:
+            type: integer
+      responses:
+        '200':
+          description: The logs for the build request
+          content:
+            text/plain:
+              schema:
+                type: string
+              example: |-
+                Processing build request 1...
+                Building image...
+                Done processing build request 1
+        '404':
+          description: Logs for build requests is not enabled in IIB
+            application/json:
+              schema:
+                type: object
+                properties:
+                  error:
+                    type: string
+                    example: The requested resource was not found
+        '410':
+          description: The logs for the build request has been removed due to expiration
+            application/json:
+              schema:
+                type: object
+                properties:
+                  error:
+                    type: string
+                    example: The logs for the build request 1 no longer exist
   /builds/add:
     post:
       description: Submit a build request to add operator bundles to an index image
@@ -666,6 +705,21 @@ components:
               type: array
               items:
                 $ref: '#/components/schemas/StateHistory'
+            logs:
+              type: object
+              properties:
+                url:
+                  type: string
+                  description: URL to fetch the logs for the request
+                expiration:
+                  type: string
+                  description: >-
+                    UTC timestamp in ISO format that indicates when the logs for the request are
+                    considered to be expired and will be removed.
+              example:
+                url: https://iib.domain.local/api/v1/builds/1/logs
+                expiration: '2019-09-19T19:35:15.722265Z'
+
     RequestIndexImageVerbose:
       allOf:
         - $ref: '#/components/schemas/RequestVerbose'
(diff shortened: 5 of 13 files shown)


lcarva:
This is definitely not anywhere near completion, but I'm interested in what you think of this approach so far.

iib/workers/tasks/utils.py
```diff
@@ -141,3 +144,30 @@ def run_cmd(cmd, params=None, exc_msg=None, cmd_repr=None):
         raise IIBError(exc_msg)
 
     return response.stdout
+
+
+def request_logger(request_id_index=2):
```
lcarva, line 149: I can't think of a nicer way to grab the request ID. I made this configurable to hopefully reduce some of the magic.

mprahl:
@lcarva how about sub-classing `logging.FileHandler` like MBS does:
```python
class ModuleBuildFileHandler(logging.FileHandler):
    """
    FileHandler subclass which handles only messages generated during
    particular module build with `build_id` set in its constructor.
    """

    def __init__(self, build_id, filename, mode="a", encoding=None, delay=0):
        logging.FileHandler.__init__(self, filename, mode, encoding, delay)
        self.build_id = build_id

    def emit(self, record):
        # Imported here because of importing cycle between __init__.py,
        # scheduler and models.
        from module_build_service.scheduler.consumer import MBSConsumer

        # Check the currently handled module build and emit the message only
        # if it's associated with current module.
        build_id = MBSConsumer.current_module_build_id
        if not build_id or build_id != self.build_id:
            return

        logging.FileHandler.emit(self, record)
```

You can see how it's used in https://pagure.io/fm-orchestrator/blob/master/f/module_build_service/common/logger.py. I'm not recommending you do exactly it does, but using that custom file handler is a nice approach. You add the handler at the beginning of the request and then remove it at the end of the request.

iib/workers/tasks/utils.py
mprahl: If you're not opposed to magic, we did something similar in MBS to find a specific argument value of a handler:
https://pagure.io/fm-orchestrator/blob/master/f/module_build_service/scheduler/route.py#_33

mprahl:
@lcarva how will the log files be exposed to the user since the logs are generated on the worker and they are not always on the same host?

lcarva:
> @lcarva how will the log files be exposed to the user since the logs are generated on the worker and they are not always on the same host?

They'd have to share a volume. We can find a more sophisticated approach if that's not sufficient.

iib/workers/tasks/utils.py
lcarva: I'll look into this. It's better magic than what I have :)

lcarva: Done.

lcarva:
The code is ready. I'll work on the unit tests now.

midnightercz:
I'm just curious, what is the motivation for this change?

lcarva:
> I'm just curious, what is the motivation for this change?

Stakeholder request in CLOUDBLD-591.

lcarva:
This is now ready for a full review. Thanks in advance!

README.md
```diff
@@ -148,6 +148,10 @@ The custom configuration options for the REST API are listed below:
   This defaults to `20`.
 * `IIB_PRIVILEGED_USERNAMES` - the list of users that can perform privileged actions such
   as overwriting the input index image with the built index image. This defaults to `[]`.
+* `IIB_REQUEST_LOGS_DIR` - the directory to load the request specific log files. If `None`, per
+  request log files information will not appear in the API response. This defaults to `None`.
+* `IIB_REQUEST_LOGS_DAYS_TO_LIVE` - the amount of days after which per request logs are considered
+  to be expired and may be removed. This defaults to 3.
```
yashvardhannanavati, line 154: Should we default this to 5? Defaulting it to 3 would mean that logs for requests made on Friday would not be available on Monday.

docker-compose.yml
```diff
@@ -57,6 +57,7 @@ services:
     volumes:
       - ./:/src:z
       - ./docker/message_broker/certs:/broker-certs:ro,z
+      - request-logs-volume:/var/log/iib-requests:z
```
yashvardhannanavati, line 60: Should we keep this consistent with what we plan to have in the deployment?
`/var/log/iib/requests`

docker-compose.yml
```diff
@@ -90,6 +91,7 @@ services:
       - ./iib_data/worker:/var/lib/containers/storage:z
       - ./docker/registry/certs:/registry-certs:ro,z
       - /etc/pki/tls/certs/ca-bundle.crt:/host-ca-bundle.crt:ro
+      - request-logs-volume:/var/log/iib-requests:z
```
yashvardhannanavati, line 94: same as above

iib/web/api_v1.py
```diff
@@ -47,6 +49,35 @@ def get_build(request_id):
...
+    """
+    Retrieve the logs for the build request.
+
+    :param int request_id: the request ID that was passed in through the URL.
+    :rtype: flask.Response
+    :raise NotFound: if the request is not found or there are no logs for the request
+    """
+    request_log_dir = flask.current_app.config['IIB_REQUEST_LOGS_DIR']
+    if not request_log_dir:
+        raise NotFound()
```
yashvardhannanavati, line 63: Should this check be done before start-up?

iib/workers/tasks/utils.py
```diff
@@ -141,3 +143,58 @@ def run_cmd(cmd, params=None, exc_msg=None, cmd_repr=None):
         raise IIBError(exc_msg)
 
     return response.stdout
+
+
+def request_logger(func):
+    """
+    Log messages relevant to the current request to a dedicated file.
+
+    If iib_request_logs_dir is set, a temporary log hanlder is added before the decorated
```
yashvardhannanavati, line 152: `log hanlder` => `log handler`

iib/workers/tasks/build.py
```diff
@@ -17,7 +17,14 @@
...
     validate_legacy_params_and_config,
 )
-from iib.workers.tasks.utils import get_image_labels, podman_pull, retry, run_cmd, skopeo_inspect
+from iib.workers.tasks.utils import (
+    get_image_labels,
+    podman_pull,
+    retry,
+    run_cmd,
+    skopeo_inspect,
+    request_logger,
```
mprahl, line 26: Optional: It'd be nice to alphabetize this

iib/workers/tasks/utils.py
mprahl, line 152:
```suggestion
    If ``iib_request_logs_dir`` is set, a temporary log hanlder is added before the decorated
```

iib/workers/tasks/utils.py
```diff
@@ -141,3 +143,58 @@ def run_cmd(cmd, params=None, exc_msg=None, cmd_repr=None):
...
+    function is invoked. It's then removed once the decorated function completes exection.
+
+    If iib_request_logs_dir is not set, the temporary log handler will not be added.
```
mprahl, line 155:
```suggestion
    If ``iib_request_logs_dir`` is not set, the temporary log handler will not be added.
```

iib/workers/tasks/utils.py
```diff
@@ -141,3 +143,58 @@ def run_cmd(cmd, params=None, exc_msg=None, cmd_repr=None):
...
+
+    :param function func: the function to be decorated. The function must take the request_id
+        parameter.
+    :return: the decorated function
+    :rtype: function
+    """
+    log_dir = get_worker_config().iib_request_logs_dir
```
mprahl, line 162: How about assigning the value of `get_worker_config()` to a variable to prevent executing it 3 times?

iib/workers/tasks/utils.py
```diff
@@ -141,3 +143,58 @@ def run_cmd(cmd, params=None, exc_msg=None, cmd_repr=None):
...
+    log_level = get_worker_config().iib_request_logs_level
+    log_formatter = logging.Formatter(get_worker_config().iib_request_logs_format)
+
+    @functools.wraps(func)
+    def wrapper(*args, **kwargs):
+        request_log_handler = None
+        if log_dir:
+            request_id = _get_function_arg_value('request_id', func, args, kwargs)
+            if not request_id:
+                log.warning('Unable to get "request_id" from %s', func)
```
mprahl, line 172: Shouldn't this be an error message?

iib/workers/tasks/utils.py
```diff
@@ -141,3 +143,58 @@ def run_cmd(cmd, params=None, exc_msg=None, cmd_repr=None):
...
+    If iib_request_logs_dir is not set, the temporary log handler will not be added.
+
+    :param function func: the function to be decorated. The function must take the request_id
+        parameter.
+    :return: the decorated function
+    :rtype: function
+    """
+    log_dir = get_worker_config().iib_request_logs_dir
+    log_level = get_worker_config().iib_request_logs_level
+    log_formatter = logging.Formatter(get_worker_config().iib_request_logs_format)
```
mprahl, line 164: It'd be nice to move this to under the `if log_dir:` branch to avoid computing this if per request logs are disabled.

tests/test_web/test_models.py
```diff
@@ -1,4 +1,5 @@
 # SPDX-License-Identifier: GPL-3.0-or-later
+from datetime import timedelta
 import pytest
```
mprahl, line 3:
```suggestion
from datetime import timedelta

import pytest
```

tests/test_workers/test_config.py
```diff
@@ -144,3 +144,32 @@ def test_validate_celery_config_invalid_organization_customizations(config, erro
     }
     with pytest.raises(ConfigError, match=error):
         validate_celery_config(conf)
+
+
+def test_validate_celery_config_request_logs_dir_missing():
+    missing_path = '/this/path/does/not/exist'
```
mprahl, line 150: Could you mock `os.path.exists` or use a path within `tmpdir` instead of this? I don't like the tests reaching out on the file system outside of the Git repo and `/tmp`.

README.md
yashvardhannanavati: Wouldn't that increase load on the DB?

mprahl: The logs aren't stored in the database, they are just text files on the file system. I think this is fine.

Ideally, the logs would be attached to the JIRA ticket, but they may not always be. Also, the ticket may be groomed by someone that doesn't know to do that.

lcarva: I'll leave it as 3 by default and we can adjust our config as needed.

iib/web/models.py
```diff
@@ -373,6 +374,11 @@ def _state_to_json(state):
             states = list(reversed(states))
             rv['state_history'] = states
             latest_state = states[0]
+            if current_app.config['IIB_REQUEST_LOGS_DIR']:
+                rv['logs'] = {
+                    'expiration': self.logs_expiration.isoformat() + 'Z',
+                    'url': url_for('.get_build_logs', request_id=self.id, _external=True),
```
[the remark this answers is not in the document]
lcarva: Should we hard code the api version in the model? This seems to work without specifying it which makes me wonder that it will use whichever api version is being used in the request.

iib/web/static/api_v1.yaml
```diff
@@ -183,6 +183,48 @@ paths:
...
+          description: The logs for the build request has been removed due to expiration
+          content:
+            application/json:
+              schema:
+                type: object
+                properties:
+                  error:
+                    type: string
+                    example: The logs for the build request 1 no longer exists
+
```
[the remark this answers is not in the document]
lcarva: I'll remove the line to keep it consistent, but this makes it really hard to navigate this file!

tests/test_workers/test_config.py
lcarva: I can't get this check to work if the directory is created by python. https://docs.python.org/3/library/os.html#os.access

There's also a security warning there.

There's value in doing this check though. I'll dig more.

lcarva: Ha! It's because I was running the tests with sudo and we can't restrict access of the root user. I'll add a conditional skip test in there.

iib/web/models.py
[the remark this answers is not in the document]
mprahl: If you can confirm that, then we can keep it as is.

iib/web/static/api_v1.yaml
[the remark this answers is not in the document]
mprahl: You can add a separate commit with the new lines then.

iib/web/api_v1.py
```diff
@@ -47,6 +49,35 @@ def get_build(request_id):
...
+
+    query = Request.query.options(*get_request_query_options(verbose=True))
```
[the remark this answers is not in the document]
lcarva: Thank you for making this suggestion. I've set `SQLALCHEMY_ECHO = True` in `iib/web/config.py::DevelopmentConfig`. This causes all the SQL queries to be logged which is super useful for this case.

I tried four different approaches:

1. `query = Request.query.options(joinedload(Request.state))`, which resulted in `SELECT ... FROM request LEFT OUTER JOIN request_state AS request_state_1 ON request_state_1.id = request.request_state_id WHERE request.id = %(param_1)s`

1. `query = Request.query`, `SELECT ... FROM request WHERE request.id = %(param_1)s`                                                                                                
                                                                                                
1. `query = Request.query.options(*get_request_query_options(verbose=True))` -> `SELECT ... FROM request LEFT OUTER JOIN request_state AS request_state_1 ON request.id = request_state_1.request_id LEFT OUTER JOIN "user" AS user_1 ON user_1.id = request.user_id WHERE request.id = %(param_1)s ORDER BY request_state_1.updated`

1. `query = Request.query.options(*get_request_query_options(verbose=False))` -> `SELECT ... FROM request LEFT OUTER JOIN request_state AS request_state_1 ON request_state_1.id = request.request_state_id LEFT OUTER JOIN "user" AS user_1 ON user_1.id = request.user_id WHERE request.id = %(param_1)s`

No other queries were done when hitting the logs api.

Let's go with option number two as it's the simplest and most efficient.

iib/web/api_v1.py
```diff
@@ -47,6 +49,35 @@ def get_build(request_id):
...
+    request = query.get_or_404(request_id)
+
+    log_file_path = os.path.join(request_log_dir, f'{request_id}.log')
+    if not os.path.exists(log_file_path):
+        expired = request.logs_expiration < datetime.utcnow()
+        if expired:
+            raise Gone(f'The logs for the build request {request_id} no longer exists')
+        else:
+            # The request may not have been initiated yet. Return empty logs until it's processed.
+            return flask.Response('', mimetype='text/plain')
```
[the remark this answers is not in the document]
lcarva: Great idea! Let me know what you think of the way I did it.

iib/workers/tasks/utils.py
lcarva: I've decided to actually raise an error here.

lcarva:
@mprahl, @yashvardhannanavati this is ready for another review. Thank you!

iib/workers/tasks/utils.py
```diff
@@ -141,3 +143,61 @@ def run_cmd(cmd, params=None, exc_msg=None, cmd_repr=None):
...
 
     return response.stdout
+
+
+def request_logger(func):
+    """
+    Log messages relevant to the current request to a dedicated file.
+
+    If ``iib_request_logs_dir`` is set, a temporary log handler is added before the decorated
+    function is invoked. It's then removed once the decorated function completes exection.
```
mprahl, line 153:
```suggestion
    function is invoked. It's then removed once the decorated function completes execution.
```

iib/workers/tasks/utils.py
```diff
@@ -141,3 +143,61 @@ def run_cmd(cmd, params=None, exc_msg=None, cmd_repr=None):
...
+
+    If ``iib_request_logs_dir`` is not set, the temporary log handler will not be added.
+
+    :param function func: the function to be decorated. The function must take the request_id
```
mprahl, line 157:
```suggestion
    :param function func: the function to be decorated. The function must take the ``request_id``
```

iib/workers/tasks/utils.py
```diff
@@ -141,3 +143,61 @@ def run_cmd(cmd, params=None, exc_msg=None, cmd_repr=None):
...
+    log_format = worker_config.iib_request_logs_format
+
+    @functools.wraps(func)
+    def wrapper(*args, **kwargs):
+        request_log_handler = None
+        if log_dir:
+            log_formatter = logging.Formatter(log_format)
+            request_id = _get_function_arg_value('request_id', func, args, kwargs)
+            if not request_id:
+                raise IIBError(f'Unable to get "request_id" from {func}')
```
mprahl, line 174:
```suggestion
                raise IIBError(f'Unable to get "request_id" from {func.__name__}')
```

iib/workers/tasks/utils.py
```diff
@@ -141,3 +143,61 @@ def run_cmd(cmd, params=None, exc_msg=None, cmd_repr=None):
...
+            else:
```
mprahl, line 175: This else is no longer needed since you raise an exception in the `if` branch.

iib/workers/tasks/utils.py
```diff
@@ -141,3 +143,61 @@ def run_cmd(cmd, params=None, exc_msg=None, cmd_repr=None):
...
+                log_file_path = os.path.join(log_dir, f'{request_id}.log')
+                request_log_handler = logging.FileHandler(log_file_path)
```
mprahl, line 177: Does this create an empty file automatically?

lcarva:
> LGTM!
> 
> Please remember to file a ticket that would actually perform the cleanup when the logs have expired.

Filed it internally as CLOUDBLD-1477.

PR #97, opened 2020-05-27
Create health check endpoint for semaphore [CLOUDDST-770]

querti:
The created endpoint will return 200 and the message 'Health check OK'. Its primary usage will be by the semaphore service which will periodically ping the endpoint and report the status on its website.


querti:
@release-engineering/distribution

iib/web/api_v1.py
```diff
@@ -90,6 +90,16 @@ def get_builds():
...
 
 
+@api_v1.route('/healthcheck')
+def get_healthcheck():
+    """
+    Respond to a health check.
+
+    :rtype: str
+    """
+    return "Health check OK"
```
mprahl, line 100: This is misleading since this API endpoint actually performs no checks. It doesn't verify that the database connection is successful or that it can communicate with the workers.

Even getting the status code of `/api/v1/builds?per_page=0` would be quick and provide a more useful indicator for health.

querti: I added some checks.

lipoja: Every response from API is in JSON format except this one. I would suggest to response in JSON format if it is possible.

querti: The semaphore service performs a regex check on the response message. Text response works much better for this.

iib/web/api_v1.py
```diff
@@ -90,6 +92,27 @@ def get_builds():
...
+@api_v1.route('/healthcheck')
+def get_healthcheck():
+    """
+    Respond to a health check.
+
+    :rtype: str
+    """
+    # Test DB connection
+    try:
+        db.engine.execute(text("SELECT 1"))
```
mprahl, line 104: Please use single quotes throughout your contribution.

iib/web/api_v1.py
```diff
@@ -90,6 +92,27 @@ def get_builds():
...
+    except Exception:
+        return flask.jsonify({"text": "Database health check failed"})
```
mprahl, line 106: This will return a 200 status code. I would not expect this if the healthcheck failed.

My suggestion is to raise an `IIBError` exception and this will automatically return a `500` status code and the JSON of `{"error": exception_msg}`.

I'm not sure 500 is the best error code but it's not a bad one in this case. I'm open to suggestions.

iib/web/api_v1.py
```diff
@@ -90,6 +92,27 @@ def get_builds():
...
+
+    # Ping the workers. One heartbeat from a worker is considered sufficient. 5 second timeout.
+    result = app.control.heartbeat(reply=True, limit=1, timeout=5)
```
mprahl, line 109: If all the workers are busy running tasks, will they respond? This seems like a normal scenario which shouldn't cause the health check to fail.

If you are unable to verify this, you can just remove this check.

iib/web/api_v1.py
```diff
@@ -90,6 +92,27 @@ def get_builds():
...
+    if result == []:
+        return flask.jsonify({"text": "Worker health check failed"})
```
mprahl, line 111: Same as above

iib/web/api_v1.py
midnightercz: Do you know what mime type sends semaphore in accept header? If it's expecting text/plain we can send output based on that

querti: I changed it to JSON and loosened the semaphore regex a bit. It shouldn't be an issue

iib/web/api_v1.py
querti: Fixed.

iib/web/api_v1.py
querti: Done.

iib/web/api_v1.py
querti: I found this in the docs: https://docs.celeryproject.org/en/stable/userguide/workers.html#remote-control

It seems that expecting a heartbeat message is not very reliable, so I removed it.

iib/web/api_v1.py
querti: Done.

iib/web/api_v1.py
```diff
@@ -90,6 +91,22 @@ def get_builds():
...
+
+    :rtype: str
+    """
+    # Test DB connection
+    try:
+        db.engine.execute(text('SELECT 1'))
+    except Exception:
+        raise IIBError('Database health check failed.')
+
+    return flask.jsonify({'text': 'Health check OK'})
```
midnightercz, line 107:
```suggestion
    return flask.jsonify({'status': 'Health check OK'})
```

tests/test_web/test_api_v1.py
```diff
@@ -134,6 +134,20 @@ def test_get_builds_invalid_batch(batch, app, client, db):
...
+    mock_db_execute.side_effect = Exception('DB failed')
+    rv = client.get('/api/v1/healthcheck')
+    assert rv.status_code == 500
+    assert rv.json == {'error': 'Database health check failed.'}
+
+
+def test_get_healthcheck_ok(app, client, db):
+    rv = client.get('/api/v1/healthcheck')
+    assert rv.status_code == 200
+    assert rv.json == {'text': 'Health check OK'}
```
midnightercz, line 148:
```suggestion
    assert rv.json == {'status': 'Health check OK'}
```

iib/web/api_v1.py
```diff
@@ -90,6 +91,22 @@ def get_builds():
...
+    """
+    Respond to a health check.
+
+    :rtype: str
+    """
+    # Test DB connection
+    try:
+        db.engine.execute(text('SELECT 1'))
+    except Exception:
+        raise IIBError('Database health check failed.')
```
rohanpm, line 105: Do the details of the underlying exception ever get logged?  If we only pass "Database health check failed" to the caller and no details of the caught exception here are logged anywhere, it doesn't seem debuggable.

I don't see any logging, but it's possible it's somehow handled by app.register_error_handler called elsewhere.

iib/web/static/api_v1.yaml
```diff
@@ -183,6 +183,17 @@ paths:
...
       security:
         - Kerberos Authentication: []
+  /healthcheck:
+    get:
+      description: Perform a health check on IIB by pinging this endpoint and awaiting expected return message
+      responses:
+        '200':
+          description: The health check OK message
+          content:
+            text/html:
```
rohanpm, line 193: This isn't the right content type, is it?

iib/web/api_v1.py
querti: Changed.

tests/test_web/test_api_v1.py
querti: Changed.

iib/web/api_v1.py
querti: I added a log message. Maybe @yashvardhannanavati or @mprahl could comment if it seems appropriate to them.

iib/web/static/api_v1.yaml
querti: Fixed.

tests/test_web/test_api_v1.py
```diff
@@ -134,6 +134,20 @@ def test_get_builds_invalid_batch(batch, app, client, db):
     assert rv.json == {'error': 'The batch must be a positive integer'}
 
 
+@mock.patch('sqlalchemy.engine.base.Engine.execute')
+def test_get_healthcheck_db_fail(mock_db_execute, app, client, db):
+    mock_db_execute.side_effect = Exception('DB failed')
```
mprahl, line 139: Can you change this to `IIBError`?

querti: Sure, but why do we expect sqlalchemy to raise IIBError?

iib/web/api_v1.py
```diff
@@ -90,6 +91,23 @@ def get_builds():
...
+    """
+    Respond to a health check.
+
+    :rtype: str
+    """
+    # Test DB connection
+    try:
+        db.engine.execute(text('SELECT 1'))
+    except Exception as e:
+        flask.current_app.logger.error('DB test failed with the following exception: %s', str(e))
```
rohanpm, line 105: If the exception details were not logged appropriately before this change, they're still not logged appropriately - str(e) is throwing away the backtrace entirely.  str(e) alone could be something virtually useless like "KeyError: foo".

If you did logger.exception("DB test failed"), or logger.error("DB test failed", exc_info=True), it would log the exception including backtrace.  I think this is the correct way to log exceptions.

querti: Fixed.

tests/test_web/test_api_v1.py
mprahl: Sorry, I was just skimming through the diff and misread your mock. Your previous code was correct.

Preferably, please change this to an exception that SQLAlchemy would actually raise.

querti: I changed it to a more realistic sqlalchemy error.

iib/web/static/api_v1.yaml
```diff
@@ -183,6 +183,17 @@ paths:
               $ref: '#/components/schemas/RequestUpdate'
       security:
         - Kerberos Authentication: []
+  /healthcheck:
+    get:
+      description: Perform a health check on IIB by pinging this endpoint and awaiting expected return message
+      responses:
+        '200':
```
chandwanitulsi, line 190: Should the responses also include 500 error?
Otherwise LGTM

querti: You're right. Added.

iib/web/api_v1.py
```diff
@@ -90,6 +91,23 @@ def get_builds():
     return flask.jsonify(response)
 
 
+@api_v1.route('/healthcheck')
+def get_healthcheck():
+    """
+    Respond to a health check.
+
+    :rtype: str
```
mprahl, line 99:
```suggestion
    :rtype: str
    :raises IIBError: if the database connection fails
```

iib/web/api_v1.py
mprahl: This looks fine to me

iib/web/static/api_v1.yaml
```diff
@@ -183,6 +183,30 @@ paths:
               $ref: '#/components/schemas/RequestUpdate'
       security:
         - Kerberos Authentication: []
+  /healthcheck:
+    get:
+      description: Perform a health check on IIB by pinging this endpoint and awaiting expected return message
```
mprahl, line 188:
```suggestion
      description: Perform a health check of the service
```

mprahl:
I left two minor comments. Once addressed, it looks good to me.

iib/web/api_v1.py
yashvardhannanavati, line 99: Can we please also add `:return: json object representing the health of IIB.`

yashvardhannanavati, line 99: Also the `rtype` isn't string anymore. Could you please change it?

iib/web/static/api_v1.yaml
querti: Changed.

iib/web/api_v1.py
querti: Added.

querti: Changed.

querti: Changed, along with Yash's suggestions.

PR #100, opened 2020-05-29
Make the skopeo inspect calls more efficient

mprahl:


iib/workers/tasks/build.py
```diff
@@ -310,8 +311,13 @@ def _get_resolved_image(pull_spec):
...
     """
     log.debug('Resolving %s', pull_spec)
-    skopeo_output = skopeo_inspect(f'docker://{pull_spec}')
-    pull_spec_resolved = f'{skopeo_output["Name"]}@{skopeo_output["Digest"]}'
+    if '@' in pull_spec:
+        name = pull_spec.rsplit('@', 1)[0]
+    else:
+        name = pull_spec.rsplit(':', 1)[0]
+    skopeo_output = skopeo_inspect(f'docker://{pull_spec}', '--raw', return_json=False)
+    digest = hashlib.sha256(skopeo_output.encode('utf-8')).hexdigest()
```
lcarva, line 319: This is probably ok.

A little background. Ideally, the digest can be pulled from the API request header. However, this would mean teaching IIB how to handle registry auth directly, and that will not work with registries that don't set the header, i.e. Pulp.

iib/workers/tasks/utils.py
```diff
@@ -124,7 +129,7 @@ def run_cmd(cmd, params=None, exc_msg=None, cmd_repr=None):
     params.setdefault('stderr', subprocess.PIPE)
     params.setdefault('stdout', subprocess.PIPE)
 
-    log.debug('Running the command "%s"', cmd_repr)
+    log.debug('Running the command "%s"', ', '.join(cmd_repr))
```
lcarva, line 132: Don't we want to join by `' '` instead? `skopeo inspect docker://foobar` is more useful than `skopeo, inspect, docker://foobar`

iib/workers/tasks/build.py
```diff
@@ -310,8 +311,13 @@ def _get_resolved_image(pull_spec):
     :rtype: str
     """
     log.debug('Resolving %s', pull_spec)
-    skopeo_output = skopeo_inspect(f'docker://{pull_spec}')
-    pull_spec_resolved = f'{skopeo_output["Name"]}@{skopeo_output["Digest"]}'
+    if '@' in pull_spec:
+        name = pull_spec.rsplit('@', 1)[0]
+    else:
+        name = pull_spec.rsplit(':', 1)[0]
```
yashvardhannanavati, line 317: This `if-else` block is used at least 2 more times in the same file. Do you think it should be a helper method instead?

mprahl: I added a new commit which moves this logic to a separate function.

PR #104, opened 2020-06-10
Use the overwrite_from_index_token when pulling the from_index image

mprahl:
This allows IIB to interact with index images it does not have access to by default.

### DIFF
--- a/iib/workers/tasks/utils.py
+++ b/iib/workers/tasks/utils.py
@@ -267,7 +267,7 @@ def _get_function_arg_value(arg_name, func, args, kwargs):
     original_func = func
     while getattr(func, '__wrapped__', None):
         original_func = func.__wrapped__
-    argspec = inspect.getargspec(original_func).args
+    argspec = inspect.getfullargspec(original_func).args
 
     arg_index = argspec.index(arg_name)
     arg_value = kwargs.get(arg_name, None)
--- a/iib/workers/tasks/utils.py
+++ b/iib/workers/tasks/utils.py
@@ -1,4 +1,6 @@
 # SPDX-License-Identifier: GPL-3.0-or-later
+import base64
+from contextlib import contextmanager
 import functools
 import inspect
 import json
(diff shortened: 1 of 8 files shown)


iib/workers/tasks/utils.py
```diff
@@ -68,6 +73,80 @@ def inner(*args, **kwargs):
...
+            log.debug('The Docker config symlink at %s does not exist', docker_config_path)
+
+        conf = get_worker_config()
+        if os.path.exists(conf.iib_docker_config_template):
+            log.debug(
+                'Copying the Docker config from %s to %s',
+                conf.iib_docker_config_template,
+                docker_config_path,
+            )
+            shutil.copy2(conf.iib_docker_config_template, docker_config_path)
```
lcarva, line 131: Why copy if the file is always re-written a few lines below. Could you just read from `conf.iib_docker_config_template` and write to `docker_config_path` once after modifications?

iib/workers/tasks/utils.py
```diff
@@ -68,6 +73,80 @@ def inner(*args, **kwargs):
...
+    This context manager will reset the authentication to the way it was after it exits. If
+    ``token`` is falsy, this context manager will do nothing.
+
+    :param str token: the token in the format of ``username:password``
+    :param str container_image: the pull specification of the container image to parse to determine
+        the registry this token is for.
+    :return: None
+    :rtype: None
+    """
+    if not token:
```
lcarva, line 108: What happens if request A uses an overwrite token and the clean up fails to execute. When a new request B comes along, which doesn't use an overwrite token, the old authfile will be used, no?

mprahl: `_cleanup` is always called at the start up of the Celery tasks, which resets the Docker config to the original.

iib/workers/tasks/utils.py
mprahl: Good point. I'll fix that.

lcarva:
Tests need updating due to code change.

PR #106, opened 2020-06-12
Fix not iterable error in regenerate-bundle

lcarva:
This fixes the following error when running a regenerate-bundle build
request:

    TypeError: argument of type 'ImageName' is not iterable


mprahl:
@lcarva could you please fix the unit tests?

lcarva:
> @lcarva could you please fix the unit tests?

Fixed.

PR #108, opened 2020-06-16
Add support for overwriting index images that use the same registry as IIB

mprahl:
When an index image is built, IIB always pushes it to its configured
registry. Additionally, if `overwrite_from_index` is provided, IIB
will copy the index image from its registry over to the location of
`from_index`.

If `overwrite_from_index_token` is provided and `from_index` is
on the same registry as IIB's configured registry, then this will
fail since the `overwrite_from_index_token` token does not have
access to the source repository to copy from.

In this case, this change makes IIB first export the index image
in IIB's configured registry and repo to local OCI images using
its token and then push the local OCI images using the
`overwrite_from_index_token` token.

When buildah is updated in RHEL 8 and manifest-tool is no longer
needed, this workaround can be removed.

Resolves CLOUDBLD-1502

### DIFF
--- a/iib/workers/tasks/build.py
+++ b/iib/workers/tasks/build.py
@@ -176,15 +176,8 @@ def _update_index_image_pull_spec(
     """
     conf = get_worker_config()
     if from_index and overwrite_from_index:
-        output_message = f'Overwriting the index image {from_index} with {output_pull_spec}'
-        log.info(output_message)
+        _overwrite_from_index(request_id, output_pull_spec, from_index, overwrite_from_index_token)
         index_image = from_index
-        exc_msg = f'Failed to overwrite the input from_index container image of {index_image}'
-        args = [f'docker://{output_pull_spec}', f'docker://{index_image}']
-        state_reason = output_message
-        set_request_state(request_id, 'in_progress', state_reason)
-        with set_registry_token(overwrite_from_index_token, from_index):
-            _skopeo_copy(*args, copy_all=True, exc_msg=exc_msg)
     elif conf['iib_index_image_output_registry']:
         index_image = output_pull_spec.replace(
             conf['iib_registry'], conf['iib_index_image_output_registry'], 1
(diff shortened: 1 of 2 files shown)


iib/workers/tasks/build.py
```diff
@@ -436,6 +429,62 @@ def _opm_index_rm(base_dir, operators, binary_image, from_index, overwrite_from_
...
+            # means that IIB's token will likely not have access to read the `from_index` image.
+            # This means IIB must first export the manifest list and all the manifests locally and
+            # then overwrite the `from_index` image with the exported version using the user
+            # supplied token.
+            #
+            # When a newer version of buildah is available in RHEL 8, then that can be used instead
+            # of the manifest-tool to create the manifest list locally which means this workaround
+            # can be removed.
+            if output_pull_spec_registry == from_index_registry:
+                temp_dir = tempfile.TemporaryDirectory(prefix='cachito-')
```
lcarva, line 466: ahem... check the prefix

mprahl: How embarrassing :stuck_out_tongue:

iib/workers/tasks/build.py
```diff
@@ -436,6 +429,62 @@ def _opm_index_rm(base_dir, operators, binary_image, from_index, overwrite_from_
...
+            # When a newer version of buildah is available in RHEL 8, then that can be used instead
+            # of the manifest-tool to create the manifest list locally which means this workaround
+            # can be removed.
+            if output_pull_spec_registry == from_index_registry:
+                temp_dir = tempfile.TemporaryDirectory(prefix='iib-')
+                new_index_src = f'oci:{temp_dir.name}'
+                log.info(
+                    'The registry used by IIB (%s) is also the registry where from_index (%s) will '
+                    'be overwritten using the user supplied token. Will perform a workaround which '
+                    'will cause the manifest digests to change but the content is the same.',
```
yashvardhannanavati, line 471: Should this be mentioned in the state_reason so that the user is aware of the change in manifest digests?

mprahl: I see your point but the manifest digests are already going to be different since the index image was modified by either removing an operator or adding a bundle. This is logging the fact that the digests will be different between the version in IIB's configured repo and the overwritten `from_index`. It's more for administrators to be aware that something isn't broken.

PR #109, opened 2020-06-18
Fix typo in the API docs

yashvardhannanavati:


iib/web/static/api_v1.yaml
```diff
@@ -230,22 +230,24 @@ paths:
                 Done processing build request 1
         '404':
           description: Logs for build requests is not enabled in IIB
-            application/json:
```
lcarva, line 233: I think this line shouldn't change, but there should be a new line above it, `content:`, which has the same indentation as the line `description: ...`

PR #110, opened 2020-06-18
Add force_export API parameter

lcarva:
* CLOUDBLD-975

### DIFF
--- a/iib/web/models.py
+++ b/iib/web/models.py
@@ -810,8 +810,13 @@ class RequestAdd(Request, RequestIndexImageMixin):
             if not isinstance(request_kwargs[param], str):
                 raise ValidationError(f'"{param}" must be a string')
 
-        # Always remove cnr_token from request_kwargs since it's not stored in the database
+        if 'force_backport' in request_kwargs:
+            if not isinstance(request_kwargs['force_backport'], bool):
+                raise ValidationError('"force_backport" must be a boolean')
+
+        # Remove attributes that are not stored in the database
         request_kwargs.pop('cnr_token', None)
+        request_kwargs.pop('force_backport', None)
 
         cls._from_json(
             request_kwargs,
--- a/iib/workers/tasks/legacy.py
+++ b/iib/workers/tasks/legacy.py
@@ -51,19 +51,23 @@ def _get_base_dir_and_pkg_name(package_dir):
     return os.path.dirname(package_dir), os.path.basename(package_dir)
 
 
-def get_legacy_support_packages(bundles):
+def get_legacy_support_packages(bundles, request_id, force_backport=False):
     """
     Get the packages that must be pushed to the legacy application registry.
 
     :param list<str> bundles: a list of strings representing the pull specifications of the bundles
         to add to the index image being built.
+    :param int request_id: the ID of the IIB build request.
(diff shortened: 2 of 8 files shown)


iib/workers/tasks/legacy.py
```diff
@@ -51,19 +51,22 @@ def _get_base_dir_and_pkg_name(package_dir):
...
     Get the packages that must be pushed to the legacy application registry.
 
     :param list<str> bundles: a list of strings representing the pull specifications of the bundles
         to add to the index image being built.
+    :param bool force_backport: if True, backport legacy support is forced for every package
     :return: a set of packages that require legacy support
     :rtype: set
     """
+    if force_backport:
+        log.info('Backport legacy support will be forced')
```
yashvardhannanavati, line 65: Do you think it might be helpful to have the state_reason should that the backport legacy support is being forced?

lcarva: Hmm that's a good point. Otherwise, this information would only be available in the logs. Should I do a new state, or maybe add a new parameter to build request representation (DB change)?

I think a state_reason should be enough.

lcarva: Done.

lcarva:
@yashvardhannanavati, could you please have another look?

yashvardhannanavati:
@lcarva Looks great! :ship:

iib/web/models.py
```diff
@@ -810,8 +810,13 @@ def from_json(cls, kwargs):
             if not isinstance(request_kwargs[param], str):
                 raise ValidationError(f'"{param}" must be a string')
 
-        # Always remove cnr_token from request_kwargs since it's not stored in the database
+        if 'force_backport' in request_kwargs:
+            if not isinstance(request_kwargs['force_backport'], bool):
```
mprahl, line 814: In the future, I prefer this syntax:
```python
if not isinstance(request_kwargs.get('force_backport', False), bool):
    raise ValidationError('"force_backport" must be a boolean')
```

iib/workers/tasks/legacy.py
```diff
@@ -51,19 +51,23 @@ def _get_base_dir_and_pkg_name(package_dir):
...
     :rtype: set
     """
+    if force_backport:
+        set_request_state(request_id, 'in_progress', 'Backport legacy support will be forced')
     packages = set()
     for bundle in bundles:
         labels = get_image_labels(bundle)
-        if labels.get('com.redhat.delivery.backport', False):
+        if force_backport or labels.get('com.redhat.delivery.backport', False):
             packages.add(labels['operators.operatorframework.io.bundle.package.v1'])
```
mprahl, line 71: We probably need to be more careful here and guard against this label not being set.

lcarva: Hmm that label defines the package name and is very much required regardless of backporting. Tests should catch this label not being set before IIB is invoked.

The `backport` label, on the other hand, is optional, and we're still handling if not set.

iib/web/models.py
lcarva: I do too, not sure why I didn't do that. Here's a PR to change it over: https://github.com/release-engineering/iib/pull/115

PR #112, opened 2020-06-23
Backport legacy packages before overwriting the index image

yashvardhannanavati:
Refers to CLOUDDST-1333

### DIFF
--- a/iib/workers/tasks/build.py
+++ b/iib/workers/tasks/build.py
@@ -775,6 +775,11 @@ def handle_add_request(
 
     set_request_state(request_id, 'in_progress', 'Creating the manifest list')
     output_pull_spec = _create_and_push_manifest_list(request_id, arches)
+    if legacy_support_packages:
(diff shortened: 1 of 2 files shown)


iib/workers/tasks/build.py
```diff
@@ -775,6 +775,11 @@ def handle_add_request(
 
     set_request_state(request_id, 'in_progress', 'Creating the manifest list')
     output_pull_spec = _create_and_push_manifest_list(request_id, arches)
+    if legacy_support_packages:
+        export_legacy_packages(
+            legacy_support_packages, request_id, output_pull_spec, cnr_token, organization
+        )
+
     _update_index_image_pull_spec(
```
lcarva, line 783: Can we add a test that if `export_legacy_packages` fails, then the `_update_index_image_pull_spec` method is not called?

yashvardhannanavati: Done

PR #116, opened 2020-06-30
Add delivery label to index images

### LINKED ISSUE #114: Provide a mechanism to identify targeted version of an index image
In some cases, IIB needs to know which OCP version an index image is being built for. For example, it may require a different `opm` command to add a bundle image to the index image.

Let's have a label, `com.redhat.index.delivery.version`, in the index image to facilitate this. The value of this label should be a single OCP version, e.g. `com.redhat.index.delivery.version=v4.5`

IIB must re-add this label to every index image it builds, add and remove operations.

If the label is not set, the default value should be `v4.5`.

shawn-hurley:
* default to 4.5 even if not found
* add the label to the opm generated dockerfile for index image build

fixes #114

### DIFF
--- a/iib/workers/tasks/build.py
+++ b/iib/workers/tasks/build.py
@@ -514,7 +514,8 @@ def _prepare_request_for_build(
         currently built for; if ``from_index`` is ``None``, then this is used as the list of arches
         to build the index image for
     :param list bundles: the list of bundles to create the bundle mapping on the request
-    :return: a dictionary with the keys: arches, binary_image_resolved, and from_index_resolved.
+    :return: a dictionary with the keys: arches, binary_image_resolved, from_index_resolved, and
+        ocp_version.
     :raises IIBError: if the container image resolution fails or the architectures couldn't be
         detected.
     """
@@ -535,9 +536,13 @@ def _prepare_request_for_build(
         with set_registry_token(overwrite_from_index_token, from_index):
             from_index_resolved = _get_resolved_image(from_index)
             from_index_arches = _get_image_arches(from_index_resolved)
+            ocp_version = (
+                get_image_label(from_index_resolved, 'com.redhat.index.delivery.version') or 'v4.5'
+            )
         arches = arches | from_index_arches
     else:
         from_index_resolved = None
+        ocp_version = 'v4.5'
 
     if not arches:
         raise IIBError('No arches were provided to build the index image')
@@ -574,6 +579,7 @@ def _prepare_request_for_build(
         'arches': arches,
         'binary_image_resolved': binary_image_resolved,
         'from_index_resolved': from_index_resolved,
+        'ocp_version': ocp_version,
     }
 
 
@@ -763,6 +769,10 @@ def handle_add_request(
             overwrite_from_index_token,
         )
 
+        _add_ocp_label_to_index(
+            prebuild_info['ocp_version'], temp_dir, 'index.Dockerfile',
+        )
+
         arches = prebuild_info['arches']
         for arch in sorted(arches):
             _build_image(temp_dir, 'index.Dockerfile', request_id, arch)
@@ -830,6 +840,10 @@ def handle_rm_request(
 
     with tempfile.TemporaryDirectory(prefix='iib-') as temp_dir:
         _opm_index_rm(temp_dir, operators, binary_image, from_index, overwrite_from_index_token)
+
+        _add_ocp_label_to_index(
+            prebuild_info['ocp_version'], temp_dir, 'index.Dockerfile',
+        )
 
         arches = prebuild_info['arches']
         for arch in sorted(arches):
@@ -1170,3 +1184,19 @@ def _adjust_csv_annotations(operator_csvs, package_name, organization):
             csv_annotations[annotation] = value
 
         operator_csv.dump()
+
+
+def _add_ocp_label_to_index(ocp_version, temp_dir, dockerfile_name):
+    """
+    Add the OCP delivery label to the provided dockerfile.
+
+    :param str ocp_version: the OCP version that the index is delivery content for.
+    :param str temp_dir: the temp directory to look for the dockerfile.
+    :param str dockerfile_name: the dockerfile name.
+    """
+    with open(os.path.join(temp_dir, dockerfile_name), 'a') as dockerfile:
+        label = f'LABEL com.redhat.index.delivery.version="{ocp_version}"'
+        dockerfile.write(f'\n{label}\n')
+        log.debug(
+            'Added the following line to %s: %s', dockerfile_name, label,
+        )
--- a/tests/test_workers/test_tasks/test_build.py
+++ b/tests/test_workers/test_tasks/test_build.py
@@ -407,6 +407,16 @@ def test_overwrite_from_index(
                 'some-bundle2': ['quay.io/some-bundle2:v1'],
             },
         ),
+        (
+            ['amd64'],
+            'some-index:latest',
+            set(),
+            ['quay.io/some-bundle:v1', 'quay.io/some-bundle2:v1'],
+            {
+                'some-bundle': ['quay.io/some-bundle:v1'],
+                'some-bundle2': ['quay.io/some-bundle2:v1'],
+            },
+        ),
     ),
 )
 @mock.patch('iib.workers.tasks.build.set_request_state')
@@ -430,6 +440,8 @@ def test_prepare_request_for_build(
     from_index_resolved = None
     expected_arches = set(add_arches) | from_index_arches
     expected_payload_keys = {'binary_image_resolved', 'state', 'state_reason'}
+    gil_side_effect = []
+    ocp_version = 'v4.5'
     if expected_bundle_mapping:
         expected_payload_keys.add('bundle_mapping')
     if from_index:
@@ -438,21 +450,29 @@ def test_prepare_request_for_build(
         mock_gri.side_effect = [binary_image_resolved, from_index_resolved]
         mock_gia.side_effect = [expected_arches, from_index_arches]
         expected_payload_keys.add('from_index_resolved')
+        gil_side_effect = ['v4.6']
+        ocp_version = 'v4.6'
     else:
         mock_gri.side_effect = [binary_image_resolved]
         mock_gia.side_effect = [expected_arches]
+        gil_side_effect = []
 
     if bundles:
-        mock_gil.side_effect = [bundle.rsplit('/', 1)[1].split(':', 1)[0] for bundle in bundles]
+        bundle_side_effects = [bundle.rsplit('/', 1)[1].split(':', 1)[0] for bundle in bundles]
+        gil_side_effect.extend(bundle_side_effects)
 
+    mock_gil.side_effect = gil_side_effect
     rv = build._prepare_request_for_build(
         'binary-image:latest', 1, from_index, None, add_arches, bundles
     )
+
     assert rv == {
         'arches': expected_arches,
         'binary_image_resolved': binary_image_resolved,
         'from_index_resolved': from_index_resolved,
+        'ocp_version': ocp_version,
     }
+
     mock_ur.assert_called_once()
     update_request_payload = mock_ur.call_args[0][1]
     if expected_bundle_mapping:
@@ -566,7 +586,9 @@ def test_skopeo_copy_fail_max_retries(mock_run_cmd):
 @mock.patch('iib.workers.tasks.build.validate_legacy_params_and_config')
 @mock.patch('iib.workers.tasks.build.gate_bundles')
 @mock.patch('iib.workers.tasks.build._get_resolved_bundles')
+@mock.patch('iib.workers.tasks.build._add_ocp_label_to_index')
 def test_handle_add_request(
+    mock_aolti,
     mock_grb,
     mock_gb,
     mock_vlpc,
@@ -589,6 +611,7 @@ def test_handle_add_request(
         'arches': arches,
         'binary_image_resolved': 'binary-image@sha256:abcdef',
         'from_index_resolved': 'from-index@sha256:bcdefg',
+        'ocp_version': 'v4.5',
     }
     mock_grb.return_value = ['some-bundle@sha']
     legacy_packages = {'some_package'}
@@ -618,6 +641,7 @@ def test_handle_add_request(
     mock_vl.assert_called_once()
     mock_prfb.assert_called_once()
     mock_gb.assert_called_once()
+    mock_aolti.assert_called_once()
     mock_glsp.assert_called_once_with(['some-bundle@sha'], 3, force_backport=force_backport)
 
     add_args = mock_oia.call_args[0]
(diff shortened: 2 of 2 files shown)


iib/workers/tasks/build.py
```diff
@@ -514,7 +514,8 @@ def _prepare_request_for_build(
         currently built for; if ``from_index`` is ``None``, then this is used as the list of arches
         to build the index image for
     :param list bundles: the list of bundles to create the bundle mapping on the request
-    :return: a dictionary with the keys: arches, binary_image_resolved, and from_index_resolved.
+    :return: a dictionary with the keys: arches, binary_image_resolved, and from_index_resolved,
+        ocp_version.
```
yashvardhannanavati, line 518: `...binary_image_resolved, and from_index_resolved, ocp_version` => `binary_image_resolved, from_index_resolved, and ocp_version`

iib/workers/tasks/build.py
```diff
@@ -535,9 +536,14 @@ def _prepare_request_for_build(
         with set_registry_token(overwrite_from_index_token, from_index):
             from_index_resolved = _get_resolved_image(from_index)
             from_index_arches = _get_image_arches(from_index_resolved)
+            ocp_version = get_image_label(from_index_resolved, "com.redhat.index.delivery.version")
```
yashvardhannanavati, line 539: Could you please use single quotes throughout? Sorry but we enforce that throughout our code base.

iib/workers/tasks/build.py
```diff
@@ -535,9 +536,14 @@ def _prepare_request_for_build(
...
         arches = arches | from_index_arches
     else:
         from_index_resolved = None
+        ocp_version = None
```
yashvardhannanavati, line 543: Why not default `ocp_version` to 'v4.5' here rather than adding another check later?

iib/workers/tasks/build.py
```diff
@@ -754,6 +761,8 @@ def handle_add_request(
         binary_image, request_id, from_index, overwrite_from_index_token, add_arches, bundles
     )
 
+    print(prebuild_info)
```
yashvardhannanavati, line 764: Could you please remove this `print`?

iib/workers/tasks/build.py
shawn-hurley: what happens if the index image does not have the label above? Still, need the check below this right?

yashvardhannanavati: You're right. Do you think we can make that line 
`ocp_version = get_image_label(from_index_resolved, 'com.redhat.index.delivery.version') or 'v4.5'`? and default here to 'v4.5' as well?

iib/workers/tasks/build.py
shawn-hurley: 🤦 sorry

iib/workers/tasks/build.py
shawn-hurley: +1 that makes sense!

iib/workers/tasks/build.py
```diff
@@ -1170,3 +1178,20 @@ def _adjust_csv_annotations(operator_csvs, package_name, organization):
...
+    :param str temp_dir: the temp directory to look for the dockerfile.
+    :param str dockerfile_name: the dockerfile name.
+    """
+    with open(os.path.join(temp_dir, dockerfile_name), 'a') as dockerfile:
+        dockerfile.write('\nLABEL com.redhat.index.delivery.version="{0}"'.format(ocp_version))
+        log.debug(
+            'Updated the dockerfile: %s with the following content:\n%s',
+            dockerfile_name,
+            dockerfile.read(),
+        )
```
mprahl, line 1197: If you're going to log the contents, you need to open the file for reading as well (`a+`). After writing to the file, you need go back to the beginning of the file to read it.
```suggestion
    with open(os.path.join(temp_dir, dockerfile_name), 'a+') as dockerfile:
        dockerfile.write('\nLABEL com.redhat.index.delivery.version="{0}"'.format(ocp_version))
        dockerfile.seek(0)
        log.debug(
            'Updated the dockerfile: %s with the following content:\n%s',
            dockerfile_name,
            dockerfile.read(),
        )
```

I think a better alternative is to just log the line you are adding to the Dockerfile.

tests/test_workers/test_tasks/test_build.py
```diff
@@ -438,6 +439,8 @@ def test_prepare_request_for_build(
         mock_gri.side_effect = [binary_image_resolved, from_index_resolved]
         mock_gia.side_effect = [expected_arches, from_index_arches]
         expected_payload_keys.add('from_index_resolved')
+        mock_gil.return_value = 'v4.6'
```
mprahl, line 442: This is problematic since `mock_gil.side_effect` is set below. You'll likely need to create a list of return values and then set `mock_gil.side_effect` to it before `_prepare_request_for_build` is called.

iib/workers/tasks/build.py
```diff
@@ -1170,3 +1178,20 @@ def _adjust_csv_annotations(operator_csvs, package_name, organization):
             csv_annotations[annotation] = value
 
         operator_csv.dump()
+
+
+def _add_ocp_label_to_index(ocp_version, temp_dir, dockerfile_name):
```
mprahl, line 1183: Could you please add a test for this?

iib/workers/tasks/build.py
```diff
@@ -1170,3 +1178,20 @@ def _adjust_csv_annotations(operator_csvs, package_name, organization):
...
+    """
+    Add ocp delivery label to the dockerfile provided.
+
+    :param str ocp_version: the ocp version that the index is delivery content for.
+    :param str temp_dir: the temp directory to look for the dockerfile.
+    :param str dockerfile_name: the dockerfile name.
+    """
+    with open(os.path.join(temp_dir, dockerfile_name), 'a') as dockerfile:
+        dockerfile.write('\nLABEL com.redhat.index.delivery.version="{0}"'.format(ocp_version))
```
mprahl, line 1192: Optional: It may be desirable to end the file with a new line.

iib/workers/tasks/build.py
```diff
@@ -763,6 +767,10 @@ def handle_add_request(
             overwrite_from_index_token,
         )
 
+        _add_ocp_label_to_index(
```
yashvardhannanavati, line 770: We might want to add this for `rm` request as well right? https://github.com/release-engineering/iib/pull/116/files#diff-055f6f8c453137f787488a4d047253a1R806

iib/workers/tasks/build.py
yashvardhannanavati, line 1183: If I remember correctly, the functions in this file are alphabetically ordered. Can I please request you to move this function if what I mentioned is true?

tests/test_workers/test_tasks/test_build.py
shawn-hurley: For my own edification can you explain why it is a problem and the tests are passing as I am not following why it would be a problem?

iib/workers/tasks/build.py
```diff
@@ -535,9 +536,11 @@ def _prepare_request_for_build(
         with set_registry_token(overwrite_from_index_token, from_index):
             from_index_resolved = _get_resolved_image(from_index)
             from_index_arches = _get_image_arches(from_index_resolved)
+            ocp_version = get_image_label(from_index_resolved, 'com.redhat.index.delivery.version') or 'v4.5'
```
mprahl, line 539: Do you think the default label value should be configurable? Could this ever change in the future as v4.5 goes EoL?

iib/workers/tasks/build.py
mprahl: @yashvardhannanavati it looks like it used to be but @lcarva added a couple functions that are not, so I won't hold @shawn-hurley to that as well but it'd still be nice to do so.

tests/test_workers/test_tasks/test_build.py
mprahl: @shawn-hurley this only works because the one test that sets `bundles` in the `pytest.parametrize` call does not set `from_index`. If a test is later added that sets `from_index` and `bundles`, it will fail because the `side_effect` only has enough values for the `get_image_label` calls for each bundle.

iib/workers/tasks/build.py
shawn-hurley: probably depends on if this is intended to be used by other folks, but once we set the label name, I don't think that it should change. 

Even as 4.5 goes EOL it would require us to convert 4.6 index image's to the new value. 

If we were to switch regardless of config value, you would need to write a migration step into the code I think. Is this required t be handled right now?

mprahl: It's not required. I was just checking with you to make sure.

iib/workers/tasks/build.py
shawn-hurley: added

tests/test_workers/test_tasks/test_build.py
shawn-hurley: I added the test that you described and I updated the test to handle it.

iib/workers/tasks/build.py
shawn-hurley: added

iib/workers/tasks/build.py
```diff
@@ -1170,3 +1184,21 @@ def _adjust_csv_annotations(operator_csvs, package_name, organization):
...
+    :param str dockerfile_name: the dockerfile name.
+    """
+    with open(os.path.join(temp_dir, dockerfile_name), 'a') as dockerfile:
+        label_str = '\nLABEL com.redhat.index.delivery.version="{0}"\n'.format(ocp_version)
+        dockerfile.write(label_str)
+        log.debug(
+            'Updated the dockerfile: %s with the following content:\n%s',
+            dockerfile_name,
+            label_str,
+        )
```
mprahl, line 1204:
```suggestion
        label = f'LABEL com.redhat.index.delivery.version="{ocp_version}"'
        dockerfile.write(f'\n{label}\n')
        log.debug(
            'Added the following line to the Dockerfile %s: %s',
            dockerfile_name,
            label,
        )
```

tests/test_workers/test_tasks/test_build.py
```diff
@@ -407,6 +407,16 @@ def test_overwrite_from_index(
                 'some-bundle2': ['quay.io/some-bundle2:v1'],
             },
         ),
+        (
+            ['amd64'],
+            "some-index:latest",
```
mprahl, line 412: Please use single quotes here

iib/workers/tasks/build.py
```diff
@@ -1170,3 +1184,21 @@ def _adjust_csv_annotations(operator_csvs, package_name, organization):
             csv_annotations[annotation] = value
 
         operator_csv.dump()
+
+
+def _add_ocp_label_to_index(ocp_version, temp_dir, dockerfile_name):
+    """
+    Add ocp delivery label to the dockerfile provided.
```
mprahl, line 1191:
```suggestion
    Add the OCP delivery label to the provided Dockerfile.
```

iib/workers/tasks/build.py
```diff
@@ -1170,3 +1184,21 @@ def _adjust_csv_annotations(operator_csvs, package_name, organization):
...
+
+    :param str ocp_version: the ocp version that the index is delivery content for.
```
mprahl, line 1193:
```suggestion
    :param str ocp_version: the OCP version that the index is delivery content for.
```

iib/workers/tasks/build.py
```diff
@@ -1170,3 +1184,21 @@ def _adjust_csv_annotations(operator_csvs, package_name, organization):
...
+    :param str temp_dir: the temp directory to look for the dockerfile.
```
mprahl, line 1194:
```suggestion
    :param str temp_dir: the temp directory to look for the Dockerfile.
```

tests/test_workers/test_tasks/test_build.py
```diff
@@ -438,21 +450,29 @@ def test_prepare_request_for_build(
...
+        ocp_version = 'v4.6'
     else:
         mock_gri.side_effect = [binary_image_resolved]
         mock_gia.side_effect = [expected_arches]
+        gil_side_effect = []
 
     if bundles:
-        mock_gil.side_effect = [bundle.rsplit('/', 1)[1].split(':', 1)[0] for bundle in bundles]
+        bundle_side_effects = [bundle.rsplit('/', 1)[1].split(':', 1)[0] for bundle in bundles]
+        gil_side_effect = [*gil_side_effect, *bundle_side_effects]
```
mprahl, line 462:
```suggestion
        gil_side_effect.extend(bundle_side_effects)
```

tests/test_workers/test_tasks/test_build.py
```diff
@@ -1267,3 +1306,19 @@ def test_adjust_csv_annotations_no_customizations(mock_yaml_dump, tmpdir):
...
+    )
+
+    build._add_ocp_label_to_index('v4.5', operator_dir, 'Dockerfile')
+
+    assert (
+        operator_dir.join('Dockerfile').read_text('utf-8')
+        == """https://access.redhat.com/containers/?tab=tags#/registry.access.redhat.com/ubi8-minimal
+LABEL com.redhat.index.delivery.version="v4.5"
+"""
+    )
```
mprahl, line 1324: This is easier to read in my opinion.
```suggestion
    expected = textwrap.dedent(
        '''\
        https://access.redhat.com/containers/?tab=tags#/registry.access.redhat.com/ubi8-minimal
        LABEL com.redhat.index.delivery.version="v4.5
        ''''
    )
    assert operator_dir.join('Dockerfile').read_text('utf-8') == expected
```

tests/test_workers/test_tasks/test_build.py
```diff
@@ -1267,3 +1306,19 @@ def test_adjust_csv_annotations_no_customizations(mock_yaml_dump, tmpdir):
     build._adjust_csv_annotations(operator_manifest.files, 'amqp-streams', 'mos-eisley')
 
     mock_yaml_dump.assert_not_called()
+
+
+def test_add_ocp_label_to_index(tmpdir):
+    operator_dir = tmpdir.mkdir('operator')
+    operator_dir.join('Dockerfile').write(
+        'https://access.redhat.com/containers/?tab=tags#/registry.access.redhat.com/ubi8-minimal'
```
mprahl, line 1314: I've never seen a Dockerfile contain a plain URL like this before... what does this do?

tests/test_workers/test_tasks/test_build.py
shawn-hurley: +1 did not know about textwrap thanks!

tests/test_workers/test_tasks/test_build.py
shawn-hurley: it is the beginning of a type of docker file, but it is not the full thing

tests/test_workers/test_tasks/test_build.py
```diff
@@ -1267,3 +1306,64 @@ def test_adjust_csv_annotations_no_customizations(mock_yaml_dump, tmpdir):
...
+    )
+    operator_dir.join('Dockerfile').write(dockerfile_txt)
+
+    build._add_ocp_label_to_index('v4.5', operator_dir, 'Dockerfile')
+
+    expected = dockerfile_txt + textwrap.dedent(
+        '''\
+        \nLABEL com.redhat.index.delivery.version="v4.5"
+        '''
+    )
```
mprahl, line 1368:
```suggestion
    expected = dockerfile_txt + '\nLABEL com.redhat.index.delivery.version="v4.5"\n'
```

PR #118, opened 2020-07-06
Update docker-compose opm version to v1.12.7

yashvardhannanavati:


yashvardhannanavati:
@midnightercz @lipoja @chandwanitulsi @querti Could you PTAL?

PR #120, opened 2020-07-09
Add support for Add and Rm Batch Requests

yashvardhannanavati:


yashvardhannanavati:
Working on adding the docs in a separate commit. It'll be a part of this PR. :)

iib/web/api_v1.py
```diff
@@ -460,3 +460,108 @@ def regenerate_bundle_batch():
...
+                # Check for the validity of a RM request
+                request = RequestRm.from_json(build_request, batch)
+            elif build_request.get('bundles'):
+                # Check for the validity of an Add request
+                request = RequestAdd.from_json(build_request, batch)
+            else:
+                raise ValidationError('Build request is not a valid Add/Rm request.')
+        except ValidationError as e:
+            # Rollback the transaction if any of the build requests are invalid
+            db.session.rollback()
```
rohanpm, line 495: I'm curious if this is actually needed given that you're just about to exit the request handler anyway?  Is there something in place to automatically rollback if a request handler raises (or if not rollback, just drop the session without committing)?

If it is needed to explicitly rollback whenever a request is about to fail: that seems awfully fragile.  It's not like ValidationError is the only kind of error which could possibly occur.  Why is it appropriate to roll back on that exception and not others?

Based on this text in https://flask-sqlalchemy.palletsprojects.com/en/2.x/quickstart/ I'd have thought that this code isn't useful:

> You have to commit the session, but you don’t have to remove it at the end of the request, Flask-SQLAlchemy does that for you.

iib/web/api_v1.py
```diff
@@ -460,3 +460,108 @@ def regenerate_bundle_batch():
...
+            raise ValidationError(
+                f'{str(e).rstrip(".")}. This occurred on the build request in '
+                f'index {payload["build_requests"].index(build_request)}.'
+            )
+        db.session.add(request)
+        requests.append(request)
+
+    db.session.commit()
+    messaging.send_messages_for_new_batch_of_requests(requests)
```
rohanpm, line 504: What happens if we're interrupted after the DB commit and before the messages are sent (or if message sending fails) - does the system ever detect and recover from the "lost" requests?

iib/web/api_v1.py
yashvardhannanavati: Nice catch! I believe I can remove that piece of code.

/cc @mprahl @lcarva

iib/web/api_v1.py
yashvardhannanavati: So if an error is encountered while sending messages, it is considered as a non-fatal error and the request continues its course. The delivery of messages is not guaranteed by IIB.

yashvardhannanavati:
@midnightercz @chandwanitulsi @lipoja @querti @JayZ12138 Could you PTAL?

iib/web/api_v1.py
mprahl: So Flask-SQLAlchemy will call `session.remove` when the Flask application context is torn down:
https://github.com/pallets/flask-sqlalchemy/blob/7352d8b941b397b58341667c455c989baba16cd1/src/flask_sqlalchemy/__init__.py#L852

This ends up calling `session.close` which rolls back the transaction as described here:
https://docs.sqlalchemy.org/en/13/orm/contextual.html#using-thread-local-scope-with-web-applications

So yes, the line was unnecessary but it was also harmless.

iib/web/static/api_v1.yaml
```diff
@@ -416,6 +416,62 @@ paths:
...
+              properties:
+                annotations:
+                  type: object
+                  description: >
+                    An arbitrary JSON object associated with the batch. This is accessed by the
+                    "batch_annotations" key's value on a build request in the batch.
+                build_requests:
+                  type: array
+                  description: >
+                    Request objects of type Add or Rm
```
mprahl, line 439:
```suggestion
                  description: Request objects of type Add or Rm
```

iib/web/static/api_v1.yaml
```diff
@@ -416,6 +416,62 @@ paths:
...
+        '400':
+          description: The input is invalid
+          content:
+            application/json:
+              schema:
+                type: object
+                properties:
+                  error:
+                    type: string
+                    example: '"organization" must be a string'
```
mprahl, line 464: This message should indicate the index the failure occurred in.

tests/test_web/test_api_v1.py
```diff
@@ -1316,3 +1316,146 @@ def test_regenerate_bundle_batch_invalid_input(payload, error_msg, app, auth_env
...
+    annotations = {'msdhoni': 'The best captain ever!'}
+    data = {
+        'annotations': annotations,
+        'build_requests': [
+            {
+                'bundles': ['registry-proxy/rh-osbs/lgallett-bundle:v1.0-9'],
+                'binary_image': 'registry-proxy/rh-osbs/openshift-ose-operator-registry:v4.5',
+                'from_index': 'registry-proxy/rh-osbs-stage/iib:v4.5',
+                'add_arches': ['amd64'],
+                'cnr_token': 'no_tom_brady_anymore',
```
mprahl, line 1334: :cry:

tests/test_web/test_api_v1.py
```diff
@@ -1316,3 +1316,146 @@ def test_regenerate_bundle_batch_invalid_input(payload, error_msg, app, auth_env
 
     assert rv.status_code == 400, rv.json
     assert rv.json == {'error': error_msg}
+
+
+@mock.patch('iib.web.api_v1.handle_add_request')
+@mock.patch('iib.web.api_v1.handle_rm_request')
+@mock.patch('iib.web.api_v1.messaging.send_messages_for_new_batch_of_requests')
+def test_add_rm_batch_success(mock_smfnbor, mock_hrr, mock_har, app, auth_env, client, db):
```
mprahl, line 1324: Optional: It'd also be good to ensure the requests were created in the database

iib/web/api_v1.py
```diff
@@ -460,3 +460,103 @@ def regenerate_bundle_batch():
...
+    request_jsons = []
+    # This list will be used for the log message below and avoids the need of having to iterate
+    # through the list of requests another time
+    request_id_strs = []
+    for build_request, request in zip(payload['build_requests'], requests):
+        request_jsons.append(request.to_json())
+        request_id_strs.append(str(request.id))
+
+        if isinstance(request, RequestAdd):
+            celery_queue = _get_user_queue()
```
lcarva, line 510: We could this once before entering the loop and use the value when calling `handle_add_request.apply_async` as well as `handle_rm_request.apply_async`.

tests/test_web/test_api_v1.py
yashvardhannanavati: haha, I was hoping you would notice it :p

iib/web/api_v1.py
```diff
@@ -460,3 +460,103 @@ def regenerate_bundle_batch():
...
+            args = [
```
lcarva, line 511: Having to maintain this list in two places is going to bite us. As a future enhancement, we should introduce helper functions that can construct `args` and `safe_args` in a single place for each request type.

iib/web/models.py
```diff
@@ -728,11 +728,6 @@ def _from_json(
         if current_user.is_authenticated:
             request_kwargs['user'] = current_user
 
-        # Add the request to a new batch
-        batch = Batch()
```
lcarva, line 732: I don't understand why this was moved.

yashvardhannanavati: The choice was between adding another arg to the `cls._from_json` function of the parent class right from the API call (when annotations are present) and creating a batch in the from_json methods of the respective child classes. I chose the latter because it now also becomes consistent with the from_json method of the RegenerateBundle class. 

If you feel strongly against it, I would be happy to move it back to where it was and introduce an extra arg.

querti:
LGTM, As far as I can tell.

iib/web/models.py
lcarva: Hang on. All you have to do is add a new optional parameter, `batch`, to `RequestIndexImageMixin._from_json`, `RequestAdd.from_json`, and `RequestRm.from_json`. The last two already call the first. It's just then a matter of forwarding whatever `batch` parameter was given.

I don't have a strong opinion on this. I just don't think the duplication is necessary though.

iib/web/api_v1.py
yashvardhannanavati: +1. Filed CLOUDDST-1705 for the same

iib/web/models.py
yashvardhannanavati: You got to me. I have made the change. :)

tests/test_web/test_api_v1.py
```diff
@@ -1316,3 +1316,149 @@ def test_regenerate_bundle_batch_invalid_input(payload, error_msg, app, auth_env
...
+                    ", 2, 'registry:8443/iib-build:11', None, None, None]"
+                ),
+                link_error=mock.ANY,
+                queue=None,
+            ),
+        )
+    )
+
+    assert db.session.query(RequestAdd).filter_by(id=1).scalar()
+    assert db.session.query(RequestRm).filter_by(id=2).scalar()
```
mprahl, line 1401: @yashvardhannanavati not a big deal, but the proper way to do this is as follows:
```python
assert db.session.query(RequestAdd).get(1)
```

PR #122, opened 2020-07-15
Restrict backporting just to v4.5

lcarva:
* CLOUDBLD-1880

### DIFF
--- a/iib/workers/tasks/build.py
+++ b/iib/workers/tasks/build.py
@@ -563,24 +563,45 @@ def _prepare_request_for_build(
         if operator:
             bundle_mapping.setdefault(operator, []).append(bundle)
 
-    payload = {
-        'binary_image_resolved': binary_image_resolved,
-        'state': 'in_progress',
-        'state_reason': f'Building the index image for the following arches: {arches_str}',
-    }
-    if bundle_mapping:
-        payload['bundle_mapping'] = bundle_mapping
-    if from_index_resolved:
-        payload['from_index_resolved'] = from_index_resolved
-    exc_msg = 'Failed setting the resolved images on the request'
-    update_request(request_id, payload, exc_msg)
-
     return {
         'arches': arches,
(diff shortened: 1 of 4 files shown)


iib/workers/tasks/build.py
```diff
@@ -747,19 +747,19 @@ def handle_add_request(
     if greenwave_config:
         gate_bundles(resolved_bundles, greenwave_config)
 
+    prebuild_info = _prepare_request_for_build(
```
mprahl, line 750: Instead of moving this function call to earlier, could you just move the logic in `_prepare_request_for_build` of determining the OCP version to a separate helper function that you call here?

I like the current flow where all validation occurs before the "Building the index image for the following arches: ..." state reason gets set.

lcarva: Hmm. It's quite odd that `_prepare_request_for_build` does set this state. But it's even worse if there's an intermediate state before the image actually gets built. I'll play with restructuring the code.

lcarva: Done. I pulled the state update out of the function.

iib/workers/tasks/build.py
```diff
@@ -563,25 +563,34 @@ def _prepare_request_for_build(
...
+    return {
+        'arches': arches,
         'binary_image_resolved': binary_image_resolved,
+        'bundle_mapping': bundle_mapping,
+        'from_index_resolved': from_index_resolved,
+        'ocp_version': ocp_version,
+    }
+
+
+def _update_index_image_build_state(request_id, prebuild_info):
```
lcarva, line 575: I need docstring here. Coming up.

lcarva: Done.

iib/workers/tasks/build.py
mprahl: That's a clean approach. Thanks!

PR #130, opened 2020-07-27
Do not fail if overwrite_from_index_token is given and from_index is not

yashvardhannanavati:
Refers to CLOUDDST-1902


yashvardhannanavati:
@midnightercz @chandwanitulsi @lipoja @querti @JayZ12138 Could you PTAL?

PR #137, opened 2020-08-05
Inspect index image to check if bundles are present

querti:
I'd like to make sure that I'm using the proper approach before proceeding with the implementation.

Also, I'd like to ask how to match the bundle from the index image data. I thought that hash would be a good approach, but it still gives unique constraint error even if bundle image with this hash is not in the index image

### DIFF
--- a/README.md
+++ b/README.md
@@ -220,6 +220,14 @@ The custom configuration options for the Celery workers are listed below:
   file though. This defaults to `~/.docker/config.json.template`.
 * `iib_greenwave_url` - the URL to the Greenwave REST API if gating is desired
   (e.g. `https://greenwave.domain.local/api/v1.0/`). This defaults to `None`.
+* `iib_grpc_init_wait_time` - time to wait for the index image service to be initialized. This
+  defaults to `3` seconds.
+* `iib_grpc_max_port_tries` - maximum ports to try when initializing the index image service.
+  This defaults to `100` tries.
+* `iib_grpc_start_port` - first port to try when starting the service (subsequent are increments).
+  This defaults to `50051`.
+* `iib_grpc_max_tries` - maximum number of times to try to start the index image service
+  before giving up. This defaults to `5` attempts.
 * `iib_index_image_output_registry` - if set, that value will replace the value from `iib_registry`
   in the output `index_image` pull specification. This is useful if you'd like users of IIB to
   pull from a proxy to a registry instead of the registry directly.
--- a/docker/Dockerfile-workers
+++ b/docker/Dockerfile-workers
@@ -38,6 +38,8 @@ RUN dnf -y install \
 
 COPY --from=builder /usr/bin/opm /usr/bin/opm
 ADD https://github.com/estesp/manifest-tool/releases/download/v1.0.0/manifest-tool-linux-amd64 /usr/bin/manifest-tool
+ADD https://github.com/fullstorydev/grpcurl/releases/download/v1.7.0/grpcurl_1.7.0_linux_x86_64.tar.gz /src/grpcurl_1.7.0_linux_x86_64.tar.gz
+RUN cd /usr/bin && tar -xf /src/grpcurl_1.7.0_linux_x86_64.tar.gz grpcurl && rm -f /src/grpcurl_1.7.0_linux_x86_64.tar.gz
 RUN chmod +x /usr/bin/manifest-tool
 COPY . .
 RUN pip3 install -r requirements.txt --no-deps --require-hashes
--- a/iib/workers/config.py
+++ b/iib/workers/config.py
@@ -16,6 +16,10 @@ class Config(object):
         os.path.expanduser('~'), '.docker', 'config.json.template'
     )
     iib_greenwave_url = None
+    iib_grpc_init_wait_time = 3
+    iib_grpc_max_port_tries = 100
+    iib_grpc_max_tries = 5
+    iib_grpc_start_port = 50051
     iib_image_push_template = '{registry}/iib-build:{request_id}'
     iib_index_image_output_registry = None
     iib_log_level = 'INFO'
--- a/iib/workers/tasks/build.py
+++ b/iib/workers/tasks/build.py
@@ -1,14 +1,18 @@
 # SPDX-License-Identifier: GPL-3.0-or-later
 import hashlib
+import json
 import logging
 import os
+import re
+import subprocess
+import time
 import tempfile
 import textwrap
 
 from operator_manifest.operator import ImageName, OperatorManifest
 import ruamel.yaml
 
-from iib.exceptions import IIBError
+from iib.exceptions import IIBError, AddressAlreadyInUse
 from iib.workers.api_utils import set_request_state, update_request
 from iib.workers.config import get_worker_config
 from iib.workers.tasks.celery import app
@@ -335,6 +339,153 @@ def _get_resolved_image(pull_spec):
     pull_spec_resolved = f'{name}@sha256:{digest}'
     log.debug('%s resolved to %s', pull_spec, pull_spec_resolved)
     return pull_spec_resolved
+
+
+def _get_index_database(from_index, base_dir):
+    """
+    Get database file from the specified index image and save it locally.
+
+    :param str from_index: index image to get database file from.
+    :param str base_dir: base directory to which the database file should be saved.
+    :return: path to the copied database file.
+    :rtype: str
+    :raises IIBError: if any podman command fails.
+    """
+    data = skopeo_inspect(f'docker://{from_index}')
+    try:
+        db_path = data['Labels']['operators.operatorframework.io.index.database.v1']
+    except KeyError:
+        raise IIBError('Index image doesn\'t have the label specifying its database location.')
+    _copy_files_from_image(from_index, db_path, base_dir)
+    local_path = os.path.join(base_dir, os.path.basename(db_path))
+    return local_path
+
+
+def _serve_index_registry(db_path):
+    """
+    Locally start OPM registry service, which can be communicated with using gRPC queries.
+
+    Due to IIB's paralellism, the service can run multiple times, which could lead to port
+    binding conflicts. Resolution of port conflicts is handled in this function as well.
+
+    :param str db_path: path to index database containing the registry data.
+    :return: tuple containing port number of the running service and the running Popen object.
+    :rtype: (int, Popen)
+    :raises IIBError: if all tried ports are in use, or the command failed for another reason.
+    """
+    conf = get_worker_config()
+    port_start = conf['iib_grpc_start_port']
+    port_end = port_start + conf['iib_grpc_max_port_tries']
+
+    for port in range(port_start, port_end):
+        try:
+            return (
+                port,
+                _serve_index_registry_at_port(
+                    db_path, port, conf['iib_grpc_max_tries'], conf['iib_grpc_init_wait_time']
+                ),
+            )
+        except AddressAlreadyInUse:
+            log.info('Port %d is in use, trying another.', port)
+
+    err_msg = f'No free port has been found after {conf.get("iib_grpc_max_port_tries")} attempts.'
+    log.error(err_msg)
+    raise IIBError(err_msg)
+
+
+@retry(attempts=2, wait_on=IIBError, logger=log)
+def _serve_index_registry_at_port(db_path, port, max_tries, wait_time):
+    """
+    Start an image registry service at a specified port.
+
+    :param str db_path: path to index database containing the registry data.
+    :param str int port: port to start the service on.
+    :param max_tries: how many times to try to start the service before giving up.
+    :param wait_time: time to wait before checking if the service is initialized.
+    :return: object of the running Popen process.
+    :rtype: Popen
+    :raises IIBError: if the process has failed to initialize too many times, or an unexpected
+        error occured.
+    :raises AddressAlreadyInUse: if the specified port is already being used by another service.
+    """
+    cmd = ['opm', 'registry', 'serve', '-p', str(port), '-d', db_path]
+    for attempt in range(max_tries):
+        rpc_proc = subprocess.Popen(
+            cmd, stdout=subprocess.PIPE, stderr=subprocess.PIPE, universal_newlines=True
+        )
+        start_time = time.time()
+        while time.time() - start_time < wait_time:
+            time.sleep(1)
+            ret = rpc_proc.poll()
+            # process has terminated
+            if ret is not None:
+                stderr = rpc_proc.stderr.read()
+                if 'address already in use' in stderr:
+                    raise AddressAlreadyInUse(f'Port {port} is already used by a different service')
+                raise IIBError(f'Command "{" ".join(cmd)}" has failed with error "{stderr}"')
+
+            # query the service to see if it has started
+            try:
+                output = run_cmd(
+                    ['grpcurl', '-plaintext', f'localhost:{port}', 'list', 'api.Registry']
+                )
+            except IIBError:
+                output = ''
+
+            if 'api.Registry.ListBundles' in output:
+                log.debug('Started the command "%s"', ' '.join(cmd))
+                log.info('Index registry service has been initialized.')
+                return rpc_proc
+
+        rpc_proc.kill()
+
+    raise IIBError(f'Index registry has not been initialized after {max_tries} tries')
+
+
+def _get_present_bundles(from_index, base_dir):
+    """
+    Get a list of bundles already present in the index image.
+
+    :param str from_index: index image to inspect.
+    :param str base_dir: base directory to create temporary files in.
+    :return: list of present bundles as provided by the grpc query.
+    :rtype: list
+    :raises IIBError: if any of the commands fail.
+    """
+    db_path = _get_index_database(from_index, base_dir)
+    port, rpc_proc = _serve_index_registry(db_path)
+
+    bundles = run_cmd(
+        ['grpcurl', '-plaintext', f'localhost:{port}', 'api.Registry/ListBundles'],
+        exc_msg='Failed to get bundle data from index image',
+    )
+    rpc_proc.kill()
+
+    # Transform returned data to parsable json
+    present_bundles = [json.loads(bundle) for bundle in re.split(r'(?<=})\n(?={)', bundles)]
+    return present_bundles
+
+
+def _get_missing_bundles(present_bundles, bundles):
+    """
+    Filter out bundles to only those not present in the index image.
+
+    :param list present_bundles: list of bundles present in the index image, as provided by opm.
+    :param list bundles: resolved bundles requested to be added to the index image.
+    :return: list of bundles not present in the index image.
+    :rtype: list
+    """
+    present_bundle_hashes = []
+    filtered_bundles = []
+    for bundle in present_bundles:
+        if '@sha256:' in bundle['bundlePath']:
+            present_bundle_hashes.append(bundle['bundlePath'].split('@sha256:')[-1])
+
+    for bundle in bundles:
+        if bundle.split('@sha256:')[-1] not in present_bundle_hashes:
+            filtered_bundles.append(bundle)
+
+    return filtered_bundles
 
 
 @retry(attempts=2, wait_on=IIBError, logger=log)
@@ -785,6 +936,30 @@ def handle_add_request(
     _update_index_image_build_state(request_id, prebuild_info)
 
     with tempfile.TemporaryDirectory(prefix='iib-') as temp_dir:
+        if from_index:
+            msg = 'Checking if bundles are already present in index image'
+            log.info(msg)
+            set_request_state(request_id, 'in_progress', msg)
+            present_bundles = _get_present_bundles(from_index, temp_dir)
+            filtered_bundles = _get_missing_bundles(present_bundles, resolved_bundles)
+            excluded_bundles = [
+                bundle for bundle in resolved_bundles if bundle not in filtered_bundles
+            ]
+            resolved_bundles = filtered_bundles
+            if not resolved_bundles:
+                log.info(
+                    'All bundles are already present in the index image.'
+                    ' No additional operations are necessary'
+                )
+                set_request_state(
+                    request_id, 'complete', 'All bundles are already present in the index image'
+                )
+                return
+            log.info(
+                'Following bundles are already present in the index image: %s',
+                ' '.join(excluded_bundles),
+            )
+
         _opm_index_add(
             temp_dir,
             resolved_bundles,
--- a/tests/test_workers/test_tasks/test_build.py
+++ b/tests/test_workers/test_tasks/test_build.py
@@ -613,7 +613,11 @@ def test_skopeo_copy_fail_max_retries(mock_run_cmd):
 @mock.patch('iib.workers.tasks.build.gate_bundles')
 @mock.patch('iib.workers.tasks.build._get_resolved_bundles')
 @mock.patch('iib.workers.tasks.build._add_ocp_label_to_index')
+@mock.patch('iib.workers.tasks.build._get_present_bundles')
+@mock.patch('iib.workers.tasks.build._get_missing_bundles')
 def test_handle_add_request(
+    mock_gmb,
+    mock_gpb,
     mock_aolti,
     mock_grb,
     mock_gb,
@@ -671,8 +675,8 @@ def test_handle_add_request(
     mock_aolti.assert_called_once()
     mock_glsp.assert_called_once_with(['some-bundle@sha'], 3, 'v4.5', force_backport=force_backport)
 
-    add_args = mock_oia.call_args[0]
-    assert ['some-bundle@sha'] in add_args
+    filter_args = mock_gmb.call_args[0]
+    assert ['some-bundle@sha'] in filter_args
     mock_oia.assert_called_once()
 
     assert mock_bi.call_count == len(arches)
@@ -687,7 +691,7 @@ def test_handle_add_request(
     mock_uiips.assert_called_once()
     mock_vii.assert_called_once()
     mock_capml.assert_called_once()
-    assert mock_srs.call_count == 3
+    assert mock_srs.call_count == 4
 
 
 @mock.patch('iib.workers.tasks.build._cleanup')
@@ -768,7 +772,11 @@ def test_handle_add_request_bundle_resolution_failure(mock_grb, mock_srs, mock_c
 @mock.patch('iib.workers.tasks.build.gate_bundles')
 @mock.patch('iib.workers.tasks.build._get_resolved_bundles')
 @mock.patch('iib.workers.tasks.build._add_ocp_label_to_index')
+@mock.patch('iib.workers.tasks.build._get_present_bundles')
+@mock.patch('iib.workers.tasks.build._get_missing_bundles')
 def test_handle_add_request_backport_failure_no_overwrite(
+    mock_gmb,
+    mock_gpb,
     mock_aolti,
     mock_grb,
     mock_gb,
@@ -1373,3 +1381,171 @@ def test_add_ocp_label_to_index(tmpdir):
 
     expected = dockerfile_txt + '\nLABEL com.redhat.index.delivery.version="v4.5"\n'
     assert operator_dir.join('Dockerfile').read_text('utf-8') == expected
+
+
+def test_get_missing_bundles_no_match():
+    assert build._get_missing_bundles(
+        [
+            {
+                'packageName': 'bundle1',
+                'version': 'v1.0',
+                'bundlePath': 'quay.io/pkg/pkg1@sha256:987654',
+            },
+            {
+                'packageName': 'bundle2',
+                'version': 'v2.0',
+                'bundlePath': 'quay.io/pkg/pkg2@sha256:111111',
+            },
+        ],
+        ['quay.io/ns/repo@sha256:123456'],
+    ) == ['quay.io/ns/repo@sha256:123456']
+
+
+def test_get_missing_bundles_match_hash():
+    assert (
+        build._get_missing_bundles(
+            [
+                {
+                    'packageName': 'bundle1',
+                    'version': 'v1.0',
+                    'bundlePath': 'quay.io/pkg/pkg1@sha256:987654',
+                },
+                {
+                    'packageName': 'bundle2',
+                    'version': 'v2.0',
+                    'bundlePath': 'quay.io/pkg/pkg2@sha256:111111',
+                },
+            ],
+            ['quay.io/pkg/pkg1@sha256:987654'],
+        )
+        == []
+    )
+
+
+@mock.patch('time.sleep')
+@mock.patch('subprocess.Popen')
+@mock.patch('iib.workers.tasks.build.run_cmd')
+@mock.patch('iib.workers.tasks.build._copy_files_from_image')
+@mock.patch('iib.workers.tasks.build.skopeo_inspect')
+def test_get_present_bundles(mock_si, mock_copy, mock_run_cmd, mock_popen, mock_sleep, tmpdir):
+    with open(tmpdir.join('cidfile.txt'), 'w+') as f:
+        f.write('container_id')
+    mock_si.return_value = {
+        'Labels': {'operators.operatorframework.io.index.database.v1': 'some-path'}
+    }
+    mock_run_cmd.side_effect = [
+        'api.Registry.ListBundles',
+        '{"packageName": "package1", "version": "v1.0"\n}'
+        '\n{\n"packageName": "package2", "version": "v2.0"}',
+    ]
+    my_mock = mock.MagicMock()
+    mock_popen.return_value = my_mock
+    my_mock.stderr.read.return_value = 'address already in use'
+    my_mock.poll.side_effect = [1, None]
+    assert build._get_present_bundles('quay.io/index-image:4.5', str(tmpdir)) == [
+        {'packageName': 'package1', 'version': 'v1.0'},
+        {'packageName': 'package2', 'version': 'v2.0'},
+    ]
+    assert mock_run_cmd.call_count == 2
+
+
+@mock.patch('time.time')
+@mock.patch('os.remove')
+@mock.patch('time.sleep')
+@mock.patch('subprocess.Popen')
+@mock.patch('iib.workers.tasks.build.run_cmd')
+@mock.patch('iib.workers.tasks.build._copy_files_from_image')
+@mock.patch('iib.workers.tasks.build.skopeo_inspect')
+def test_get_present_bundles_grpc_not_initialize(
+    mock_si, mock_copy, mock_run_cmd, mock_popen, mock_sleep, mock_remove, mock_time, tmpdir,
+):
+    with open(tmpdir.join('cidfile.txt'), 'w+') as f:
+        f.write('container_id')
+    mock_run_cmd.side_effect = ['', '', '', '', ''] * 4
+    mock_time.side_effect = list(range(1, 80))
+    mock_si.return_value = {
+        'Labels': {'operators.operatorframework.io.index.database.v1': 'some-path'}
+    }
+    my_mock = mock.MagicMock()
+    mock_popen.return_value = my_mock
+    my_mock.poll.return_value = None
+    with pytest.raises(IIBError, match='Index registry has not been initialized after 5 tries'):
+        build._get_present_bundles('quay.io/index-image:4.5', str(tmpdir))
+    assert mock_run_cmd.call_count == 20
+
+
+@mock.patch('time.time')
+@mock.patch('os.remove')
+@mock.patch('time.sleep')
+@mock.patch('subprocess.Popen')
+@mock.patch('iib.workers.tasks.build.run_cmd')
+@mock.patch('iib.workers.tasks.build._copy_files_from_image')
+@mock.patch('iib.workers.tasks.build.skopeo_inspect')
+def test_get_present_bundles_grpc_delayed_initialize(
+    mock_si, mock_copy, mock_run_cmd, mock_popen, mock_sleep, mock_remove, mock_time, tmpdir,
+):
+    with open(tmpdir.join('cidfile.txt'), 'w+') as f:
+        f.write('container_id')
+    mock_time.side_effect = [i * 0.5 for i in range(1, 80)]
+    mock_si.return_value = {
+        'Labels': {'operators.operatorframework.io.index.database.v1': 'some-path'}
+    }
+    mock_run_cmd.side_effect = [
+        '',
+        '',
+        '',
+        '',
+        '',
+        '',
+        'api.Registry.ListBundles',
+        '{"packageName": "package1", "version": "v1.0"\n}'
+        '\n{\n"packageName": "package2", "version": "v2.0"}',
+    ]
+    my_mock = mock.MagicMock()
+    mock_popen.return_value = my_mock
+    my_mock.poll.return_value = None
+    assert build._get_present_bundles('quay.io/index-image:4.5', str(tmpdir)) == [
+        {'packageName': 'package1', 'version': 'v1.0'},
+        {'packageName': 'package2', 'version': 'v2.0'},
+    ]
+    assert mock_run_cmd.call_count == 8
+
+
+@mock.patch('time.sleep')
+@mock.patch('subprocess.Popen')
+@mock.patch('iib.workers.tasks.build.run_cmd')
+def test_serve_image_registry(mock_run_cmd, mock_popen, mock_sleep, tmpdir):
+    my_mock = mock.MagicMock()
+    mock_popen.return_value = my_mock
+    my_mock.stderr.read.side_effect = [
+        'address already in use',
+        'address already in use',
+    ]
+    mock_run_cmd.return_value = 'api.Registry.ListBundles'
+    my_mock.poll.side_effect = [1, 1, None]
+    port, _ = build._serve_index_registry('some_path.db')
+    assert port == 50053
+    assert my_mock.poll.call_count == 3
+
+
+@mock.patch('iib.workers.tasks.build.get_worker_config')
+@mock.patch('time.sleep')
+@mock.patch('subprocess.Popen')
+def test_serve_image_registry_no_ports(mock_popen, mock_sleep, mock_config, tmpdir):
+    my_mock = mock.MagicMock()
+    mock_popen.return_value = my_mock
+    my_mock.stderr.read.side_effect = [
+        'address already in use',
+        'address already in use',
+        'address already in use',
+    ]
+    my_mock.poll.side_effect = [1, 1, 1, None]
+    mock_config.return_value = {
+        'iib_grpc_start_port': 50051,
+        'iib_grpc_init_wait_time': 1,
+        'iib_grpc_max_port_tries': 3,
+        'iib_grpc_max_tries': 3,
+    }
+    with pytest.raises(IIBError, match='No free port has been found after 3 attempts.'):
+        build._serve_index_registry('some_path.db')
+    assert my_mock.poll.call_count == 3
--- a/iib/exceptions.py
+++ b/iib/exceptions.py
@@ -15,3 +15,7 @@ class IIBError(BaseException):
 
 class ValidationError(BaseException):
     """Denote invalid input."""
+
+
+class AddressAlreadyInUse(BaseException):
+    """Adress is already used by other service."""


querti:
@yashvardhannanavati @mprahl @lcarva @shawn-hurley pls check if this is the correct approach. Thanks.

shawn-hurley:
Can we not use the error codes returned by OPM?

querti:
> Can we not use the error codes returned by OPM?

@shawn-hurley This implementation is a result of a discussion in the exd-guild-hello-operator channel. The usage of error returned by OPM was my original solution, but the generic error message makes it tricky. The error is 'UNIQUE constraint failed: operatorbundle.name', and it doesn't give the names of the conflicting bundles, so the only solution was to re-run OPM for each bundle separately, which could be very slow.

Has the error message been improved? I think Yash asked this in the guild channel, but it hasn't been answered.

iib/workers/tasks/build.py
```diff
@@ -337,6 +341,35 @@ def _get_resolved_image(pull_spec):
...
+    data = json.loads(run_cmd(["podman", "image", "inspect", from_index], exc_msg='Failed to fetch image metadata'))
+    db_path = data[0]["Config"]["Labels"]["operators.operatorframework.io.index.database.v1"]
+
+    run_cmd(["podman", "cp", "{0}:{1}".format(cid, db_path), base_dir], exc_msg='Failed to copy db file from container')
+    run_cmd(["podman", "rm", cid], exc_msg='Failed to delete container')
+    local_path = os.path.join(base_dir, db_path.split("/")[-1])
+    rpc_proc = subprocess.Popen(["opm","registry", "serve", "-p", '50051', "-d", local_path])
+
+    # sleeping as intialization might take a while
+    time.sleep(5)
```
mprahl, line 363: Could you continuously check if the port is open rather than sleep an arbitrary amount?

iib/workers/tasks/build.py
```diff
@@ -337,6 +341,35 @@ def _get_resolved_image(pull_spec):
...
+        cid = f.read()
+    os.remove(cidfile)
+
+    data = json.loads(run_cmd(["podman", "image", "inspect", from_index], exc_msg='Failed to fetch image metadata'))
+    db_path = data[0]["Config"]["Labels"]["operators.operatorframework.io.index.database.v1"]
+
+    run_cmd(["podman", "cp", "{0}:{1}".format(cid, db_path), base_dir], exc_msg='Failed to copy db file from container')
+    run_cmd(["podman", "rm", cid], exc_msg='Failed to delete container')
+    local_path = os.path.join(base_dir, db_path.split("/")[-1])
+    rpc_proc = subprocess.Popen(["opm","registry", "serve", "-p", '50051', "-d", local_path])
```
mprahl, line 360: How do you know port 50051 is open? You need to pick an open port since multiple workers on the same VM could be using the port or even some other process unrelated to IIB.

iib/workers/tasks/build.py
```diff
@@ -337,6 +341,35 @@ def _get_resolved_image(pull_spec):
...
+
+    # sleeping as intialization might take a while
+    time.sleep(5)
+    bundles = run_cmd(["grpcurl", "-plaintext", "localhost:{0}".format('50051'), "api.Registry/ListBundles"],
+        exc_msg='Failed to get bundle data from index image')
+    rpc_proc.kill()
```
mprahl, line 366: Does this need to be in a context manager in the event `run_cmd` fails? Does the process automatically get killed when `rpc_proc` gets garbage collected?

iib/workers/tasks/build.py
```diff
@@ -337,6 +341,35 @@ def _get_resolved_image(pull_spec):
     return pull_spec_resolved
 
 
+def _serve_index_db(from_index, base_dir):
```
yashvardhannanavati, line 344: We should probably add a retry to this.

iib/workers/tasks/build.py
```diff
@@ -358,6 +391,22 @@ def _opm_index_add(
         ``overwrite_from_index``. The format of the token must be in the format "user:password".
     :raises IIBError: if the ``opm index add`` command fails.
     """
+    index_image_bundles = _serve_index_db(from_index, base_dir)
```
yashvardhannanavati, line 394: This should probably be a separate helper method.

iib/workers/tasks/build.py
```diff
@@ -358,6 +391,22 @@ def _opm_index_add(
...
+    present_bundle_hashes = [x["bundlePath"].split('@sha256:')[-1] for x in index_image_bundles]
+    bundle_hashes = [(x, x.split('@sha256:')[-1]) for x in bundles]
+    for b in present_bundle_hashes:
```
yashvardhannanavati, line 397: Could you please use `bundle_hash` instead of just b? I believe it would be more intuitive

iib/workers/tasks/build.py
```diff
@@ -358,6 +391,22 @@ def _opm_index_add(
...
+        log.info('present hash: {0}'.format(b))
+    for b in bundle_hashes:
```
yashvardhannanavati, line 399: Same as above.

iib/workers/tasks/build.py
```diff
@@ -358,6 +391,22 @@ def _opm_index_add(
...
+        log.info('bundle hash: {0}'.format(b[1]))
+    filtered_bundles = []
+    for bundle, b_hash in bundle_hashes:
```
yashvardhannanavati, line 402: Could you use `bundle_hash` instead of `b_hash`?

yashvardhannanavati:
@querti Overall direction looks good to me. Could you please also make sure that you use single quotes throughout? We kind of enforce that styling requirement throughout the IIB codebase.

iib/workers/tasks/build.py
querti: I tried using context manager, but the command wouldn't work for some reason. This is the only way I've managed to get it running.

querti:
@yashvardhannanavati @mprahl @lcarva pls review

querti:
@release-engineering/distribution review as well pls

iib/workers/tasks/build.py
```diff
@@ -337,6 +341,101 @@ def _get_resolved_image(pull_spec):
...
 
+@retry(attempts=2, wait_on=IIBError, logger=log)
+def _get_present_bundles(from_index, base_dir):
+    """
+    Get a list of bundles already present in the index image.
+
+    :param str from_index: index image to inspect
+    :param base_dir: base directory to create temporary files in
+    :return: list of tuples containing bundle names and tags
+    :raises IIBError: if any of the commands fail
```
lipoja, line 352: Return type can be added here `:rtype: list(tuple)`.

iib/workers/tasks/build.py
```diff
@@ -337,6 +341,101 @@ def _get_resolved_image(pull_spec):
...
+
+@retry(attempts=2, wait_on=IIBError, logger=log)
+def _filter_bundles(present_bundles, bundles):
+    """
+    Filter out bundles to only those not present in the index image.
+
+    :param list present_bundles: list of bundles present in the index image.
+        Represented as tuples containing package name and version.
+    :param list bundles: bundles requested to be added to the index image
+    :return: list of bundles not present in the index image
```
lipoja, line 419: `:rtype:` can be added here as well.

iib/workers/tasks/utils.py
```diff
@@ -280,3 +280,26 @@ def _get_function_arg_value(arg_name, func, args, kwargs):
...
+
+
+def get_free_port(start_port, end_port):
+    """
+    Get port that's not currently in use from the specified range.
+
+    :param int start_port: Start of the requested port range
+    :param int end_port: End of the requested port range
+    :return: First free port from the port range. None if no port is free.
+    :raises IIBError: If netstat command fails
```
lipoja, line 292: `:rtype:` can be added here

docker/Dockerfile-workers
```diff
@@ -34,10 +34,14 @@ RUN dnf -y install \
     python3-devel \
     python3-pip \
     skopeo \
+    net-tools \
```
lipoja, line 37: Do not forget to modify Ansible role iib-role in our gitlab instance or you will miss this package on stage/prod. 
This Dockerfile is used just for local development If I am not mistaken.

docker/Dockerfile-workers
```diff
@@ -34,10 +34,14 @@ RUN dnf -y install \
...
     && dnf clean all
 
 COPY --from=builder /usr/bin/opm /usr/bin/opm
 ADD https://github.com/estesp/manifest-tool/releases/download/v1.0.0/manifest-tool-linux-amd64 /usr/bin/manifest-tool
+RUN curl -LJO https://github.com/fullstorydev/grpcurl/releases/download/v1.7.0/grpcurl_1.7.0_linux_x86_64.tar.gz && \
+    tar -zxvf grpcurl_1.7.0_linux_x86_64.tar.gz && \
+    cp grpcurl /usr/bin
```
lipoja, line 44: I think that, Ansible script has to be modified as well.

iib/workers/tasks/build.py
```diff
@@ -784,6 +883,23 @@ def handle_add_request(
     _update_index_image_build_state(request_id, prebuild_info)
 
     with tempfile.TemporaryDirectory(prefix='iib-') as temp_dir:
+        if from_index:
+            log.info('Checking if bundles are already present in index image')
```
lipoja, line 887: If the message is the same can we put it to variable?

iib/workers/tasks/utils.py
```diff
@@ -280,3 +280,26 @@ def _get_function_arg_value(arg_name, func, args, kwargs):
     if arg_value is None and len(args) > arg_index:
         arg_value = args[arg_index]
     return arg_value
+
+
+def get_free_port(start_port, end_port):
```
rohanpm, line 285: Are you sure on this approach of parsing netstat output to find the free port?  It doesn't seem designed for parsing.

In my experience, the usual approach for this kind of thing is just to actually open the ports.  e.g. if you tried to open port `start_port` and it succeeded, you know that port was available.  You can then close it and return that number, or increment and try the next number if it wasn't available.  Just use SO_REUSEADDR on the socket to ensure it can be reused with no timeout.

That approach is racy, but equally so as the current approach (a port which appeared unused when you ran netstat could become used immediately afterward).  Not sure exactly what are your requirements around having to use a specific range of ports in the first place, maybe there is an easy way to redesign it so there's no race conditions, or maybe not.  Retries are probably enough to deal with the race condition.

iib/workers/tasks/build.py
querti: Done.

iib/workers/tasks/build.py
querti: I have added port availability check

iib/workers/tasks/build.py
querti: Done

iib/workers/tasks/build.py
```diff
@@ -337,6 +341,101 @@ def _get_resolved_image(pull_spec):
...
+            ['podman', 'cp', f'{cid}:{db_path}', base_dir],
+            exc_msg='Failed to copy DB file from the container',
+        )
+    finally:
+        run_cmd(['podman', 'rm', cid], exc_msg='Failed to delete the container')
+    local_path = os.path.join(base_dir, db_path.split('/')[-1])
+
+    free_port = get_free_port(50051, 50051 + 1000)
+    if not free_port:
+        log.error('No free port is available')
```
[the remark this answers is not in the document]
querti: Done

iib/workers/tasks/build.py
```diff
@@ -337,6 +341,101 @@ def _get_resolved_image(pull_spec):
...
+        raise RuntimeError('No free port is available')
+
+    # Get present bundle information
+    cmd = ['opm', 'registry', 'serve', '-p', str(free_port), '-d', local_path]
+    log.debug('Running the command "{0}"'.format(' '.join(cmd)))
```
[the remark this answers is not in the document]
querti: Done.

iib/workers/tasks/build.py
querti: Done.

iib/workers/tasks/build.py
querti: Done.

iib/workers/tasks/utils.py
querti: Done.

docker/Dockerfile-workers
querti: I have created additional MRs

iib/workers/tasks/build.py
querti: Done.

docker/Dockerfile-workers
querti: I have created additional MRs

iib/workers/tasks/utils.py
querti: I tried using socket.connect and socket.bind, but I couldn't get it to reliably recognize that the port is in use. I think the current solution is more reliable, even though the parsing can be a pain.

Regarding the race condition, I think that the probability of an issue is low enough that it can be left as-is. During prototyping I tried to gauge the port availability directly from the command's error message (which would solve the race condition), but had trouble catching output and making sure that the process is not left hanging. Parent method has retries, so that should minimize the probability of an issue further.

mprahl: Have you tried something like this?
https://gist.github.com/gabrielfalcao/20e567e188f588b65ba2#file-get-free-tcp-port-py-L3-L8

To workaround the race condition, you can use a loop around this function call and running the `opm` command until success or an error that doesn't relate to the port not being available.

iib/workers/tasks/build.py
```diff
@@ -337,6 +341,104 @@ def _get_resolved_image(pull_spec):
     return pull_spec_resolved
 
 
+@retry(attempts=2, wait_on=IIBError, logger=log)
+def _get_present_bundles(from_index, base_dir):
```
yashvardhannanavati, line 345: Should we split this one big function into multiple helper functions? I feel this function is doing too many things at once.

iib/workers/tasks/build.py
```diff
@@ -337,6 +341,104 @@ def _get_resolved_image(pull_spec):
...
+            rpc_proc.kill()
+
+    # Transform returned data to parsable json
+    present_bundles = json.loads('[' + bundles.replace('\n}\n{\n', '\n},\n{\n') + ']')
+    present_bundles = [(x['packageName'], x['version']) for x in present_bundles]
+    return present_bundles
+
+
+@retry(attempts=2, wait_on=IIBError, logger=log)
+def _filter_bundles(present_bundles, bundles):
```
yashvardhannanavati, line 414: Can we please rename this to `get_missing_bundles` instead? `_filter_bundles` seems a bit generic

yashvardhannanavati: Or `filter_missing bundles`?

iib/workers/tasks/build.py
```diff
@@ -784,6 +886,22 @@ def handle_add_request(
     _update_index_image_build_state(request_id, prebuild_info)
 
     with tempfile.TemporaryDirectory(prefix='iib-') as temp_dir:
+        if from_index:
+            msg = 'Checking if bundles are already present in index image'
+            log.info(msg)
+            set_request_state(request_id, 'in_progress', msg)
+            present_bundles = _get_present_bundles(from_index, temp_dir)
+            resolved_bundles = _filter_bundles(present_bundles, resolved_bundles)
```
yashvardhannanavati, line 894: Can we please have a log statement printing out `present_bundles` and `resolved_bundles` so that the user is aware?

iib/workers/tasks/build.py
querti: I have separated the code into more methods.

iib/workers/tasks/utils.py
querti: I have rewritten it to simply retry the service in case there's a port collision. It eliminates the race condition and also seems like the most reliable method to detect port collision.

iib/workers/tasks/build.py
querti: Done

iib/workers/tasks/build.py
querti: done

iib/workers/tasks/build.py
querti: I added a log which says which bundles will still be added. But I don't think printing every single bundles in the index image is a good idea

iib/workers/tasks/build.py
```diff
@@ -784,6 +931,26 @@ def handle_add_request(
...
+            if not resolved_bundles:
+                log.info(
+                    'All bundles are already present in the index image.'
+                    ' No additional operations are necessary'
+                )
+                set_request_state(
+                    request_id, 'complete', 'All bundles are already present in the index image'
+                )
+                return
+            log.info(
```
yashvardhannanavati, line 949: It might be more useful to print the ones which are already present and suggest that they will not be added again.

yashvardhannanavati: I'm thinking if we can also add a state_reason update with the same info as above. thoughts?

iib/workers/tasks/build.py
```diff
@@ -337,6 +340,150 @@ def _get_resolved_image(pull_spec):
...
+                log.info('Port %s is in use, trying another.', port)
+                port += 1
+            else:
+                raise IIBError('Command "%s" has failed with error "%s"', ' '.join(cmd), stderr)
+
+    return (port, rpc_proc)
+
+
+@retry(attempts=2, wait_on=IIBError, logger=log)
+def _get_present_bundles(from_index, base_dir, grpc_max_tries=5, grpc_wait_time=1):
```
yashvardhannanavati, line 422: How about using `get_worker_config().iib_total_attempts` for grpc_max_tries instead of static values?

yashvardhannanavati: It might be useful to make the wait_time configurable too?

querti: I have added these values to config

iib/workers/tasks/build.py
querti: I have switched the log message. But I don't think it makes much sense to put it as a state

yashvardhannanavati: We can add it later if stakeholders need it.

iib/workers/config.py
```diff
@@ -42,6 +42,11 @@ class Config(object):
     # Don't allow the worker to fetch more messages than it can handle at a time. This is so that
     # other tasks aren't starved. This will only be useful once more workers are enabled.
     worker_prefetch_multiplier = 1
+    grpc_max_tries = 5
```
yashvardhannanavati, line 45: Could you please prefix these with `iib_` and move it up alphabetically. Also, a comment explaining the use of these variables would be great. 

Also, could you please update the README with these variables?

querti: Done.

midnightercz (changes requested):

tests/test_workers/test_tasks/test_build.py
```diff
@@ -613,7 +613,11 @@ def test_skopeo_copy_fail_max_retries(mock_run_cmd):
 @mock.patch('iib.workers.tasks.build.gate_bundles')
 @mock.patch('iib.workers.tasks.build._get_resolved_bundles')
 @mock.patch('iib.workers.tasks.build._add_ocp_label_to_index')
+@mock.patch('iib.workers.tasks.build._get_present_bundles')
+@mock.patch('iib.workers.tasks.build._get_missing_bundles')
 def test_handle_add_request(
+    mock_fb,
```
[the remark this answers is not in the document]
querti: You're right, I forgot to change it

iib/workers/tasks/build.py
```diff
@@ -337,6 +340,151 @@ def _get_resolved_image(pull_spec):
...
+    :return: path to the copied database file.
+    :rtype: str
+    :raises IIBError: if any podman command fails
+    """
+    cidfile = os.path.join(base_dir, 'cidfile.txt')
+    run_cmd(
+        ['podman', 'create', '--cidfile', cidfile, from_index],
+        exc_msg='Failed to create index image container',
+    )
+    with open(cidfile, 'r') as f:
```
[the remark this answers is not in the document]
querti: Hmm, this seems cleaner since podman might return other junk.

iib/workers/tasks/build.py
```diff
@@ -337,6 +340,151 @@ def _get_resolved_image(pull_spec):
...
+    """
+    Get a list of bundles already present in the index image.
+
+    :param str from_index: index image to inspect
+    :param str base_dir: base directory to create temporary files in
+    :return: list of present bundles as provided by the grpc query
+    :rtype: list
+    :raises IIBError: if any of the commands fail
+    """
+    db_path = _get_index_database(from_index, base_dir)
```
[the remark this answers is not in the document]
querti: The container itself serves the DB, so just running it would suffice. Also, I think that grpcurl isn't installed in the container, so it would have to be run locally. 

In any case, I believe Matt was against running a container in IIB to avoid arbitrary code execution on IIB worker. If nothing else, running the DB directly has probably less overhead than running a container. Also, if we ever decide to run IIB in containers (as we do with dev-env), running another container in it might be hard to achieve (I think).

iib/workers/tasks/build.py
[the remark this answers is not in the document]
yashvardhannanavati: `➜  iib-web git:(greenwave_conf) podman create 11866d6bc6aa ls
a4289a7fcba0be5515c8c3ed90c686f7ce012d4200145377e943225148477f99`

It actually only returns the container id. Am I missing anything?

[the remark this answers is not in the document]
querti: Well, I don't really know if this command can return something other than the container ID, so to eliminate this uncertainty I decided to use the cidfile since it was designed for this purpose.

In any case, I feel like it should be ok one way or the other.

midnightercz:
For some reason I'm not able to review this PR again. So +1

README.md
```diff
@@ -218,6 +218,11 @@ The custom configuration options for the Celery workers are listed below:
   file though. This defaults to `~/.docker/config.json.template`.
 * `iib_greenwave_url` - the URL to the Greenwave REST API if gating is desired
   (e.g. `https://greenwave.domain.local/api/v1.0/`). This defaults to `None`.
+* `iib_grpc_init_wait_time` - time to wait for the index image service to be initialized
```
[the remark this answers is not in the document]
querti: done

README.md
```diff
@@ -218,6 +218,11 @@ The custom configuration options for the Celery workers are listed below:
...
+* `iib_grpc_max_port_tries` - maximum ports to try when initializing the index image service
+* `iib_grpc_max_tries` - maximum number of attempts to make when getting data from the service
+* `iib_grpc_start_port` - first port to try when starting the service (subsequent are increments)
+* `iib_grpc_wait_time` - time to wait between failed grpc queries
```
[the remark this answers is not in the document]
querti: done

docker/Dockerfile-workers
```diff
@@ -38,6 +38,9 @@ RUN dnf -y install \
 
 COPY --from=builder /usr/bin/opm /usr/bin/opm
 ADD https://github.com/estesp/manifest-tool/releases/download/v1.0.0/manifest-tool-linux-amd64 /usr/bin/manifest-tool
+RUN curl -LJO https://github.com/fullstorydev/grpcurl/releases/download/v1.7.0/grpcurl_1.7.0_linux_x86_64.tar.gz && \
```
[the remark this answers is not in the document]
querti: done

docker/Dockerfile-workers
```diff
@@ -38,6 +38,9 @@ RUN dnf -y install \
...
+    tar -zxvf grpcurl_1.7.0_linux_x86_64.tar.gz && \
```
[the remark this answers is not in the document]
querti: done

iib/workers/tasks/build.py
```diff
@@ -337,6 +340,151 @@ def _get_resolved_image(pull_spec):
...
+        cid = f.read()
+    os.remove(cidfile)
+
+    try:
+        data = json.loads(
+            run_cmd(
+                ['podman', 'image', 'inspect', from_index], exc_msg='Failed to fetch image metadata'
+            )
+        )
+        db_path = data[0]['Config']['Labels']['operators.operatorframework.io.index.database.v1']
```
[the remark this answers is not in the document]
querti: done

iib/workers/tasks/build.py
```diff
@@ -337,6 +340,151 @@ def _get_resolved_image(pull_spec):
...
+        run_cmd(
+            ['podman', 'cp', f'{cid}:{db_path}', base_dir],
+            exc_msg='Failed to copy DB file from the container',
+        )
+    finally:
+        run_cmd(['podman', 'rm', cid], exc_msg='Failed to delete the container')
+    local_path = os.path.join(base_dir, db_path.split('/')[-1])
```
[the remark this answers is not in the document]
querti: done

iib/workers/tasks/build.py
```diff
@@ -337,6 +340,151 @@ def _get_resolved_image(pull_spec):
     return pull_spec_resolved
 
 
+def _get_index_database(from_index, base_dir):
```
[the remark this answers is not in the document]
querti: done

iib/workers/tasks/build.py
```diff
@@ -337,6 +340,151 @@ def _get_resolved_image(pull_spec):
...
+
+def _serve_index_registry(db_path):
+    """
+    Locally start OPM registry service, which can be communicated with using gRPC queries.
+
+    Due to IIB's paralellism, the service can run multiple times, which could lead to port
+    binding conflicts. Resolution of port conflicts is handled in this function as well.
+
+    :param str db_path: path to index database containing the registry data
+    :return: tuple containing port number of the running service and the runnung Popen object
```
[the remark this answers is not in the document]
querti: done

iib/workers/tasks/build.py
```diff
@@ -337,6 +340,151 @@ def _get_resolved_image(pull_spec):
...
+            stderr = rpc_proc.stderr.read()
+            if 'address already in use' in stderr:
+                if i == conf['iib_grpc_max_port_tries'] - 1:
+                    err_msg = (
+                        'No free port has been found after %s attempts.'
+                        % conf['iib_grpc_max_port_tries']
+                    )
+                    log.error(err_msg)
+                    raise IIBError(err_msg)
+                log.info('Port %s is in use, trying another.', port)
```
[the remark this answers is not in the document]
querti: done

iib/workers/tasks/build.py
```diff
@@ -337,6 +340,151 @@ def _get_resolved_image(pull_spec):
...
+    :raises IIBError: if all tried ports are in use, or the command failed for another reason
+    """
+    conf = get_worker_config()
+    port = conf['iib_grpc_start_port']
+    for i in range(conf['iib_grpc_max_port_tries']):
+        cmd = ['opm', 'registry', 'serve', '-p', str(port), '-d', db_path]
+        rpc_proc = subprocess.Popen(
+            cmd, stdout=subprocess.PIPE, stderr=subprocess.PIPE, universal_newlines=True
+        )
+        time.sleep(conf['iib_grpc_init_wait_time'])
```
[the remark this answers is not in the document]
querti: done

iib/workers/tasks/build.py
```diff
@@ -337,6 +340,151 @@ def _get_resolved_image(pull_spec):
...
+        ret = rpc_proc.poll()
+        if not ret:
+            log.debug('Running the command "%s"', ' '.join(cmd))
```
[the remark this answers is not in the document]
querti: done

iib/workers/tasks/build.py
```diff
@@ -337,6 +340,151 @@ def _get_resolved_image(pull_spec):
...
+    conf = get_worker_config()
+
+    for i in range(conf['iib_grpc_max_tries']):
+        try:
+            bundles = run_cmd(
+                ['grpcurl', '-plaintext', f'localhost:{port}', 'api.Registry/ListBundles'],
+                exc_msg='Failed to get bundle data from index image',
+            )
+            break
+        except IIBError:
```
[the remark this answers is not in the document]
querti: done

iib/workers/tasks/build.py
```diff
@@ -337,6 +340,151 @@ def _get_resolved_image(pull_spec):
...
+            if i == conf['iib_grpc_max_tries'] - 1:
+                log.error('Data cannot be retrieved from the index image database gRPC.')
+                raise
+            time.sleep(conf['iib_grpc_wait_time'])
+        finally:
+            rpc_proc.kill()
+
+    # Transform returned data to parsable json
+    present_bundles = json.loads('[' + bundles.replace('\n}\n{\n', '\n},\n{\n') + ']')
```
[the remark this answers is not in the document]
querti: done

iib/workers/tasks/build.py
```diff
@@ -337,6 +340,151 @@ def _get_resolved_image(pull_spec):
...
+    return present_bundles
+
+
+@retry(attempts=2, wait_on=IIBError, logger=log)
+def _get_missing_bundles(present_bundles, bundles):
+    """
+    Filter out bundles to only those not present in the index image.
+
+    :param list present_bundles: list of bundles present in the index image, as provided by opm.
```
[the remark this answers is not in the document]
querti: done

iib/workers/tasks/build.py
```diff
@@ -337,6 +340,151 @@ def _get_resolved_image(pull_spec):
...
+    :param list bundles: bundles requested to be added to the index image
+    :return: list of bundles not present in the index image
+    :rtype: list
+    """
+    # we only need package name and version
+    present_bundles = [(x['packageName'], x['version']) for x in present_bundles]
```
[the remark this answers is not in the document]
querti: done

iib/workers/tasks/build.py
```diff
@@ -337,6 +340,151 @@ def _get_resolved_image(pull_spec):
...
+        data = skopeo_inspect(f'docker://{bundle}')
+        bundles_info.append(
+            (
+                bundle,
+                data["Labels"]["operators.operatorframework.io.bundle.package.v1"],
+                data["Labels"]["version"],
+            )
+        )
+
+    filtered_bundles = []
```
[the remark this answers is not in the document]
querti: done

querti:
@mprahl  @yashvardhannanavati During additional testing I noticed that the current bundle matching mechanism might not be sufficient. Currently, IIB assumes that two bundles are the same if their package name and version matches. However, the bundle 'registry-proxy.engineering.redhat.com/rh-osbs/openshift-service-mesh-kiali-operator-metadata:1.0-7' has a version '1.0', but in the index image database the version is '1.0.5'. Thus, they are evaluated as different and unique constraint error happens once again. Not sure if the metadata is just invalid, or if this is something that can happen in real-life use-cases. If it could, then a different approach to match bundles must be used. The previous iteration used images hashes, but ran intio a similar issue, as non-matching hash could still mean matching name+version. 

Thoughts?

yashvardhannanavati:
@querti I would suggest going down the route of hash. If the hashes differ, then the bundles have to be different afaik. Unless you are comparing a manifest list sha to a manifest sha

querti:
@yashvardhannanavati During prototyping I got 'unique constraint' DB error despite the bundle hashes being different, which is why I changed the approach in the first place (I don't remember which bundle it was, so I can't provide an example). I don't know if this is a peculiarity of the stage data, or if this can occur in production, but I'd go with the approach which has the most false negatives. Maybe using both checks would be the most efficient. I'd say it's still worth merging this even if it doesn't eliminate all duplicates (worst case scenario is that IIB will fail with unique constraint error). Thoughts?

yashvardhannanavati:
I would still suggest using the hash comparison approach. It's most likely to have less false negatives.

@lcarva Can I please ask if you have encountered any such instances before?

querti:
@yashvardhannanavati I decided to add both conditions, and if either one matches the bundle, then it will be skipped when adding. One issue I noticed with the hash approach is that 'bundlePath' attribute received from the index image does not have to necessarily contain the image hash, but instead could be specified via tag (I think it depends on how it was specified when added to the index image). If it's defined via tag, I've added a check for this tag as well. If we wanted to force matching with hashes, we would probably need to resolve every bundle image in the index image, which probably wouldn't scale very well. 

I think this is kind of an edge case since most bundles are added via iib, and iib adds them specified by hash.

lcarva:
> I would still suggest using the hash comparison approach. It's most likely to have less false negatives.
> 
> @lcarva Can I please ask if you have encountered any such instances before?

Sorry, instances of what? :smiley:

iib/workers/tasks/build.py
```diff
@@ -337,6 +340,151 @@ def _get_resolved_image(pull_spec):
...
+        ret = rpc_proc.poll()
+        if not ret:
+            log.debug('Running the command "%s"', ' '.join(cmd))
+            break
+        else:
+            stderr = rpc_proc.stderr.read()
+            if 'address already in use' in stderr:
+                if i == conf['iib_grpc_max_port_tries'] - 1:
+                    err_msg = (
+                        'No free port has been found after %s attempts.'
```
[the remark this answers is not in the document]
querti: done

iib/workers/tasks/build.py
```diff
@@ -337,6 +340,151 @@ def _get_resolved_image(pull_spec):
...
+                        % conf['iib_grpc_max_port_tries']
+                    )
+                    log.error(err_msg)
+                    raise IIBError(err_msg)
+                log.info('Port %s is in use, trying another.', port)
+                port += 1
+            else:
+                raise IIBError('Command "%s" has failed with error "%s"', ' '.join(cmd), stderr)
```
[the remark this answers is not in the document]
querti: done

iib/workers/tasks/build.py
```diff
@@ -337,6 +342,156 @@ def _get_resolved_image(pull_spec):
...
+                stdout = get_running_subprocess_output(rpc_proc.stdout)
+                if 'serving registry' in stdout:
+                    registry_initialized = True
+                    log.debug('Started the command "%s"', ' '.join(cmd))
+                    break
+                elif j == conf['iib_grpc_max_tries'] - 1:
+                    rpc_proc.kill()
+                    raise IIBError(
+                        'Index registry has not been initialized in %d seconds'
+                        % (conf['iib_grpc_init_wait_time'] * conf['iib_grpc_max_tries'])
```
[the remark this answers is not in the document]
querti: done

iib/workers/tasks/build.py
```diff
@@ -337,6 +342,156 @@ def _get_resolved_image(pull_spec):
...
+    """
+    Filter out bundles to only those not present in the index image.
+
+    :param list present_bundles: list of bundles present in the index image, as provided by opm.
+    :param list bundles: bundles requested to be added to the index image.
+    :return: list of bundles not present in the index image.
+    :rtype: list
+    """
+    # we only need package name and version
+    present_bundle_ids = [(x['packageName'], x['version']) for x in present_bundles]
```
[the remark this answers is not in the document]
querti: done

iib/workers/tasks/build.py
```diff
@@ -337,6 +342,157 @@ def _get_resolved_image(pull_spec):
...
+                    )
+                # not yet initialized, retry the loop
+                else:
+                    # sometimes, the registry service doesn't output any messages despite running
+                    # restarting it is an attempt to mitigate this issue
+                    rpc_proc.kill()
+                    continue
+            # process has terminated, something went wrong
+            else:
+                break
```
yashvardhannanavati, line 409: Could you please add a log message here for easy debugging?

iib/workers/tasks/build.py
```diff
@@ -337,6 +342,157 @@ def _get_resolved_image(pull_spec):
...
+        if registry_initialized:
```
yashvardhannanavati, line 410: Could you please add a log message here for easy debugging?

iib/workers/tasks/build.py
```diff
@@ -337,6 +342,157 @@ def _get_resolved_image(pull_spec):
...
+                elif j == conf['iib_grpc_max_tries'] - 1:
+                    rpc_proc.kill()
+                    raise IIBError(
+                        'Index registry has not been initialized after '
+                        f'{conf.get("iib_grpc_max_tries")} tries'
+                    )
+                # not yet initialized, retry the loop
+                else:
+                    # sometimes, the registry service doesn't output any messages despite running
+                    # restarting it is an attempt to mitigate this issue
```
yashvardhannanavati, line 404: Could you please add a log message here for easy debugging?

iib/workers/tasks/build.py
```diff
@@ -337,6 +342,157 @@ def _get_resolved_image(pull_spec):
...
+    :param list bundles: bundles requested to be added to the index image.
+    :return: list of bundles not present in the index image.
+    :rtype: list
+    """
+    # we only need package name and version
+    present_bundle_ids = []
+    present_bundle_ids_path = []
+    present_bundle_hashes = []
+    for bundle in present_bundles:
+        present_bundle_ids.append((bundle['packageName'], bundle['version']))
```
yashvardhannanavati, line 470: I am sorry I am a bit confused with `present_bundle_ids` here. Why can't we only use `present_bundle_hashes` and `present_bundle_ids_path`? 

IIUC, `present_bundle_ids_path` is going to be (Name, version-release) which should be sufficient to compare in the second for loop you have to go over the bundles submitted by the user. Am I missing something here?

iib/workers/tasks/build.py
```diff
@@ -785,6 +941,28 @@ def handle_add_request(
     _update_index_image_build_state(request_id, prebuild_info)
 
     with tempfile.TemporaryDirectory(prefix='iib-') as temp_dir:
+        if from_index:
+            msg = 'Checking if bundles are already present in index image'
+            log.info(msg)
+            set_request_state(request_id, 'in_progress', msg)
+            present_bundles = _get_present_bundles(from_index, temp_dir)
+            filtered_bundles = _get_missing_bundles(present_bundles, resolved_bundles)
+            excluded_bundles = [b for b in resolved_bundles if b not in filtered_bundles]
```
yashvardhannanavati, line 950: Can we please use `bundle` instead of `b`. I'm usually not a fan of single letter variable names unless absolutely required.

iib/workers/tasks/build.py
querti: I think it's not necessary, as it's analysed and communicated with the user in better detail in the following code.

iib/workers/tasks/build.py
querti: done

iib/workers/tasks/build.py
querti: done

iib/workers/tasks/build.py
querti: Basically, 'bundlePath' entry provided by opm will contain bundle specified by digest about 99% of the time (as this is how it was added to the index image). On the other hand, nothing prevents it being specified by tag, so I wanted to account for that possibility as well.

During prototyping and testing I found out that the data matching might still result in some false negatives, so I wanted to make the process as robust as possible. This is why i compare both name+version and digests. Unfortunately, neither name+version difference nor digest difference guarantee that opm won't give unique constraint error. But I should note that I tested this on stage data, and in production these inconsistencies will be far less likely.

iib/workers/tasks/build.py
querti: done

iib/workers/tasks/build.py
yashvardhannanavati: Correct, I get the point of checking both NVR and the sha but I don't think you have to keep a third list of present_bundle_ids which is essentially only NV. You will anyways check NVR when you compare the list with `present_bundle_ids_path`. So I think you can simply remove the parts with `present_bundle_ids`. Unless I am missing something.

yashvardhannanavati (changes requested):
Some final touches! Let's merge it once these comments are resolved. Thanks Lubo! :)

iib/workers/tasks/build.py
```diff
@@ -337,6 +342,162 @@ def _get_resolved_image(pull_spec):
...
+    try:
+        db_path = data['Labels']['operators.operatorframework.io.index.database.v1']
+    except KeyError:
+        raise IIBError('Index image doesn\'t have the label specifying its database location.')
+    _copy_files_from_image(from_index, db_path, base_dir)
+    local_path = os.path.join(base_dir, os.path.basename(db_path))
+    return local_path
+
+
+def _serve_index_registry(db_path):
```
lcarva, line 365: The [cyclomatic complexity|https://en.wikipedia.org/wiki/Cyclomatic_complexity] of this function is high, 11 (give or take). Can we make use of custom exceptions and helper methods so it's easier to read?


```python
def _serve_index_registry(db_path):
  for port in range(...):
    try:
      return (port, _serve_index_registry_at_port(db_path, port))
    except AddressAlreadyInUse:
      pass

  # If it got here, no ports are available
  raise IIBError(...)

def _serve_index_registry_at_port(db_path, port):
  for _ in range(...):
    rpc_proc = subprocess.Popen(...)
    ret = rpc_proc.poll()

    if ret is not None:
       # Process terminated unexpectedly 
       if 'address already in use' in stderr:
         raise AddressAlreadyInUse(...)
       raise IIBError(...)

     if 'serving registry' in stdout:
       return rpc_proc

     # We always want to kill the process before trying again, do it unconditionally
     rpc_proc.kill()

  # If it got here, index registry has not been initialized after all retries
  raise IIBError(...)
```

iib/workers/tasks/build.py
querti: As I've said, I encountered some divergence in data provided by grpc queries and actual index image data. For example, grpc said bundle package name + version were 'bundle:1.0', while in reality they were saved in the index image as 'bundle:1.0.5'. This ccaused a unique constraint error, even though my script hasn't detected a match. I don't know if this is likely (or possible) to happen in production data, but I thought that increasing redundancy will reduce the probability of these inconsistencies.

So, I check for a match in packagename+version, then if bundle was specified by digest (most cases), I check for a match in digests, and if it was specified by tag then I check for name+version again from a different source.

iib/workers/tasks/build.py
```diff
@@ -337,6 +342,162 @@ def _get_resolved_image(pull_spec):
...
+    binding conflicts. Resolution of port conflicts is handled in this function as well.
+
+    :param str db_path: path to index database containing the registry data.
+    :return: tuple containing port number of the running service and the running Popen object.
+    :rtype: (int, Popen)
+    :raises IIBError: if all tried ports are in use, or the command failed for another reason.
+    """
+    conf = get_worker_config()
+    port = conf['iib_grpc_start_port']
+    for i in range(conf['iib_grpc_max_port_tries']):
```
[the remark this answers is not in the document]
querti: done

README.md
```diff
@@ -220,6 +220,16 @@ The custom configuration options for the Celery workers are listed below:
...
   (e.g. `https://greenwave.domain.local/api/v1.0/`). This defaults to `None`.
+* `iib_grpc_init_wait_time` - time to wait for the index image service to be initialized. This
+  defaults to `1` second.
+* `iib_grpc_max_port_tries` - maximum ports to try when initializing the index image service.
+  This defaults to `100` tries.
+* `iib_grpc_start_port` - first port to try when starting the service (subsequent are increments).
+  This defaults to `50051`.
+* `iib_grpc_max_tries` - maximum number of times to try to start the index image service
+  before giving up. This defaults to `5` attempts.
+* `iib_grpc_wait_time` - time to wait between checking if index image service has initialized.
```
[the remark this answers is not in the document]
querti: That's 'iib_grpc_init_wait_time'. 'iib_grpc_wait_time' is a remnant of previous implementation I forgot to remove

iib/workers/tasks/build.py
querti: I have simplified the code to contain less nested conditions

iib/workers/tasks/build.py
```diff
@@ -337,6 +342,166 @@ def _get_resolved_image(pull_spec):
...
+    :return: list of bundles not present in the index image.
+    :rtype: list
+    """
+    # we only need package name and version
+    present_bundle_ids = []
+    present_bundle_ids_path = []
+    present_bundle_hashes = []
+    for bundle in present_bundles:
+        present_bundle_ids.append((bundle['packageName'], bundle['version']))
+        # bundlePath could be defined either by hash or tag. Regardless, we'll use it for matching
```
mprahl, line 480: Could the tag be a floating tag? If so, we wouldn't want to match against that. I don't have enough context to know for sure.

I imagine that this is assuming that the tag is an NVR, however, that is an implementation specific to the internal Red Hat pipeline and not generic.

iib/workers/tasks/build.py
```diff
@@ -337,6 +342,166 @@ def _get_resolved_image(pull_spec):
...
+            present_bundle_hashes.append(bundle['bundlePath'].split('@sha256:')[-1])
+        else:
+            present_bundle_ids_path.append(
+                (bundle['packageName'], bundle['bundlePath'].split(':')[-1])
+            )
+    filtered_bundles = []
+    for bundle in bundles:
+        data = skopeo_inspect(f'docker://{bundle}')
+        bundle_id = (
+            data['Labels']['operators.operatorframework.io.bundle.package.v1'],
```
mprahl, line 491: Are these labels guaranteed to be set? I'd want to avoid an unexpected `KeyError`.

mprahl (changes requested):

iib/workers/tasks/build.py
```diff
@@ -337,6 +342,166 @@ def _get_resolved_image(pull_spec):
...
+    :raises IIBError: if te process has failed to initialize too many times, or an unexpected
+        error occured.
+    :raises AddressAlreadyInUse: if the specified port is already being used by another service.
+    """
+    cmd = ['opm', 'registry', 'serve', '-p', str(port), '-d', db_path]
+    for attempt in range(max_tries):
+        rpc_proc = subprocess.Popen(
+            cmd, stdout=subprocess.PIPE, stderr=subprocess.PIPE, universal_newlines=True
+        )
+        time.sleep(wait_time)
```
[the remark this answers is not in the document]
querti: The problem I most frequently encountered when prototyping was that ocp would arbitrarily not output any messages at all, requiring a restart. This is why I opted for restarting the service multiple times instead of retrying for the output.

I suppose I could add the querying mechanism if we expect  a very high variability in start-up time (If we expect the service start up time to sometimes take 10+ seconds then the passive wait time would be inefficient), but I haven't encountered this problem during prototyping. Then again, I didn't really stress test it. 

Thoughts?

iib/workers/tasks/build.py
querti: Should I remove this? Should we ask someone else?

iib/workers/tasks/build.py
querti: If they're not guaranteed then we're in pretty big trouble. The only remaining match strategy is digest comparison, and digest doesn't have to be provided via opm grpc.

midnightercz (approved):

iib/workers/tasks/build.py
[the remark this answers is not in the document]
mprahl: I feel strongly about not relying on sleeps.

iib/workers/tasks/build.py
mprahl: Could we perhaps add this label to the `iib_required_labels` configuration in `configure_celery` so that it can't be overwritten by the administrator?

querti: It seems that this check verifies both label key and value. The values will differ for each bundle, so this check is probably not usable for us.

iib/workers/tasks/build.py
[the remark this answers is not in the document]
querti: I added a loop to check program output until timeout. But I had to add a one second sleep between attempts because otherwise the program didn't give me any output for some reason. Also, I feel that adding CPU intensive active loops is not a good idea. 

I increased the default timeout to 3 seconds so that it would loop at least a couple times and because I don't know the precision of time provided by the system.

iib/workers/tasks/build.py
```diff
@@ -324,8 +329,8 @@ def _get_resolved_image(pull_spec):
     """
     Get the pull specification of the container image using its digest.
 
-    :param str pull_spec: the pull specification of the container image to resolve
-    :return: the resolved pull specification
+    :param str pull_spec: the pull specification of the container image to resolve.
```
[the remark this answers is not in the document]
querti: removed.

iib/workers/tasks/build.py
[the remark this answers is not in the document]
mprahl: Sorry, I should be clearer. I feel strongly about not relying on an arbitrary sleep time to hope that it was long enough. Using sleeps to create small delays between checks is totally fine.

iib/workers/tasks/build.py
```diff
@@ -337,6 +342,152 @@ def _get_resolved_image(pull_spec):
...
+        exc_msg='Failed to get bundle data from index image',
+    )
+    rpc_proc.kill()
+
+    # Transform returned data to parsable json
+    present_bundles = [json.loads(bundle) for bundle in re.split(r'(?<=})\n(?={)', bundles)]
+    return present_bundles
+
+
+@retry(attempts=2, wait_on=IIBError, logger=log)
```
yashvardhannanavati, line 463: I don't think this one needs to have a retry.

iib/workers/tasks/build.py
```diff
@@ -337,6 +342,152 @@ def _get_resolved_image(pull_spec):
...
+    :param list present_bundles: list of bundles present in the index image, as provided by opm.
+    :param list bundles: resolved bundles requested to be added to the index image.
+    :return: list of bundles not present in the index image.
+    :rtype: list
+    """
+    present_bundle_hashes = []
+    filtered_bundles = []
+    for bundle in present_bundles:
+        if '@sha256:' not in bundle['bundlePath']:
+            raise IIBError(
```
yashvardhannanavati, line 477: I don't think we should raise an error here if the index image contains a bundle image specified by tag. I would recommend removing this check and simply skipping that bundle

yashvardhannanavati: The reason I say this is if the bundle with tag clashes with a bundle image that we have a sha for, OPM will anyways throw an error. So let OPM handle that.

iib/workers/tasks/build.py
```diff
@@ -785,6 +936,30 @@ def handle_add_request(
     _update_index_image_build_state(request_id, prebuild_info)
 
     with tempfile.TemporaryDirectory(prefix='iib-') as temp_dir:
+        if from_index:
+            msg = 'Checking if bundles are already present in index image'
+            log.info(msg)
+            set_request_state(request_id, 'in_progress', msg)
+            present_bundles = _get_present_bundles(from_index, temp_dir)
+            filtered_bundles = _get_missing_bundles(present_bundles, resolved_bundles)
+            excluded_bundles = [
```
yashvardhannanavati, line 945: You might want to put this after the `if` block to create this variable only if necessary.

yashvardhannanavati (changes requested):
Minor changes

iib/workers/tasks/build.py
querti: Removed.

iib/workers/tasks/build.py
querti: Removed.

iib/workers/tasks/build.py
```diff
@@ -337,6 +342,146 @@ def _get_resolved_image(pull_spec):
...
+                log.info('Index registry service has been initialized.')
+                return rpc_proc
+
+        rpc_proc.kill()
+
+    raise IIBError(f'Index registry has not been initialized after {max_tries} tries')
+
+
+@retry(attempts=2, wait_on=IIBError, logger=log)
+def _get_present_bundles(from_index, base_dir):
```
yashvardhannanavati, line 439: Instead of having a retry here, I would rather have it on `_serve_index_registry_at_port` and `_serve_index_registry`

iib/workers/tasks/build.py
querti: Resolved bundles actually gets overwritten by filtered bundles the following line, so that wouldn't work. I guess I could not overwrite it and use filtered_bundles instead, but that would break the function's convention.

So I decided to keep it as is. I hope I don't get -1 for this lol

iib/workers/tasks/build.py
querti: I have removed this retry and added a retry to _serve_index_registry_at_port. However, I don't see any operations in _serve_index_registry that could benefit from a retry.

querti:
@yashvardhannanavati I have resolved the issue where celery hangs at 100% CPU for seemingly no reason. It seems to be a rare celery bug, and I don't know exactly what happens under the hood but I know the steps for reproducing it.

It's caused by blocking reading the output of a running process by a thread. Killing that process will make Celery go haywire. 

To solve it I changed how IIB determines if the opm service has been initialized. Instead of reading its output, IIB does a simple query, and if it gets an expected answer then the service is considered to be initialized. Otherwise, it waits and tries again.

Do you consider this solution acceptable? I believe it's functionally equivalent and much safer than the previous one.

iib/workers/tasks/build.py
yashvardhannanavati: Lol no, I don't give -1's. This works :)

yashvardhannanavati (approved):

yashvardhannanavati:
@mprahl and @lcarva are good with our review and I think we can merge this. :)

PR #138, opened 2020-08-06
Create .python-version file

lcarva:
This allows dependabot to run the expected version of python when
updating dependencies, see #124 PR


.python-version
```diff
@@ -0,0 +1 @@
+3.6.8
```
mprahl, line 1: Could you add a comment explaining that this is tied to the Python version in RHEL 8?

lcarva: It looks like comments are not allowed.

PR #139, opened 2020-08-07
Add support for podman compose

lcarva:
This is rough draft of what it takes to run IIB with podman.

I managed to make this work on fedora 31 by using my fork of podman-compose on the branch super-duper, https://github.com/lcarva/podman-compose/tree/super-duper. This fork has a few different things that improve the experience drastically (PRs have been created for each commit):

- Add support for `down -v` so volumes are removed.
- Create non-existing binding volumes.

But more importantly, it's based on the devel branch of podman-compose which contains a lot of fixes not found in the rpm distribution.

The changes in this PR are the bare minimum to work with podman-compose. Each commit, hopefully, explain why the change is needed.

If this is something we want to pursue further, we can probably create a `Makefile` to handle the copy of the system ca bundle while starting up the docker/podman-compose.

### DIFF
--- a/docker/Dockerfile-workers
+++ b/docker/Dockerfile-workers
@@ -26,7 +26,9 @@ RUN dnf -y install \
     --setopt=deltarpm=0 \
     --setopt=install_weak_deps=false \
     --setopt=tsflags=nodocs \
+    /etc/containers/storage.conf \
     buildah \
+    fuse-overlayfs \
     gcc \
     krb5-devel \
     https://kojipkgs.fedoraproject.org/packages/podman/1.7.0/3.fc30/x86_64/podman-1.7.0-3.fc30.x86_64.rpm \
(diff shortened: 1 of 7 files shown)


docker/Dockerfile-workers
```diff
@@ -16,21 +16,20 @@ RUN git clone https://github.com/operator-framework/operator-registry $OPM_SRC &
     (cd $OPM_SRC && git checkout $OPM_TAG && make build) && \
     cp $OPM_SRC/bin/opm /usr/bin/opm
 
-FROM centos:8
```
mprahl, line 19: Don't we want to develop against CentOS 8 since that is the OS we deploy IIB on? Otherwise, we could end up using newer skopeo/podman features without realizing that they don't work in the RPMs present in CentOS 8.

lcarva: Yeah... I couldn't make it work with centos at all :disappointed:

lcarva: Finally got this to work with `centos:8`. It turns out that the changes I had to do in the Dockerfile also enabled this.

yashvardhannanavati:
@lcarva Is this still a draft? Or is it ready for review?

lcarva:
> @lcarva Is this still a draft? Or is it ready for review?

I think the one thing I'd like to do is create a little `Makefile` so it's a little bit easier to run, and document how to user podman-compose vs docker-compose. I'll do that in the next couple of weeks.

lcarva:
This is ready for review and no longer in draft state.

mprahl (approved):

lcarva:
Fixed the travis issue by rebasing on master and picking up the safety fix.

yashvardhannanavati (approved):

PR #141, opened 2020-08-17
Adding ability to update binary image of index image from Add

shawn-hurley:
* making bundles optional
* if bundles are optional then from_index and binary_image are required

### DIFF
--- a/iib/web/models.py
+++ b/iib/web/models.py
@@ -803,12 +803,23 @@ class RequestAdd(Request, RequestIndexImageMixin):
         request_kwargs = deepcopy(kwargs)
 
         bundles = request_kwargs.get('bundles', [])
-        if (
-            not isinstance(bundles, list)
-            or len(bundles) == 0
-            or any(not item or not isinstance(item, str) for item in bundles)
+        if not isinstance(bundles, list) or any(
+            not item or not isinstance(item, str) for item in bundles
         ):
-            raise ValidationError(f'"bundles" should be a non-empty array of strings')
+            raise ValidationError(
+                '"bundles" should be either an empty array or an array of non-empty strings'
+            )
+
+        # Check if no bundles `from_index and `binary_image` are specified
+        # if no bundles and and no from index then a empty index will be created
+        # if no binary image and just a from_index then we are not updating anything and it would
+        # be a no-op
+        if not request_kwargs.get('bundles') and (
+            not request_kwargs.get('from_index') or not request_kwargs.get('binary_image')
+        ):
+            raise ValidationError(
+                '"from_index" and "binary_image" must be specified if no bundles are specified'
+            )
 
         for param in ('cnr_token', 'organization'):
             if param not in request_kwargs:
@@ -826,8 +837,7 @@ class RequestAdd(Request, RequestIndexImageMixin):
 
         cls._from_json(
             request_kwargs,
-            additional_required_params=['bundles'],
-            additional_optional_params=['from_index', 'organization'],
+            additional_optional_params=['from_index', 'organization', 'bundles'],
             batch=batch,
         )
 
--- a/iib/workers/tasks/build.py
+++ b/iib/workers/tasks/build.py
@@ -360,13 +360,14 @@ def _opm_index_add(
     """
     # The bundles are not resolved since these are stable tags, and references
     # to a bundle image using a digest fails when using the opm command.
+    bundle_str = ','.join(bundles) or "''"
     cmd = [
         'opm',
         'index',
         'add',
         '--generate',
         '--bundles',
-        ','.join(bundles),
+        bundle_str,
         '--binary-image',
         binary_image,
     ]
--- a/tests/test_web/test_api_v1.py
+++ b/tests/test_web/test_api_v1.py
@@ -230,15 +230,6 @@ def test_get_build_logs_not_configured(client, db, minimal_request_add):
         ),
         (
             {
-                'bundles': [],
-                'from_index': 'pull:spec',
-                'binary_image': 'binary:img',
-                'add_arches': ['s390x'],
-            },
-            '"bundles" should be a non-empty array of strings',
-        ),
-        (
-            {
                 'bundles': ['some:thing'],
                 'from_index': 32,
                 'binary_image': 'binary:image',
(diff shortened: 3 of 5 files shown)


iib/web/models.py
```diff
@@ -803,10 +811,8 @@ def from_json(cls, kwargs, batch=None):
...
 
         bundles = request_kwargs.get('bundles', [])
-        if (
-            not isinstance(bundles, list)
-            or len(bundles) == 0
-            or any(not item or not isinstance(item, str) for item in bundles)
+        if not isinstance(bundles, list) or any(
+            not item or not isinstance(item, str) for item in bundles
         ):
             raise ValidationError(f'"bundles" should be a non-empty array of strings')
```
lcarva, line 817: Empty lists are now allowed. Should this say `"bundles" should be either an empty array or an array of non-empty strings`?

iib/web/models.py
```diff
@@ -823,11 +829,9 @@ def from_json(cls, kwargs, batch=None):
         # Remove attributes that are not stored in the database
         request_kwargs.pop('cnr_token', None)
         request_kwargs.pop('force_backport', None)
-
```
lcarva, line 826: Accidental line removal?

iib/web/models.py
lcarva: Unrelated to your changes, we should remove the `f` prior to the string literal.

iib/workers/tasks/build.py
```diff
@@ -360,13 +360,16 @@ def _opm_index_add(
     """
     # The bundles are not resolved since these are stable tags, and references
     # to a bundle image using a digest fails when using the opm command.
+    bundle_str = "''"
+    if bundles:
+        bundle_str = ','.join(bundles)
```
lcarva, line 365: Consider: 
```python
bundle_str = ','.join(bundles) or "''"
```

tests/test_web/test_api_v1.py
```diff
@@ -317,10 +308,6 @@ def test_add_bundles_overwrite_not_allowed(mock_smfsc, client, db):
 @pytest.mark.parametrize(
     'data, error_msg',
     (
-        (
-            {'from_index': 'pull:spec', 'binary_image': 'binary:image', 'add_arches': ['s390x']},
-            '"operators" should be a non-empty array of strings',
```
lcarva, line 322: Accidental removal?

tests/test_web/test_api_v1.py
```diff
@@ -371,13 +358,17 @@ def test_rm_operators_overwrite_not_allowed(mock_smfsc, client, db):
...
+            'Missing required parameter(s): binary_image',
         ),
         (
-            {'bundles': ['some:thing'], 'from_index': 'pull:spec', 'add_arches': ['s390x']},
+            {'from_index': 'pull:spec', 'add_arches': ['s390x']},
             'Missing required parameter(s): binary_image',
         ),
+        (
+            {'add_arches': ['s390x'], 'binary_image': 'binary:image'},
+            '"from_index" and "binary_image" must be specified if no bundles are specified',
```
lcarva, line 370: Let's add two more tests for when only either of them are missing.

tests/test_web/test_api_v1.py
```diff
@@ -461,18 +452,30 @@ def test_add_bundle_from_index_and_add_arches_missing(mock_smfsc, db, auth_env,
     mock_smfsc.assert_not_called()
 
 
-@pytest.mark.parametrize('overwrite_from_index', (False, True))
+@pytest.mark.parametrize(
```
lcarva, line 455: Would it make sense to do this instead?
```python
@pytest.mark.parametrize('overwrite_from_index', (False, True))
@pytest.mark.parametrize('bundles', (['some:thing'], []))
@pytest.mark.parametrize('from_index', ('some:thing', None))
```

Then let pytest create the permutations.

iib/web/models.py
```diff
@@ -678,6 +678,14 @@ def _from_json(
             request_kwargs, required_params=required_params, optional_params=optional_params,
         )
 
+        # Check if no bundles `from_index and `binary_image` are specified
```
lcarva, line 681: Could you clarify why `from_index` is needed? If omitted, wouldn't a new empty database be created?

tests/test_web/test_api_v1.py
shawn-hurley: no because that can generate invalid permutations right?

tests/test_web/test_api_v1.py
shawn-hurley: no, because that is a valid case now and should not fail right?

tests/test_web/test_api_v1.py
shawn-hurley: sorry not understanding what you want to add a test for

https://github.com/release-engineering/iib/pull/141/files#diff-abace6c6541ec4270d1bab50b005634eR369
and 
https://github.com/release-engineering/iib/pull/141/files#diff-abace6c6541ec4270d1bab50b005634eR369

Cover the cases when no bundles and:
1. from_index not set
2. binary_image not set

tests/test_web/test_api_v1.py
lcarva: This is for the `test_rm_operators_invalid_params_format` test and the error message is about the `operators` parameter. How is that related to these changes?

tests/test_web/test_api_v1.py
lcarva: (Those two links you sent refer to the same line, which is the one right above my comment)

This use case you added covers when `bundles` is not provided, but only `binary_image` is provided. Good.

We also need:

1. `bundles` not provided and only `from_index` is provided. But the kicker is that, even before these changes, if `binary_image` is not provided, then `add_arches` is required. So this test case should be: `from_index` and `add_arches` is provided and nothing else. This should get past the error "Missing required parameter(s): binary_image" and raise instead ""from_index" and "binary_image" must be specified if no bundles are specified'
1. None of `bundles`, `from_index`, `binary_image` are provided. For the same reason as the previous item, this would require the `add_arches` parameter to ensure the right error is being triggered.

tests/test_web/test_api_v1.py
lcarva: That's true. Ignore this.

PR #142, opened 2020-08-19
Adding dogpile.cache for caching skopeo_inspect

lipoja:
- added dogpile.cachce configuration
- added caching for skopeo_inspect method
- updated all requirements files and setup.py
- added unit test for testing skopeo_inspect cache

For more information refer to [CLOUDDST-116]


midnightercz (changes requested):

iib/workers/tasks/utils.py
```diff
@@ -147,7 +154,7 @@ def set_registry_token(token, container_image):
 
 
 @retry(wait_on=IIBError, logger=log)
-def skopeo_inspect(*args, return_json=True):
+def skopeo_inspect_no_cache(*args, return_json=True):
```
yashvardhannanavati, line 157: Why not simply keep this `skopeo_inspect` and have a boolean `use_cache` defaulting to `False`. You can probably use CacheRegion.get_or_create() if `use_cache` is `True`. I believe that would be a cleaner implementation. Thoughts?

iib/workers/dogpile_cache.py
```diff
@@ -0,0 +1,54 @@
...
+    def get_dogpile_namespace(self):
+        """Get dogpile namespace."""
+        return self.dogpile_namespace
+
+    def set_dogpile_namespace(self, ns):
+        """Set dogpile namespace."""
+        self.dogpile_namespace = str(ns)
+
+
+dp_ns = dict()
```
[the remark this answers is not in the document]
lipoja: Yeah, I think the code snippet could work. I will try that tomorrow.

iib/workers/tasks/utils.py
lipoja: OK, I used just the single function and specify cache by parameter. Default to using cache.

iib/workers/dogpile_cache.py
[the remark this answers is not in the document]
lipoja: I tried to implement that suggestion. However I am somehow not fully convinced about the solution. Could you please have a look on it?

iib/workers/tasks/utils.py
```diff
@@ -1,27 +1,30 @@
...
 import base64
-from contextlib import contextmanager
 import functools
 import inspect
 import json
 import logging
 import os
 import re
 import subprocess
+from contextlib import contextmanager
```
midnightercz, line 10: why you moved it here?

midnightercz (approved):

iib/workers/tasks/utils.py
lipoja: I did not have to, I used "Optimize imports" from PyCharm. On the other hand I think it should belong to that place.  Because I am used to have first imports starting with `import`, then `from ... import ...`. And also try to preserve order suggested in [PEP8](https://www.python.org/dev/peps/pep-0008/#imports): 
```
Standard library imports.
Related third party imports.
Local application/library specific imports.
```

midnightercz: yes and to follow that order
from contextlib import contextmanager 
should be first

lipoja: It is order like that, separate to groups just `import foo` comes before `from foo import bar`. Which, at least for me, has it logical meaning. You have another section of imports grouped and visually separated. It is better for reading the code. 

However I can revert that change if it is not according our code style.

iib/workers/tasks/utils.py
lipoja: Code was updated and now it uses just one function.

iib/workers/tasks/utils.py
```diff
@@ -163,11 +172,21 @@ def skopeo_inspect(*args, return_json=True):
...
+        output_cache = dogpile_cache.get(cache_key)
+        if output_cache:
+            return output_cache
+
     output = run_cmd(cmd, exc_msg=exc_msg)
     if return_json:
-        return json.loads(output)
+        output = json.loads(output)
+
+    dogpile_cache.set(cache_key, output)
```
querti, line 189: Is there a JSON/non-JSON control implemented? It seems like the cache is set to whatever type was last used. Doesn't that mean that the next call might return the wrong type?

iib/workers/tasks/utils.py
```diff
@@ -147,12 +155,13 @@ def set_registry_token(token, container_image):
...
 
 @retry(wait_on=IIBError, logger=log)
-def skopeo_inspect(*args, return_json=True):
+def skopeo_inspect(*args, return_json=True, use_cache=True):
     """
     Wrap the ``skopeo inspect`` command.
 
     :param args: any arguments to pass to ``skopeo inspect``
     :param bool return_json: if ``True``, the output will be parsed as JSON and returned
+    :param bool use_cache: Set True to get result form cache, False to disable cachce
```
querti, line 164: typo

iib/workers/config.py
```diff
@@ -26,6 +26,11 @@ class Config(object):
     )
     iib_request_logs_level = 'DEBUG'
     iib_required_labels = {}
+    # Configuration for dogpile.cache
+    # To disable dogpile.cache you can use 'dogpile.cache.null' backend
+    iib_dogpile_backend = 'dogpile.cache.memcached'
```
querti, line 31: I think these should be documented.

iib/workers/tasks/build.py
```diff
@@ -295,7 +295,7 @@ def _get_resolved_bundles(bundles):
     log.info('Resolving bundles %s', ', '.join(bundles))
     resolved_bundles = set()
     for bundle_pull_spec in bundles:
-        skopeo_raw = skopeo_inspect(f'docker://{bundle_pull_spec}', '--raw')
+        skopeo_raw = skopeo_inspect(f'docker://{bundle_pull_spec}', '--raw', use_cache=True)
```
querti, line 298: This is default value, no?

querti (changes requested):

iib/workers/tasks/utils.py
lipoja: The key is generated based on arguments (arg, kwarg) You can check [generate_cache_key](https://github.com/release-engineering/iib/pull/142/files#diff-e93fe72aa218e5b071f12554fcb09d85R11) and it is called on line 175: `cache_key = generate_cache_key(skopeo_inspect.__name__, *args, return_json=return_json)` 
So the key is generated different for both values (True/False)

iib/workers/tasks/utils.py
lipoja: Fixed

iib/workers/config.py
lipoja: Good point, thanks.

iib/workers/tasks/build.py
lipoja: Yes.

querti (approved):

README.md
```diff
@@ -218,6 +218,9 @@ The custom configuration options for the Celery workers are listed below:
   Additionally, it will use this file as a base and set the `overwrite_from_index_token` for the
   registry of the `from_index` container image when applicable. IIB will never directly modify this
   file though. This defaults to `~/.docker/config.json.template`.
+*  `iib_dogpile_backend` - configuration for dogpile.cache backend. Default value is
```
yashvardhannanavati, line 221: Could you please document `iib_dogpile_expiration_time` and `iib_dogpile_arguments` in the readme too?

iib/workers/dogpile_cache.py
```diff
@@ -0,0 +1,72 @@
...
+from celery import current_task
+from dogpile.cache import make_region
+
+from iib.workers.config import get_worker_config
+
+
+def generate_cache_key(fname, *arg, **kw):
+    """Generate key that is used in dogpile cache."""
+    namespace = dogpile_namespace.get_dogpile_namespace()
+    return namespace + "_" + fname + "_".join([str(s) for s in arg] + [f"{k}={kw[k]}" for k in kw])
```
yashvardhannanavati, line 14: It would be nice to use meaningful variable names and not use single letters :)

yashvardhannanavati (changes requested):

iib/workers/dogpile_cache.py
```diff
@@ -0,0 +1,72 @@
...
+import os
+import threading
+
+from celery import current_task
+from dogpile.cache import make_region
+
+from iib.workers.config import get_worker_config
+
+
+def generate_cache_key(fname, *arg, **kw):
```
yashvardhannanavati, line 11: Why not have this as a staticmethod inside the class?

iib/workers/dogpile_cache.py
```diff
@@ -0,0 +1,72 @@
...
+    """Generate key that is used in dogpile cache."""
+    namespace = dogpile_namespace.get_dogpile_namespace()
+    return namespace + "_" + fname + "_".join([str(s) for s in arg] + [f"{k}={kw[k]}" for k in kw])
+
+
+def create_dogpile_region():
```
yashvardhannanavati, line 17: Same here.

README.md
lipoja: Sure, thanks

iib/workers/dogpile_cache.py
lipoja: I would not suggest to putting these inside the class that is declared below (DogpileNamespace), because these functions are not related to namespace.
I am not sure what will be the benefit to putting those functions to class? When we do not have to store any context.

yashvardhannanavati: Correct, I think you're right. We have two options here, either we could name the class as DogpileCache and move these as @staticmethod or keep it as it is. Your call. :)

yashvardhannanavati:
@lipoja...@mprahl comes back from PTO tomorrow, once he has glanced this, we can merge it. He is more familiar with dogpile cache than I am.

iib/workers/dogpile_cache.py
lipoja: I am voting for keeping it this way unless other reviewers disagree. 
Thank you.

iib/workers/dogpile_cache.py
```diff
@@ -0,0 +1,75 @@
...
+            celery_id = current_task.request.id
+        except AttributeError:
+            celery_id = ''
+        if current_task.request.id not in self.ns[pid][thread_id]:
+            self.ns[pid][thread_id][celery_id] = ''
+
+        self.ns[pid][thread_id][celery_id] = str(build_id)
+
+
+dogpile_namespace = DogpileNamespace()
```
yashvardhannanavati, line 75: Could you please define this at the top of the file?

yashvardhannanavati:
Also, could you please make sure the cache is cleared before ending every request?

README.md
```diff
@@ -218,6 +218,11 @@ The custom configuration options for the Celery workers are listed below:
...
+   `'dogpile.cache.memcached'`. In case you want to disable caching
```
mprahl, line 222: This looks like an incomplete sentence.

README.md
```diff
@@ -218,6 +218,11 @@ The custom configuration options for the Celery workers are listed below:
   Additionally, it will use this file as a base and set the `overwrite_from_index_token` for the
   registry of the `from_index` container image when applicable. IIB will never directly modify this
   file though. This defaults to `~/.docker/config.json.template`.
+*  `iib_dogpile_backend` - configuration for dogpile.cache backend. Default value is
+   `'dogpile.cache.memcached'`. In case you want to disable caching 
+*  `iib_dogpile_expiration_time` - Number of seconds after which cached item is expired.
```
mprahl, line 223:
```suggestion
*  `iib_dogpile_expiration_time` - The number of seconds after which the cached item is expired.
```

README.md
```diff
@@ -218,6 +218,11 @@ The custom configuration options for the Celery workers are listed below:
...
+   `iib_dogpile_arguments` - Additional arguments for dogpile backed.
```
mprahl, line 224:
```suggestion
   `iib_dogpile_arguments` - Additional arguments for dogpile backend.
```

docker/Dockerfile-workers
```diff
@@ -7,6 +7,7 @@ RUN dnf -y install \
     git \
     golang \
     make \
+    memcached \
```
mprahl, line 10: This looks to be in the wrong space in the Dockerfile since this is a multistage build. Try putting it in the second stage of the build.

iib/workers/config.py
mprahl, line 31: I would prefer that the default be to have caching disabled. This would make this an opt-in feature by the person deploying IIB.

iib/workers/config.py
```diff
@@ -81,6 +86,7 @@ class TestingConfig(DevelopmentConfig):
     iib_greenwave_url = 'some_url'
     iib_omps_url = 'some_url'
     iib_request_logs_dir = None
+    iib_dogpile_expiration_time = 5
```
mprahl, line 89: Don't we want caching disabled in testing by default?

iib/workers/dogpile_cache.py
mprahl: I don't think it's possible since the class is defined in the file and Python source is read starting at the top.

iib/workers/tasks/utils.py
```diff
@@ -147,12 +155,13 @@ def set_registry_token(token, container_image):
 
 
 @retry(wait_on=IIBError, logger=log)
-def skopeo_inspect(*args, return_json=True):
+def skopeo_inspect(*args, return_json=True, use_cache=True):
```
mprahl, line 158: I would prefer if you used [cache_on_arguments](https://dogpilecache.sqlalchemy.org/en/latest/api.html#dogpile.cache.region.CacheRegion.cache_on_arguments) and then passed in function for the `should_cache_fn` keyword argument that will only return `True` if the pull spec passed into `skopeo_inspect` uses a digest. Thent he caller does not have to think about if using the cache is necessary.

iib/workers/dogpile_cache.py
```diff
@@ -0,0 +1,75 @@
...
+    conf = get_worker_config()
+
+    return make_region().configure(
+        conf.iib_dogpile_backend,
+        expiration_time=conf.iib_dogpile_expiration_time,
+        arguments=conf.iib_dogpile_arguments,
+    )
+
+
+class DogpileNamespace:
```
mprahl, line 31: Could you please explain why a namespace is required? Based on my understanding, if we use `cache_on_arguments` as I describe in another comment, it will use [get_or_create](https://dogpilecache.sqlalchemy.org/en/latest/api.html#dogpile.cache.region.CacheRegion.get_or_create). The documentation for `get_or_create` says the following:
> Whether or not the function is used depends on if the dogpile lock can be acquired or not. If it can’t, it means a different thread or process is already running a creation function for this key against the cache. When the dogpile lock cannot be acquired, the method will block if no previous value is available, until the lock is released and a new value available. If a previous value is available, that value is returned immediately without blocking.

Since locking is in place automatically and we only want to cache skopeo inspect calls that use a digest for now, it should be fine to use the default behavior.

iib/workers/dogpile_cache.py
```diff
@@ -0,0 +1,75 @@
...
+import os
+import threading
+
+from celery import current_task
+from dogpile.cache import make_region
+
+from iib.workers.config import get_worker_config
+
+
+def generate_cache_key(fname, *args, **kwargs):
```
mprahl, line 11: If you use my suggestion of using `cache_on_arguments`, I believe the default function to generate the cache key will work and this custom function would not be needed.

iib/workers/dogpile_cache.py
mprahl: To further clarify, I think sharing the same cache for all requests is fine since we are only caching something that is sure to be a stable value for the duration of `conf.iib_dogpile_expiration_time`.

iib/workers/dogpile_cache.py
lipoja: Yeah, Matt is right.

iib/workers/config.py
lipoja: There is a test for caching, so I guess no?

iib/workers/config.py
lipoja: yes, this might be good approach, thanks.

iib/workers/config.py
mprahl: In this case, I would enable caching just for the test that tests caching. I imagine having caching enabled for all tests could produce unexpected results based on things being mocked.

iib/workers/tasks/utils.py
lipoja: Hi @mprahl , I could not do it so I am not using `cache_on_argument`. Since `should_cache_fn` does not have data about pull spec input. There is passed only the result of `skopeo_inspect` function. I am caching only data when the image was specified with digest.

Could you please check the code if it is OK, or if you know about different solution please suggest one. 
Thank you!

iib/workers/dogpile_cache.py
lipoja: Thank you, I changed the code, could you have a look please?

README.md
```diff
@@ -230,6 +230,11 @@ The custom configuration options for the Celery workers are listed below:
   Additionally, it will use this file as a base and set the `overwrite_from_index_token` for the
   registry of the `from_index` container image when applicable. IIB will never directly modify this
   file though. This defaults to `~/.docker/config.json.template`.
+*  `iib_dogpile_backend` - configuration for dogpile.cache backend. Default value is
+   `'dogpile.cache.memcached'`. In case you want to disable caching use `'dogpile.cache.null'`.
```
mprahl, line 234: The default here needs to be changed to the new default of 'dogpile.cache.null'`.

iib/workers/tasks/build.py
```diff
@@ -33,7 +33,6 @@
     skopeo_inspect,
 )
 
-
```
mprahl, line 36: This file has a bunch of whitespace lines removed that have nothing to do with the PR. Could you please remove this deletion from the commit?

iib/workers/tasks/build.py
```diff
@@ -9,14 +9,14 @@
 import tempfile
 import textwrap
 
-from operator_manifest.operator import ImageName, OperatorManifest
 import ruamel.yaml
+from operator_manifest.operator import ImageName, OperatorManifest
```
mprahl, line 13: This change is not relevant

iib/workers/tasks/build.py
```diff
@@ -9,14 +9,14 @@
...
 
 from iib.exceptions import IIBError, AddressAlreadyInUse
 from iib.workers.api_utils import set_request_state, update_request
 from iib.workers.config import get_worker_config
-from iib.workers.tasks.celery import app
 from iib.workers.greenwave import gate_bundles
+from iib.workers.tasks.celery import app
```
mprahl, line 19: This change is not relevant

iib/workers/tasks/utils.py
mprahl: If you want to change the order of imports, please make a separate PR.

iib/workers/tasks/utils.py
```diff
@@ -165,9 +168,19 @@ def skopeo_inspect(*args, return_json=True):
...
+        output_cache = dogpile_cache.get(cache_key)
+        if output_cache:
+            return output_cache
+
     output = run_cmd(cmd, exc_msg=exc_msg)
     if return_json:
-        return json.loads(output)
+        output = json.loads(output)
+
+    dogpile_cache.set(cache_key, output)
```
mprahl, line 183: Wouldn't you not want to cache it if `should_cache` is `False`?

iib/workers/tasks/utils.py
mprahl: I see what you mean. `should_cache_fn` does something different than I thought. Perhaps a decorator that does this would be cleaner. The decorator could accept a function which analyzes the arguments to determine if the cache should be used.

mprahl:
I gave it a quick review but I don't have time for an in-depth review at the moment. Sorry.

README.md
lipoja: Thank you, updated.

iib/workers/tasks/utils.py
lipoja: Thanks for this suggestion, I've created a decorator so the `skopeo_inspect` function remains untouched. 
Could you check it, please.

iib/workers/dogpile_cache.py
```diff
@@ -0,0 +1,63 @@
...
+
+from iib.workers.config import get_worker_config
+
+
+def should_use_cache(*args):
+    """Return true in case this requests can be taken from or stored in cache."""
+    return any([arg.find('@sha256:') != -1 for arg in args])
+
+
+def skopeo_cache(dogpile_region):
```
mprahl, line 14: I think you could make this a generic decorator which accepts a parameter for a function that analyzes the arguments and determines if the cache should be used.

iib/workers/dogpile_cache.py
```diff
@@ -0,0 +1,63 @@
+# SPDX-License-Identifier: GPL-3.0-or-later
+import functools
+
+from dogpile.cache import make_region
+
+from iib.workers.config import get_worker_config
+
+
+def should_use_cache(*args):
```
mprahl, line 9: Could you make this a clearer name?
```suggestion
def skopeo_inspect_should_use_cache(*args):
```

tox.ini
```diff
@@ -59,12 +59,20 @@ commands =
 [testenv:py36]
 description = Python 3.6 unit tests [Mandatory]
 commands =
-    {[testenv]pytest_command}
+    {[testenv]pytest_command} --ignore-glob='*test_utils_cache.py'
+
+[testenv:py36-cache]
+description = Python 3.6 unit tests for dogpile cache
+setenv =
+	IIB_TESTING=true
```
mprahl, line 67:
```suggestion
    IIB_TESTING=true
```

README.md
```diff
@@ -230,6 +230,11 @@ The custom configuration options for the Celery workers are listed below:
   Additionally, it will use this file as a base and set the `overwrite_from_index_token` for the
   registry of the `from_index` container image when applicable. IIB will never directly modify this
   file though. This defaults to `~/.docker/config.json.template`.
+*  `iib_dogpile_backend` - configuration for dogpile.cache backend. Default value is
+   `'dogpile.cache.null'`. In case you want to enable caching use `'dogpile.cache.memcached'`.
```
mprahl, line 234:
```suggestion
*  `iib_dogpile_backend` - The configuration for the dogpile.cache backend. The default value is
   `'dogpile.cache.null'`. In case you want to enable caching, set this to `'dogpile.cache.memcached'`.
```

README.md
```diff
@@ -230,6 +230,11 @@ The custom configuration options for the Celery workers are listed below:
...
+*  `iib_dogpile_expiration_time` - The number of seconds after which the cached item is expired.
+   `iib_dogpile_arguments` - Additional arguments for dogpile backend.
```
mprahl, line 236:
```suggestion
   `iib_dogpile_arguments` - Additional arguments for the dogpile backend.
```

README.md
```diff
@@ -230,6 +230,11 @@ The custom configuration options for the Celery workers are listed below:
...
+   (e.g. for testing), you can set this variable to `'dogpile.cache.null'`.
```
mprahl, line 237:
```suggestion
```

iib/workers/dogpile_cache.py
```diff
@@ -0,0 +1,63 @@
...
+def generate_cache_key(fn, *args, **kwargs):
+    """Generate key that is used in dogpile cache."""
+    arguments = "|".join(
+        [str(arg) for arg in args] + [f"{kwarg}={kwargs[kwarg]}" for kwarg in kwargs]
+    )
+    return f"{fn}|{arguments}"
+
+
+def create_dogpile_region():
+    """Load setting from config and creates dogpile region."""
```
mprahl, line 56:
```suggestion
    """Create and configure a dogpile region."""
```

tests/test_workers/test_tasks/test_utils_cache.py
```diff
@@ -0,0 +1,60 @@
...
+def test_should_cache(value, result):
+    assert should_use_cache(value) == result
+
+
+@pytest.mark.parametrize(
+    'wait_time, called', [(0, True), (0, False), (TestingConfig.iib_dogpile_expiration_time, True)]
+)
+@mock.patch('iib.workers.tasks.utils.run_cmd')
+def test_skopeo_inspect_cache(mock_run_cmd, wait_time, called):
+    # wait 0 or dogpile_expiration_time before running skopeo_inspect
```
mprahl, line 27:
```suggestion
    # Wait for the dogpile_expiration_time before running skopeo_inspect
```

tests/test_workers/test_tasks/test_utils_cache.py
```diff
@@ -0,0 +1,60 @@
...
+    if called:
+        skopeo_args = mock_run_cmd.call_args[0][0]
+        args_expected = ['skopeo', '--command-timeout', '300s', 'inspect', image]
+        assert skopeo_args == args_expected
+    else:
+        # We are expecting to hit cached result
+        assert mock_run_cmd.call_args is None
+
+
+@pytest.mark.parametrize('called', [True, True, True])
```
mprahl, line 47: This doesn't need to be a parameter since it's always the same value.

mprahl:
This is looking really good. I left a few comments for improvements.

iib/workers/dogpile_cache.py
lipoja: sure, thanks its better this way

iib/workers/dogpile_cache.py
lipoja: Hi @mprahl, I changed the decorator to be more general. Could you check it if it is implemented in a way you suggested. 
Thank you, Jan

iib/workers/dogpile_cache.py
```diff
@@ -0,0 +1,64 @@
+# SPDX-License-Identifier: GPL-3.0-or-later
+import functools
+
+from dogpile.cache import make_region
+
+from iib.workers.config import get_worker_config
+
+
+def skopeo_inspect_should_use_cache(*args):
```
mprahl, line 9: I think it'd be good to consistently have the `should_use_cache` functions to accepts all arguments that were passed in the function call even though it doesn't matter in the case of `skopeo_inspect`.
```suggestion
def skopeo_inspect_should_use_cache(*args, **kwargs):
```

README.md
```diff
@@ -232,6 +232,10 @@ The custom configuration options for the Celery workers are listed below:
   Additionally, it will use this file as a base and set the `overwrite_from_index_token` for the
   registry of the `from_index` container image when applicable. IIB will never directly modify this
   file though. This defaults to `~/.docker/config.json.template`.
+*  `iib_dogpile_backend` -  The configuration for the dogpile.cache backend. The default value is
```
mprahl, line 235: To be consistent with the rest of the document.
```suggestion
*  `iib_dogpile_backend` -  the configuration for the dogpile.cache backend. The default value is
```

README.md
```diff
@@ -232,6 +232,10 @@ The custom configuration options for the Celery workers are listed below:
...
+   `'dogpile.cache.null'`. In case you want to enable caching, set this to `'dogpile.cache.memcached'`.
+*  `iib_dogpile_expiration_time` - The number of seconds after which the cached item is expired.
```
mprahl, line 237: To be consistent with the rest of the document.
```suggestion
*  `iib_dogpile_expiration_time` - the number of seconds after which the cached item is expired.
```

README.md
```diff
@@ -232,6 +232,10 @@ The custom configuration options for the Celery workers are listed below:
...
+   `iib_dogpile_arguments` - Additional arguments for the dogpile backend.
```
mprahl, line 238: To be consistent with the rest of the document.
```suggestion
   `iib_dogpile_arguments` - additional arguments for the dogpile backend.
```

iib/workers/config.py
```diff
@@ -30,6 +30,12 @@ class Config(object):
     )
     iib_request_logs_level = 'DEBUG'
     iib_required_labels = {}
+    # Configuration for dogpile.cache
+    # Disabled by default (by using 'dogpile.cache.null').
+    # To enable caching set 'dogpile.cache.memcached' as backend.
+    iib_dogpile_backend = 'dogpile.cache.null'
+    iib_dogpile_expiration_time = 5
```
mprahl, line 37: This seems like a very short default and would not last the whole request. Do you want to set this to something like 10 minutes by default?

iib/workers/dogpile_cache.py
```diff
@@ -0,0 +1,64 @@
...
+    """Return true in case this requests can be taken from or stored in cache."""
+    return any([arg.find('@sha256:') != -1 for arg in args])
+
+
+def dogpile_cache(dogpile_region, should_use_cache_fn):
+    """
+    Dogpile cache decorator.
+
+    Ensures that cache is used only when pull spec contains digest.
```
mprahl, line 18:
```suggestion
```

tests/test_workers/test_tasks/test_utils_cache.py
```diff
@@ -0,0 +1,61 @@
...
+
+@pytest.mark.parametrize(
+    "value, result",
+    [
+        ('docker://with_digest@sha256:93120347593478509347tdsvzkljbn', True),
+        ('docker://without_digest:tag', False),
+    ],
+)
+def test_should_cache(value, result):
+    assert skopeo_inspect_should_use_cache(value) == result
```
mprahl, line 19:
```suggestion
    assert skopeo_inspect_should_use_cache(value) is result
```

tests/test_workers/test_tasks/test_utils_cache.py
```diff
@@ -0,0 +1,61 @@
...
+
+
+@pytest.mark.parametrize(
```
mprahl, line 22: The use of parametrize implies that each test case is independent, however, they all rely on each other. So all those tests have to be run serially one after the other.

The other issue is that it is susceptible to race conditions, though highly unlikely as you are assuming the second test is executed before the cache expires.

iib/workers/config.py
```diff
@@ -89,6 +95,9 @@ class TestingConfig(DevelopmentConfig):
     iib_greenwave_url = 'some_url'
     iib_omps_url = 'some_url'
     iib_request_logs_dir = None
+    if os.getenv('IIB_ENABLE_CACHE', '').lower() == 'true':
+        iib_dogpile_backend = 'dogpile.cache.memcached'
```
mprahl, line 99: I don't like how I have to install and run memcached to run some of the unit tests. I think we can just assume that dogpile and the memcached library work and you can just mock out the interactions with dogpile for testing the caching decorator.

iib/workers/config.py
lipoja: Oh noes, this one was committed by mistake. 
Thanks!

iib/workers/config.py
lipoja: @mprahl Yeah, dogpile is now mocked. Thanks for the suggestion.

iib/workers/dogpile_cache.py
```diff
@@ -0,0 +1,62 @@
...
+    Dogpile cache decorator.
+
+    :params dogpile_region: Dogpile CacheRegion object
+    :params should_use_cache_fn: function which determines if cache should be used
+    """
+
+    def cache_decorator(func):
+        @functools.wraps(func)
+        def inner(*args, **kwargs):
+            should_cache = should_use_cache_fn(*args)
```
mprahl, line 25:
```suggestion
            should_cache = should_use_cache_fn(*args, **kwargs)
```

docker/Dockerfile-workers
```diff
@@ -59,6 +59,7 @@ RUN dnf -y install \
     python3-devel \
     python3-pip \
     skopeo \
+    memcached \
```
mprahl, line 62: This should be removed. A separate slimmed down container image can be used for running memcached in production.

iib/workers/config.py
```diff
@@ -30,6 +30,12 @@ class Config(object):
     )
     iib_request_logs_level = 'DEBUG'
     iib_required_labels = {}
+    # Configuration for dogpile.cache
+    # Disabled by default (by using 'dogpile.cache.null').
+    # To enable caching set 'dogpile.cache.memcached' as backend.
+    iib_dogpile_backend = 'dogpile.cache.null'
+    iib_dogpile_expiration_time = 600
+    iib_dogpile_arguments = {'url': ["127.0.0.1"]}
```
mprahl, line 38:
```suggestion
    iib_dogpile_arguments = {'url': ['127.0.0.1']}
```

iib/workers/dogpile_cache.py
```diff
@@ -0,0 +1,62 @@
...
+import functools
+
+from dogpile.cache import make_region
+
+from iib.workers.config import get_worker_config
+
+
+def skopeo_inspect_should_use_cache(*args, **kwargs):
+    """Return true in case this requests can be taken from or stored in cache."""
+    return any([arg.find('@sha256:') != -1 for arg in args])
```
mprahl, line 11: It'd be more efficient to use a generator rather than creating a list. Then you don't have to evaluate all the args.
```suggestion
    return any(arg.find('@sha256:') != -1 for arg in args)
```

setup.py
```diff
@@ -10,12 +10,14 @@
...
     install_requires=[
         'celery',
+        'dogpile.cache',
         'flask',
         'flask-login',
         'flask-migrate',
         'flask-sqlalchemy',
         'operator-manifest',
         'psycopg2-binary',
+        'python-memcached',
```
mprahl, line 20: This is not a requirement since the default dogpile backend is null so it shouldn't be installed/required by default. That being said, it'd be nice to have it preinstalled in the workers container image. Perhaps [extras_require](https://setuptools.readthedocs.io/en/latest/userguide/dependency_management.html#optional-dependencies) might be the best option.

Thoughts @lcarva and @yashvardhannanavati?

tests/test_workers/test_tasks/test_utils_cache.py
```diff
@@ -0,0 +1,49 @@
+from unittest import mock
+
+import pytest
+
+from iib.workers.dogpile_cache import skopeo_inspect_should_use_cache
+from iib.workers.tasks import utils
+
+
+@pytest.mark.parametrize(
+    "value, result",
```
mprahl, line 10:
```suggestion
    'value, result',
```

tests/test_workers/test_tasks/test_utils_cache.py
```diff
@@ -0,0 +1,49 @@
...
+def test_should_cache(value, result):
+    assert skopeo_inspect_should_use_cache(value) is result
+
+
+@mock.patch('dogpile.cache.region.CacheRegion.get')
+@mock.patch('iib.workers.tasks.utils.run_cmd')
+def test_skopeo_inspect_cache(mock_run_cmd, moc_dpr_get):
+    mock_run_cmd.return_value = '{"Name": "some-image-cache"}'
+    image = 'docker://some-image-cache@sha256:129bfb6af3e03997eb_not_real_sha_c7c18d89b40d97'
+    rv_expected = {"Name": "some-image-cache"}
```
mprahl, line 25:
```suggestion
    rv_expected = {'Name': 'some-image-cache'}
```

tox.ini
```diff
@@ -59,12 +59,12 @@ commands =
 [testenv:py36]
 description = Python 3.6 unit tests [Mandatory]
 commands =
-    {[testenv]pytest_command}
+    {[testenv]pytest_command} --ignore-glob='*test_utils_cache.py'
```
mprahl, line 62: Can you remove this change?

tox.ini
```diff
@@ -59,12 +59,12 @@ commands =
...
 
 [testenv:py37]
 description = Python 3.7 unit tests [Mandatory]
 commands =
-    {[testenv]pytest_command}
+    {[testenv]pytest_command} --ignore-glob='*test_utils_cache.py'
```
mprahl, line 67: Can you remove this change?

tox.ini
lipoja: :facepalm: thanks

iib/workers/dogpile_cache.py
lipoja: Agree

docker/Dockerfile-workers
lipoja: @mprahl I am sorry I am not sure what you are suggesting here. Remove it completely so there is not memcached install in dev env? Or should I add memcached container to docker-compose? 
Not sure if you mean production - setting in our gitlab or if you meant this dev. env.

mprahl: @lipoja yeah, I'm thinking we don't want to install memcached in the same container as the worker. It'd be good to have a separate image for docker-compose though for testing. I like that idea, though it's not necessary.

setup.py
lcarva: How would that work with the requirements.txt file? We'd also need another one of those?

mprahl: Yeah... it looks like pip-compile does not supports extras_require:
https://github.com/jazzband/pip-tools/blob/d2007dab6b0cd0c33860fcbf14da793fe34705da/piptools/scripts/compile.py#L379-L398

We could have it in `extras_require` and then install it via Ansible. I think it's too much hassle and we can move forward with just making it a dependency.

mprahl:
After https://github.com/release-engineering/iib/pull/142#discussion_r532869127 is addressed, this is good to merge in my opinion.

setup.py
```diff
@@ -10,6 +10,7 @@
     zip_safe=False,
     install_requires=[
         'celery',
+        'dogpile.cache',
```
mprahl, line 13: Based on our discussion, we'll also need to add python-memcached as a requirement here.

docker-compose.yml
```diff
@@ -101,6 +104,7 @@ services:
     depends_on:
       - rabbitmq
       - registry
+      - memcached
```
mprahl, line 107: Can you configure the `DevelopmentConfig` class to connect to this memached instance but override it in `TestConfiguration` back to `dogpile.cache.null`?

docker-compose.yml
```diff
@@ -25,6 +25,9 @@ services:
       POSTGRES_DB: iib
       POSTGRES_INITDB_ARGS: "--auth='ident' --auth='trust'"
 
+  memcached:
+    image: memcached
```
mprahl, line 29: Can you define the ports?

setup.py
yashvardhannanavati: Agreed. +1 on making it a dependency :+1:

docker-compose.yml
lipoja: Sure, sorry this was incomplete merge just to test few things up. 
I am on it, thanks.

README.md
```diff
@@ -232,6 +232,10 @@ The custom configuration options for the Celery workers are listed below:
   Additionally, it will use this file as a base and set the `overwrite_from_index_token` for the
   registry of the `from_index` container image when applicable. IIB will never directly modify this
   file though. This defaults to `~/.docker/config.json.template`.
+*  `iib_dogpile_backend` -  the configuration for the dogpile.cache backend. The default value is
```
mprahl, line 235:
```suggestion
*  `iib_dogpile_backend` - the configuration for the dogpile.cache backend. The default value is
```

mprahl (approved):
This looks great. Nice job!

README.md
lipoja: Updated, thanks.

lipoja:
@mprahl Thank you for your time spent on code review.

iib/workers/dogpile_cache.py
```diff
@@ -0,0 +1,62 @@
...
+
+        return inner
+
+    return cache_decorator
+
+
+def generate_cache_key(fn, *args, **kwargs):
+    """Generate key that is used in dogpile cache."""
+    arguments = "|".join(
+        [str(arg) for arg in args] + [f"{kwarg}={kwargs[kwarg]}" for kwarg in kwargs]
```
yashvardhannanavati, line 49: I apologize for knit-picking but can we please use single quotes here? :)

iib/workers/dogpile_cache.py
```diff
@@ -0,0 +1,62 @@
...
+    )
+    return f"{fn}|{arguments}"
```
yashvardhannanavati, line 51: same here

yashvardhannanavati (approved):
@lipoja This looks amazing! Thank you for all the work! :100: :+1:

iib/workers/dogpile_cache.py
lipoja: oh, I am just wondering why we have single quotes as standard... 
it is just not natural for me :D

lipoja: updated, thanks :)

PR #144, opened 2020-08-25
Return the OMPS-pushed version when IIB backports a bundle

lipoja:
- log response from OMPS API for every successful request
- collecting operator version for all packages
- updated patch_request function to process request with omps_operator_version
- adding propagating of OMPS response to _get_base_dir_and_pkg_name
- adding omps_operator_version to RequestAdd database table model
- added set_omps_operator_version; set the set_omps_operator_version of the request using the IIB API.
- added call of set_omps_operator_version after collecting all versions

For more information refer to [CLOUDDST-2201]

### DIFF
--- a/iib/web/models.py
+++ b/iib/web/models.py
@@ -788,6 +788,8 @@ class RequestAdd(Request, RequestIndexImageMixin):
     bundles = db.relationship('Image', secondary=RequestAddBundle.__table__)
     organization = db.Column(db.String, nullable=True)
 
+    omps_operator_version = db.Column(db.String, nullable=True)
+
     __mapper_args__ = {
         'polymorphic_identity': RequestTypeMapping.__members__['add'].value,
     }
@@ -860,6 +862,9 @@ class RequestAdd(Request, RequestIndexImageMixin):
         rv = super().to_json(verbose=verbose)
         rv.update(self.get_common_index_image_json())
         rv['organization'] = self.organization
+        rv['omps_operator_version'] = {}
+        if self.omps_operator_version:
+            rv['omps_operator_version'] = json.loads(self.omps_operator_version)
 
         for bundle in self.bundles:
             if bundle.operator:
@@ -879,7 +884,7 @@ class RequestAdd(Request, RequestIndexImageMixin):
         """
         rv = super().get_mutable_keys()
         rv.update(self.get_index_image_mutable_keys())
-        rv.update({'bundles', 'bundle_mapping'})
+        rv.update({'bundles', 'bundle_mapping', 'omps_operator_version'})
         return rv
 
 
--- a/iib/workers/tasks/legacy.py
+++ b/iib/workers/tasks/legacy.py
@@ -10,7 +10,7 @@ import requests
 import ruamel.yaml
 
 from iib.exceptions import IIBError
-from iib.workers.api_utils import set_request_state
+from iib.workers.api_utils import set_request_state, set_omps_operator_version
 from iib.workers.config import get_worker_config
 from iib.workers.tasks.utils import get_image_labels, retry, run_cmd
 
@@ -30,14 +30,17 @@ def export_legacy_packages(packages, request_id, rebuilt_index_image, cnr_token,
         backported packages should be pushed to.
     :raises IIBError: if the export of packages fails.
     """
+    operator_versions = dict()
     with tempfile.TemporaryDirectory(prefix='iib-') as temp_dir:
         for package in packages:
             _opm_index_export(rebuilt_index_image, package, temp_dir)
             package_dir = os.path.join(temp_dir, package)
             _verify_package_info(package_dir, rebuilt_index_image)
             _zip_package(package_dir)
-            _push_package_manifest(package_dir, cnr_token, organization)
+            omps_response = _push_package_manifest(package_dir, cnr_token, organization)
+            operator_versions[package] = omps_response.get('version')
 
+    set_omps_operator_version(request_id, operator_versions)
     set_request_state(request_id, 'in_progress', 'Back ported packages successfully pushed to OMPS')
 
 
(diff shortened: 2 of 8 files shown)


lipoja:
Hello @yashvardhannanavati @mprahl @lcarva @querti @midnightercz @chandwanitulsi can I ask you for code review?

iib/workers/tasks/legacy.py
```diff
@@ -141,6 +146,9 @@ def _push_package_manifest(package_dir, cnr_token, organization):
             raise IIBError(
                 f'Push to {organization} in the legacy app registry was unsucessful: {error_msg}'
             )
+        else:
```
yashvardhannanavati, line 149: I don't think you need this `else`.

tests/test_workers/test_tasks/test_legacy.py
```diff
@@ -127,17 +127,19 @@ def test_opm_index_export(mock_run_cmd):
...
+def test_export_legacy_packages(mock_oie, mock_srs, mock_soov, mock_ppm, mock_zp, mock_vpi):
+    packages = {'prometheus'}
     legacy.export_legacy_packages(packages, 3, 'from:index', 'token', 'org')
 
     mock_oie.assert_called_once()
     mock_vpi.assert_called_once()
     mock_zp.assert_called_once()
     mock_ppm.assert_called_once()
     mock_srs.assert_called_once()
+    mock_soov.assert_called_once()
```
yashvardhannanavati, line 142: optional: you could probably use `assert_called_once_with` here.

iib/web/models.py
```diff
@@ -837,7 +839,12 @@ def from_json(cls, kwargs, batch=None):
 
         cls._from_json(
             request_kwargs,
-            additional_optional_params=['from_index', 'organization', 'bundles'],
+            additional_optional_params=[
+                'from_index',
+                'organization',
+                'bundles',
+                'omps_operator_version',
```
yashvardhannanavati, line 846: Why are we allowing the user to specify `omps_operator_version` when we don't use it while calling OMPS? Am I missing something?

yashvardhannanavati (changes requested):

querti (commented):
Was is agreed upon how CVP will consume this data? Their merge request mentions reading state history, but this PR expects it to read the logs or querying for task metadata.

lipoja:
> Was is agreed upon how CVP will consume this data? Their merge request mentions reading state history, but this PR expects it to read the logs or querying for task metadata.

Yes I talked with Krunoslav from CVP and he is fine this implementation.

querti (approved):

midnightercz (approved):

iib/web/models.py
```diff
@@ -860,6 +862,10 @@ def to_json(self, verbose=True):
         rv = super().to_json(verbose=verbose)
         rv.update(self.get_common_index_image_json())
         rv['organization'] = self.organization
+        if self.omps_operator_version:
```
yashvardhannanavati, line 865: You can initialize  `rv['omps_operator_version'] = {}` and then check the `if`. That way, you can remove the else block, no?

yashvardhannanavati:
Just one comment. This is looking good. Let's merge it once it's fixed. Thanks Jan!

iib/web/models.py
lipoja: yes, however I am used to do it this way, because it will save you some processing time (allocating memory and creating empty dictionary) in case the `if` is true.
```python
rv['omps_operator_version'] = {}  # <--- This is done every time
if self.omps_operator_version:
    rv['omps_operator_version'] = json.loads(self.omps_operator_version)
```
vs.

```python
if self.omps_operator_version:
    rv['omps_operator_version'] = json.loads(self.omps_operator_version)
else:
    rv['omps_operator_version'] = {}  # <--- This is done only in case omps_operator_version is False
```

yashvardhannanavati: Yup, but it's a battle between readability and negligible processing time, in which according to me readability wins. But I'll leave the final call to you

iib/workers/tasks/legacy.py
```diff
@@ -120,7 +123,9 @@ def _push_package_manifest(package_dir, cnr_token, organization):
         app registry via OMPS.
     :param str organization: the organization name in the legacy app registry to which
          the backported packages should be pushed to.
-    :raises IIBError: if the push is unsucessful
+    :return: Response from OMPS in JSON format
+    :rtype: dict|None
+    :raises IIBError: if the push is unsuccessful
```
yashvardhannanavati, line 128: Optional: `if the push is unsuccessful` => `if the push fails`

tests/test_web/test_api_v1.py
```diff
@@ -504,6 +505,7 @@ def test_add_bundle_success(
             'url': 'http://localhost/api/v1/builds/1/logs',
             'expiration': '2020-02-15T17:03:00Z',
         },
+        'omps_operator_version': {},
```
yashvardhannanavati, line 508: May I ask why is the org not None but still the `omps_operator_version` missing? Could you please add the test where both of them are not None?

iib/workers/tasks/legacy.py
lipoja: I guess your suggestion is better

tests/test_web/test_api_v1.py
lipoja: I guess here is tested just the first response (and not waiting for finals state). `omps_operator_version` is filled after OMPS is API called and I guess this is not the test for it.

The test for this is covered in our integration test (iib-api-tests).

yashvardhannanavati: You can probably mock the response of the OMPS call and return a mock value, no?

yashvardhannanavati (approved):

lipoja:
re-based and squashed all commits requested by @yashvardhannanavati

PR #145, opened 2020-09-03
Use the version of OPM based on the ocp version label of the index

shawn-hurley:
* opm now must be installed to opmv4.5 and opmv4.6
* added tests to make sure that OCP_VERSION label was respected

### DIFF
--- a/iib/workers/tasks/build.py
+++ b/iib/workers/tasks/build.py
@@ -339,7 +339,7 @@ def _get_resolved_image(pull_spec):
 
 @retry(attempts=2, wait_on=IIBError, logger=log)
 def _opm_index_add(
-    base_dir, bundles, binary_image, from_index=None, overwrite_from_index_token=None
+    base_dir, bundles, binary_image, ocp_version, from_index=None, overwrite_from_index_token=None
 ):
     """
     Add the input bundles to an operator index.
@@ -351,6 +351,8 @@ def _opm_index_add(
         add to the index image being built.
     :param str binary_image: the pull specification of the container image where the opm binary
         gets copied from. This should point to a digest or stable tag.
+    :param str ocp_version: the version that is specified on the from index. Will choose which
+        version of OPM to run.
     :param str from_index: the pull specification of the container image containing the index that
         the index image build will be based from.
     :param str overwrite_from_index_token: the token used for overwriting the input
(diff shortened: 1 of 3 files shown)


lcarva:
Should we have a mechanism for setting up a default binary? Maybe that's not necessary. 

Let's have the corresponding changes in the [dev environment|https://github.com/release-engineering/iib/blob/master/docker/Dockerfile-workers] though. Otherwise, it'll break.

iib/workers/tasks/build.py
```diff
@@ -339,7 +339,7 @@ def _get_resolved_image(pull_spec):
 
 @retry(attempts=2, wait_on=IIBError, logger=log)
 def _opm_index_add(
-    base_dir, bundles, binary_image, from_index=None, overwrite_from_index_token=None
+    base_dir, bundles, binary_image, ocp_version, from_index=None, overwrite_from_index_token=None
```
lipoja, line 342: Could you add docstring for this parameter? 
Thank you!

iib/workers/tasks/build.py
```diff
@@ -386,7 +386,9 @@ def _opm_index_add(
 
 
 @retry(attempts=2, wait_on=IIBError, logger=log)
-def _opm_index_rm(base_dir, operators, binary_image, from_index, overwrite_from_index_token=None):
+def _opm_index_rm(
+    base_dir, operators, binary_image, from_index, ocp_version, overwrite_from_index_token=None
```
lipoja, line 390: Could you add docstring for this parameter as well?
Thank you!

iib/workers/tasks/build.py
```diff
@@ -386,7 +386,9 @@ def _opm_index_add(
 
 
 @retry(attempts=2, wait_on=IIBError, logger=log)
-def _opm_index_rm(base_dir, operators, binary_image, from_index, overwrite_from_index_token=None):
+def _opm_index_rm(
```
lipoja, line 389: Could you please update a call of this function (on line 870). Otherwise a wrong value is passed to the `ocp_version` parameter.
Thank you.

lipoja (changes requested):

querti:
May I ask the reason for needing both versions? Do they have different functionality?

yashvardhannanavati (approved):

querti (changes requested):
With new merged changes, it also has to be changed in the registry serve command

shawn-hurley:
Closing this PR for now. I believe that we may want to come back and do this but we should not add this right now.

PR #146, opened 2020-09-04
Add merge index image endpoint

querti:
- The stages were implemented according to the design doc
- '_prepare_request_for_build' function has been updated to
  accomodate the new request type.
- Deprecation based on the image labels was implemented as well,
  although it will currently not be used.
- Stages can be skipped if it's discovered that they are a no-op.
- Intermediate index images have to be uploaded to the registry as
  opm cannot work with local images.
- One thing which hasn't been implemented is the propagation of
  default channel deletion error. The reason is that the error
  format has not yet been agreed upon.
- Added new endpoint, along with new relational models specification,
  updated data validation, database migration, and documentation.
- Pull specifications of the source and target index images are
  updated during the course of the task.

Unfortunately, this PR contains a commit from #137. It requires functions introduced by that PR, which hasn't yet been merged. I will rebase the new changes on top of master once #137 is merged.

I should mention that I haven't successfully tested overwrite_from_index, because quay always gives error "manifest invalid: manifest invalid". Has anyone encountered this?


querti:
Data I used for testing:
`curl -u: --negotiate -X POST -H "Content-Type: application/json"  http://localhost:8080/api/v1/builds/merge-index-image -d \
    '{
       "source_from_index": "quay.io/lgallovi/iib-test:4.5.10",
       "target_from_index": "quay.io/lgallovi/iib-test:4.6.4",
       "deprecation_list": ["registry-proxy.engineering.redhat.com/rh-osbs/quay-quay-container-security-operator-metadata@sha256:5c089eb32a2ba03e8901f4e4f55b5b0724b4629ab79aa71ba392b02c3c8c3b52"],
       "binary_image": "registry-proxy.engineering.redhat.com/rh-osbs/openshift-ose-operator-registry:v4.6"
     }'`

querti:
@yashvardhannanavati @mprahl @lcarva @lipoja @midnightercz @chandwanitulsi review pls

querti (commented):
After #145 is merged, I have to incorporate the changes

iib/workers/tasks/build.py
```diff
@@ -648,6 +648,8 @@ def _prepare_request_for_build(
     overwrite_from_index_token=None,
     add_arches=None,
     bundles=None,
+    source_from_index=None,
+    target_from_index=None,
```
yashvardhannanavati, line 652: Could you please change target_from_index to target_index everywhere?

iib/workers/tasks/build.py
```diff
@@ -669,7 +671,11 @@ def _prepare_request_for_build(
     :param list add_arches: the list of arches to build in addition to the arches ``from_index`` is
         currently built for; if ``from_index`` is ``None``, then this is used as the list of arches
         to build the index image for
-    :param list bundles: the list of bundles to create the bundle mapping on the request
+    :param list bundles: the list of bundles to create the bundle mapping on the request.
+    :param str source_from_index: the pull specification of the container image containing the index
+        that will be used as a base of the merged index image.
+    :param str target_from_index: the pull specification of the container image containing the index
```
yashvardhannanavati, line 677: This doesn't seem right, the source_from_index data will be added here and a few more. Could you please reword this?

iib/workers/tasks/build.py
```diff
@@ -700,6 +706,30 @@ def _prepare_request_for_build(
...
+                get_image_label(source_from_index_resolved, 'com.redhat.index.delivery.version')
+                or 'v4.5'
+            )
+    else:
+        source_from_index_resolved = None
+        source_ocp_version = 'v4.5'
+
+    if target_from_index:
+        with set_registry_token(overwrite_from_index_token, target_from_index):
+            target_from_index_resolved = _get_resolved_image(target_from_index)
```
yashvardhannanavati, line 722: This code seems to be duplicated in the same function. How about creating a helper method?

iib/workers/tasks/build.py
```diff
@@ -756,6 +790,14 @@ def _update_index_image_build_state(request_id, prebuild_info):
     if from_index_resolved:
         payload['from_index_resolved'] = from_index_resolved
 
+    source_from_index_resolved = prebuild_info.get('source_from_index_resolved')
+    if source_from_index_resolved:
+        payload['source_from_index_resolved'] = source_from_index_resolved
+
+    target_from_index_resolved = prebuild_info.get('target_from_index_resolved')
```
yashvardhannanavati, line 797: Same, could you please rename `target_from_index_resolved` to `target_index_resolved`?

yashvardhannanavati (changes requested):

iib/web/models.py
```diff
@@ -1049,6 +1071,99 @@ def get_mutable_keys(self):
...
 
+class RequestMergeIndexImage(Request, RequestIndexImageMixin):
+    """An "merge-index-image" build request."""
+
+    __tablename__ = 'request_merge_index_image'
+
+    id = db.Column(db.Integer, db.ForeignKey('request.id'), autoincrement=False, primary_key=True)
+    source_from_index_id = db.Column(db.Integer, db.ForeignKey('image.id'), nullable=True)
+    source_from_index = db.relationship('Image', foreign_keys=[source_from_index_id], uselist=False)
+    target_from_index_id = db.Column(db.Integer, db.ForeignKey('image.id'), nullable=True)
```
yashvardhannanavati, line 1082: Could you please rename target_from_index to target_index here too?

iib/web/models.py
```diff
@@ -1049,6 +1071,99 @@ def get_mutable_keys(self):
...
+    deprecation_list = db.relationship('Image', secondary=BundleDeprecation.__table__)
+
+    __mapper_args__ = {
+        'polymorphic_identity': RequestTypeMapping.__members__['merge_index_image'].value,
+    }
+
+    @classmethod
+    def from_json(cls, kwargs, batch=None):
+        """
+        Handle JSON requests for the Add API endpoint.
```
yashvardhannanavati, line 1093: for the merge-index-image endpoint?

iib/web/models.py
```diff
@@ -1049,6 +1071,99 @@ def get_mutable_keys(self):
...
+
+        :param dict kwargs: the JSON payload of the request.
+        :param Batch batch: the batch to specify with the request.
+        """
+        request_kwargs = deepcopy(kwargs)
+
+        bundles = request_kwargs.pop('deprecation_list', [])
+        if not isinstance(bundles, list) or any(
+            not item or not isinstance(item, str) or '@' not in item for item in bundles
```
yashvardhannanavati, line 1102: Why do you need the bundles to be specified with their hash? Doesn't IIB already resolve them to their hases? You could reuse that method

iib/workers/tasks/build.py
```diff
@@ -337,6 +342,156 @@ def _get_resolved_image(pull_spec):
...
+    binding conflicts. Resolution of port conflicts is handled in this function as well.
+
+    :param str db_path: path to index database containing the registry data.
+    :return: tuple containing port number of the running service and the running Popen object.
+    :rtype: (int, Popen)
+    :raises IIBError: if all tried ports are in use, or the command failed for another reason.
+    """
+    conf = get_worker_config()
+    port = conf['iib_grpc_start_port']
+    for i in range(conf['iib_grpc_max_port_tries']):
```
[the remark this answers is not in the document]
querti: This should be discussed in #137. Although I see where you're coming from, the code is too intertwined to be easily separable.

iib/workers/tasks/utils.py
```diff
@@ -280,3 +282,32 @@ def _get_function_arg_value(arg_name, func, args, kwargs):
...
+            queue.put(line)
+        out.close()
+
+    lines = []
+    q = Queue()
+    t = Thread(target=enqueue_output, args=(stream, q))
+    t.daemon = True
+    t.start()
+
+    while True:
```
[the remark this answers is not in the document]
querti: This should be discussed in #137. Reading output from a running process is blocking, and since there's no deterministic way to know when to stop reading, the read would end up hanging forever. I've actually tried many different solutions for non-blocking reading, and this was the only one that worked reliably.

iib/workers/tasks/build.py
```diff
@@ -978,6 +1440,139 @@ def handle_regenerate_bundle_request(from_bundle_image, organization, request_id
...
+        if deprecate_bundles:
+            _deprecate_bundles(
+                deprecate_bundles,
+                temp_dir,
+                binary_image,
+                intermediate_image_name,
+                prebuild_info['target_ocp_version'],
+                overwrite_from_index_token,
+            )
+            # _build_image(temp_dir, 'index.Dockerfile', request_id, arch)
```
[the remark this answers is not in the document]
querti: This is a part of the code that will be uncommented later. Right now, it's not necessary to rebuild the intermediate index image, so it's commented.

iib/workers/tasks/build.py
```diff
@@ -978,6 +1440,139 @@ def handle_regenerate_bundle_request(from_bundle_image, organization, request_id
...
+            # _push_image(request_id, arch)
+            # _create_and_push_manifest_list(request_id, [arch])
+
+        # Deprecating based on label should not be used for now
```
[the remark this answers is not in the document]
querti: I made the message a bit more detailed

iib/workers/tasks/build.py
querti: All occurrences should be changed

iib/workers/tasks/build.py
querti: Hopefully it's more accurate now.

iib/workers/tasks/build.py
querti: Hmm it's only two lines that are duplicated everywhere. And they are both variable value setters. Does it really make sense to create a helper function?

iib/workers/tasks/build.py
querti: done

iib/web/models.py
querti: done

iib/web/models.py
querti: fixed.

iib/web/models.py
querti: I think I saw the requirement for the bundles to be specified by a digest in the design doc. "It’ll be a yaml file with package name as the key and the bundle pull_spec list as value specified with digest. " I guess the data validation doesn't have to be done on IIB's side, but should instead be handled by the script. Should I changed it to accept tags as well?

querti:
@midnightercz @yashvardhannanavati @lcarva @mprahl @chandwanitulsi @lipoja check pls

yashvardhannanavati:
@querti Could you please rebase this PR?

iib/web/api_v1.py
```diff
@@ -624,3 +626,48 @@ def add_rm_batch():
...
+    args = [
+        payload['binary_image'],
+        payload['source_from_index'],
+        payload['target_index'],
+        payload.get('deprecation_list'),
+        request.id,
+        overwrite_from_index,
+        payload.get('overwrite_from_index_token'),
+    ]
+    safe_args = copy.copy(args)
```
yashvardhannanavati, line 660: Why not simply call _get_safe_args here?

iib/web/models.py
```diff
@@ -1054,6 +1076,97 @@ def get_mutable_keys(self):
...
+        ):
+            raise ValidationError(
+                f'"deprecation_list" should be an array of strings. '
```
yashvardhannanavati, line 1110: f string is not required here

iib/web/models.py
yashvardhannanavati: I think all you need to check here is that bundles is an array of strings. Let the bundle resolution in the code take care of valid/invalid bundles.

iib/web/models.py
```diff
@@ -1054,6 +1076,97 @@ def get_mutable_keys(self):
...
+    def from_json(cls, kwargs, batch=None):
+        """
+        Handle JSON requests for the merge-index-image API endpoint.
+
+        :param dict kwargs: the JSON payload of the request.
+        :param Batch batch: the batch to specify with the request.
+        """
+        request_kwargs = deepcopy(kwargs)
+
+        bundles = request_kwargs.pop('deprecation_list', [])
```
yashvardhannanavati, line 1105: Let's use `deprecation_list` as the variable name instead of bundles to avoid confusion.

yashvardhannanavati (changes requested):

iib/workers/tasks/build.py
```diff
@@ -876,6 +918,248 @@ def get_image_label(pull_spec, label):
...
+        operators_bundles_asoc[bundle['channelName']].append(bundle)
+
+    channel_heads = []
+    for bundles in operators_bundles_asoc.values():
+        # comparing versions as strings seems the safest
+        channel_heads.append(sorted(bundles, key=lambda x: x['version'], reverse=True)[0])
+
+    bundles_with_metadata = []
+    for head_bundle in channel_heads:
+        skopeo_out = skopeo_inspect('docker://{0}'.format(head_bundle['bundlePath']))
```
yashvardhannanavati, line 1121: `skopeo_inspect` is an expensive operation. Since you're simply using a single label, could you use the get_image_label method to get it? (It uses the --config option which is extremely fast). Check out https://github.com/release-engineering/iib/commit/f58767585bfedbbb4d771c9f238d23d27a1aeaf6 for reference. If you're using it in any other place, could you please replace those too?

iib/workers/tasks/build.py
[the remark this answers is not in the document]
yashvardhannanavati: I'm not a fan of keeping non-functional commented code. Could you please remove it and store it in your branch or a snippet?

yashvardhannanavati (changes requested):

iib/web/api_v1.py
querti: Fixed

iib/web/models.py
querti: Fixed

iib/web/models.py
querti: The reason why I added it was to check if each bundle is specified via digest, not to verify bundle validity.

I can still remove it if you'd like.

iib/web/models.py
querti: Fixed

iib/web/models.py
```diff
@@ -1054,6 +1076,97 @@ def get_mutable_keys(self):
         return rv
 
 
+class RequestMergeIndexImage(Request, RequestIndexImageMixin):
+    """An "merge-index-image" build request."""
```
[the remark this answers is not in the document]
querti: Fixed

iib/web/static/api_v1.yaml
```diff
@@ -709,6 +748,48 @@ components:
...
+            - 'quay.io/user/openshift-bundle-example-container@sha256:29e4531e4625653'
+        source_from_index:
+          type: string
+          description: >
+            Index image that will be used as a base for the new index image
+          example: 'quay.io/iib-stage/iib:4.5'
+        target_index:
+          type: string
+          description: >
+            Target stage index image for the corresponding target index image tag
```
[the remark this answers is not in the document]
querti: Fixed

iib/workers/tasks/build.py
```diff
@@ -705,6 +711,30 @@ def _prepare_request_for_build(
...
+        source_from_index_resolved = None
+        source_ocp_version = 'v4.5'
+
+    if target_index:
+        with set_registry_token(overwrite_from_index_token, target_index):
+            target_index_resolved = _get_resolved_image(target_index)
+            target_index_arches = _get_image_arches(target_index_resolved)
+            target_ocp_version = (
+                get_image_label(target_index_resolved, 'com.redhat.index.delivery.version')
+                or 'v4.5'
```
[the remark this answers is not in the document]
querti: Changed it for now, but will wait for confirmation

iib/workers/tasks/build.py
```diff
@@ -705,6 +711,30 @@ def _prepare_request_for_build(
         from_index_resolved = None
         ocp_version = 'v4.5'
 
+    if source_from_index:
```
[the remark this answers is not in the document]
querti: I added a comment to the API docs. I guess the other option would be to use two tokens, but that seems a bit too contrived and doesn't seem like our use-case.

iib/workers/tasks/build.py
```diff
@@ -876,6 +918,248 @@ def get_image_label(pull_spec, label):
...
+    :param str base_dir: base directory where operation files will be located.
+    :param int request_id: the ID of the IIB build request.
+    :param str arch: the architecture to build this image for.
+    :param str overwrite_from_index_token: the token used for overwriting the input
+        ``from_index`` image. This is required for non-privileged users to use
+        ``overwrite_from_index``. The format of the token must be in the format "user:password".
+    """
+    set_request_state(request_id, 'in_progress', 'Creating intermediate index image')
+    log.info('Creating intermediate index image with a new binary image')
+    cmd = [
```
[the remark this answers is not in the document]
querti: You're right. Changed.

iib/workers/tasks/build.py
```diff
@@ -876,6 +918,248 @@ def get_image_label(pull_spec, label):
     return get_image_labels(pull_spec).get(label)
 
 
+def _create_intermediate_index_image(
```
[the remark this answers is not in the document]
querti: Changed.

iib/workers/tasks/build.py
```diff
@@ -876,6 +918,248 @@ def get_image_label(pull_spec, label):
...
+    with set_registry_token(overwrite_from_index_token, from_index):
+        run_cmd(
+            cmd, {'cwd': base_dir}, exc_msg='Failed to generate new index image Dockerfile',
+        )
+
+    _add_ocp_label_to_index(ocp_version, base_dir, 'index.Dockerfile')
+    _build_image(base_dir, 'index.Dockerfile', request_id, arch)
+    _push_image(request_id, arch)
+    _create_and_push_manifest_list(request_id, [arch])
+    log.info('Intermediate image created')
```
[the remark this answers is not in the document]
querti: Added.

iib/workers/tasks/build.py
```diff
@@ -876,6 +918,248 @@ def get_image_label(pull_spec, label):
...
+    ]
+    target_bundle_ids = [
+        (bundle['packageName'], bundle['version'], bundle) for bundle in target_index_bundles
+    ]
+
+    for package_name, version, bundle in target_bundle_ids:
+        if (package_name, version) not in source_bundle_ids:
+            missing_bundles.append(bundle)
+
+    if missing_bundles != []:
```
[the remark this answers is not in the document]
querti: Fixed.

iib/workers/tasks/build.py
```diff
@@ -876,6 +918,248 @@ def get_image_label(pull_spec, label):
...
+    :param int request_id: the ID of the IIB build request.
+    :param str arch: the architecture to build this image for.
+    :param str ocp_version: ocp version which will be added as a label to the image.
+    :return: bundles which were added to the index image.
+    :rtype: list
+    """
+    set_request_state(request_id, 'in_progress', 'Adding bundles missing in source index image')
+    log.info('Adding bundles from target index image which are missing from source index image')
+    missing_bundles = []
+    source_bundle_ids = [
```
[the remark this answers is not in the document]
querti: Good catch. Fixed.

iib/workers/tasks/build.py
```diff
@@ -876,6 +918,248 @@ def get_image_label(pull_spec, label):
...
+        if resolved_bundle in deprecation_list:
+            deprecate_bundles.append(resolved_bundle)
+
+    log.info(
+        'Bundles that will be deprecated from the index image: %s', ','.join(deprecate_bundles)
+    )
+    return deprecate_bundles
+
+
+def _get_deprecated_bundles_by_ocp_version(bundles_with_metadata, ocp_version):
```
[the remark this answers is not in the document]
querti: Sorry, I don't think I understand. I thought that currently this type of deprecation should not be used at all. '_get_deprecated_bundles_by_ocp_version' is not currently used anywhere in the code.

iib/workers/tasks/build.py
querti: Fixed.

iib/workers/tasks/build_merge_index_image.py
```diff
@@ -0,0 +1,366 @@
...
+    :param str overwrite_from_index_token: the token used for overwriting the input
+        ``from_index`` image. This is required for non-privileged users to use
+        ``overwrite_from_index``. The format of the token must be in the format "user:password".
+    """
+    set_request_state(request_id, 'in_progress', 'Creating intermediate index image')
+    log.info('Creating intermediate index image with a new binary image')
+    _opm_index_add(base_dir, [], binary_image, from_index, overwrite_from_index_token)
+
+    _add_ocp_label_to_index(ocp_version, base_dir, 'index.Dockerfile')
+    _build_image(base_dir, 'index.Dockerfile', request_id, arch)
```
yashvardhannanavati, line 58: Why not call build_image here for every arch present? Check out handle_add_request for reference.

iib/workers/tasks/build_merge_index_image.py
```diff
@@ -0,0 +1,366 @@
...
+        _create_intermediate_index_image(
+            binary_image,
+            source_from_index,
+            prebuild_info['target_ocp_version'],
+            temp_dir,
+            request_id,
+            arch,
+        )
+
+        missing_bundles = _add_bundles_missing_in_source(
```
yashvardhannanavati, line 317: Instead of calling opm_index_add twice in this function and in _create_intermediate_index_image, what's stopping us from finding the missing bundles first and then simply calling opm_index_add once along with the missing bundles to form the intermediate index image. Then simply call build_image once for each arch on the index image. Wouldn't this approach reduce complexity?

yashvardhannanavati (changes requested):

iib/workers/tasks/build_merge_index_image.py
querti: I thought that it would be more efficient to built only one arch for the intermediate stages. All archs are build for the final index image. Was that a bad thinking on my part?

iib/workers/tasks/build_merge_index_image.py
querti: Yeah, I noticed that the process could be optimized, but decided to implement it as it was described in the design doc in case I was missing some information.

I have simplified the code according to your suggestion.

midnightercz (changes requested):

iib/workers/tasks/build_merge_index_image.py
```diff
@@ -0,0 +1,337 @@
...
+                deprecate_bundles.append(bundle)
+        # multiple csv versions
+        elif ',' in bundle_ocp_version:
+            str_versions = [
+                version.split('.') for version in bundle_ocp_version.replace('v', '').split(',')
+            ]
+            versions = []
+            for version in str_versions:
+                versions.append([int(num) for num in version])
+            min_version = sorted(versions)[-1]
```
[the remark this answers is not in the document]
querti: Whoops, you're right. Fixed and updated tests to catch it.

iib/workers/tasks/build.py
```diff
@@ -707,6 +713,30 @@ def _prepare_request_for_build(
...
+        source_from_index_resolved = None
+        source_ocp_version = 'v4.5'
+
+    if target_index:
+        with set_registry_token(overwrite_from_index_token, target_index):
+            target_index_resolved = _get_resolved_image(target_index)
+            target_index_arches = _get_image_arches(target_index_resolved)
+            target_ocp_version = (
+                get_image_label(target_index_resolved, 'com.redhat.index.delivery.version')
+                or 'v4.6'
```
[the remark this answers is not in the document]
querti: @yashvardhannanavati thoughts?

iib/workers/tasks/build_merge_index_image.py
```diff
@@ -0,0 +1,366 @@
...
+    :rtype: list
+    """
+    set_request_state(request_id, 'in_progress', 'Adding bundles missing in source index image')
+    log.info('Adding bundles from target index image which are missing from source index image')
+    missing_bundles = []
+    source_bundle_digests = []
+    target_bundle_digests = []
+
+    for bundle in source_index_bundles:
+        if '@sha256:' in bundle['bundlePath']:
```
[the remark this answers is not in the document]
querti: IIB Already performs a similar operation when checking for already present bundles, and it doesn't display any warnings.
I suppose we could, but it might become very spammy.

If such a scenario does happen, it would be pretty big trouble, as IIB won't be guaranteed to find all the missing bundles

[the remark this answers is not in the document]
querti: @yashvardhannanavati  Thoughts?

iib/workers/tasks/build.py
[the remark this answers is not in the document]
shawn-hurley: I think we probably want to point to n+1 of OCP. I agree making this configurable is desirable, I think we have two configuration stories that we probably want to make:

1. make ocp_version configurable (the label and the default value)
2. make distribution_scoep configurable (the label and the default value) 

at least that is my opinion

iib/workers/tasks/build.py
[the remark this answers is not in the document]
querti: I removed the label deprecation logic, as discussed in the meeting.

iib/workers/tasks/build_merge_index_image.py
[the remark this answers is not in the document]
querti: I added an error message if this happens.

iib/workers/tasks/build.py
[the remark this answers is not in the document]
querti: Hmm so what do I do here?

yashvardhannanavati:
Closing this in favor of #162

PR #149, opened 2020-09-11
Revert "Adding ability to update binary image of index image from the Add bundle command

lcarva:
This reverts commit d2f44775b899a31329d7a50da0ca562c9e0c1cca.

OPM does not like the parameter `--bundles ''`. It fails with:
```
time="2020-09-11T12:57:51-04:00" level=error msg="permissive mode disabled" bundles="['']" error="error resolving name : object required"
Error: error resolving name : object required
```

PR #150, opened 2020-09-13
Fixing adding empty list of bundles

shawn-hurley:
another fix for #149 depending on if we want to keep this API change or remove it based on the ability to remove with out a bundle being in the index.


yashvardhannanavati (approved):

lcarva (approved):
It works for me on the local dev environment

PR #160, opened 2020-09-19
Adding distribution scope parameter

shawn-hurley:
* Optional add and remove paratameter for distribution scope
* distribution scope must be one of dev, stage, or prod
* Added validation during preparing requests because of the needed skopeo inspect
  to get the label.

### DIFF
--- a/iib/web/api_v1.py
+++ b/iib/web/api_v1.py
@@ -38,38 +38,39 @@ from iib.workers.tasks.general import failed_request_callback
 api_v1 = flask.Blueprint('api_v1', __name__)
 
 
-def _get_rm_args(payload, request_id, overwrite_from_index):
+def _get_rm_args(payload, request, overwrite_from_index):
     """
     Generate arguments for remove request.
 
     :param dict payload: Payload from the remove request
-    :param int request_id: ID of request
+    :param Request request: request saved in the database
     :param bool overwrite_from_index: determines if the overwrite should be forced
     """
     return [
         payload['operators'],
         payload['binary_image'],
-        request_id,
+        request.id,
         payload['from_index'],
         payload.get('add_arches'),
         overwrite_from_index,
         payload.get('overwrite_from_index_token'),
+        request.distribution_scope,
     ]
 
 
-def _get_add_args(payload, request_id, overwrite_from_index, celery_queue):
+def _get_add_args(payload, request, overwrite_from_index, celery_queue):
     """
     Generate arguments for add request.
 
     :param dict payload: Payload from the remove request
-    :param int request_id: ID of request
+    :param Request request: request saved in the database
     :param bool overwrite_from_index: determines if the overwrite should be forced
     :param str celery_queue: name of celery queue
     """
     return [
         payload.get('bundles', []),
         payload['binary_image'],
-        request_id,
+        request.id,
         payload.get('from_index'),
         payload.get('add_arches'),
         payload.get('cnr_token'),
@@ -77,6 +78,7 @@ def _get_add_args(payload, request_id, overwrite_from_index, celery_queue):
         payload.get('force_backport'),
         overwrite_from_index,
         payload.get('overwrite_from_index_token'),
+        request.distribution_scope,
         flask.current_app.config['IIB_GREENWAVE_CONFIG'].get(celery_queue),
     ]
 
@@ -274,7 +276,7 @@ def add_bundles():
 
     overwrite_from_index = _should_force_overwrite() or payload.get('overwrite_from_index')
     celery_queue = _get_user_queue(serial=overwrite_from_index)
-    args = _get_add_args(payload, request.id, overwrite_from_index, celery_queue)
+    args = _get_add_args(payload, request, overwrite_from_index, celery_queue)
     safe_args = _get_safe_args(args, payload)
     error_callback = failed_request_callback.s(request.id)
 
@@ -423,7 +425,7 @@ def rm_operators():
 
     overwrite_from_index = _should_force_overwrite() or payload.get('overwrite_from_index')
 
-    args = _get_rm_args(payload, request.id, overwrite_from_index)
+    args = _get_rm_args(payload, request, overwrite_from_index)
     safe_args = _get_safe_args(args, payload)
 
     error_callback = failed_request_callback.s(request.id)
@@ -590,9 +592,9 @@ def add_rm_batch():
         )
         celery_queue = _get_user_queue(serial=overwrite_from_index)
         if isinstance(request, RequestAdd):
-            args = _get_add_args(build_request, request.id, overwrite_from_index, celery_queue)
+            args = _get_add_args(build_request, request, overwrite_from_index, celery_queue)
         elif isinstance(request, RequestRm):
-            args = _get_rm_args(build_request, request.id, overwrite_from_index)
+            args = _get_rm_args(build_request, request, overwrite_from_index)
 
         safe_args = _get_safe_args(args, build_request)
 
--- a/iib/web/migrations/versions/bc29053265ba_add_distribution_scope.py
+++ b/iib/web/migrations/versions/bc29053265ba_add_distribution_scope.py
@@ -1,0 +1,32 @@
+"""Add distribution scope.
+
+Revision ID: bc29053265ba
+Revises: 2ab3d4558cb6
+Create Date: 2020-09-20 02:26:45.531336
+
+"""
+from alembic import op
+import sqlalchemy as sa
+
+
+# revision identifiers, used by Alembic.
+revision = 'bc29053265ba'
+down_revision = '2ab3d4558cb6'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    with op.batch_alter_table('request_add', schema=None) as batch_op:
+        batch_op.add_column(sa.Column('distribution_scope', sa.String(), nullable=True))
+
+    with op.batch_alter_table('request_rm', schema=None) as batch_op:
+        batch_op.add_column(sa.Column('distribution_scope', sa.String(), nullable=True))
+
+
+def downgrade():
+    with op.batch_alter_table('request_rm', schema=None) as batch_op:
+        batch_op.drop_column('distribution_scope')
+
+    with op.batch_alter_table('request_add', schema=None) as batch_op:
+        batch_op.drop_column('distribution_scope')
--- a/iib/web/models.py
+++ b/iib/web/models.py
@@ -652,6 +652,11 @@ class RequestIndexImageMixin:
         """Return the relationship to the built index image."""
         return db.relationship('Image', foreign_keys=[cls.index_image_id], uselist=False)
 
+    @declared_attr
+    def distribution_scope(cls):
+        """Return the distribution_scope for the request."""
+        return db.Column(db.String, nullable=True)
+
     @staticmethod
     def _from_json(
         request_kwargs, additional_required_params=None, additional_optional_params=None, batch=None
@@ -672,6 +677,7 @@ class RequestIndexImageMixin:
             'add_arches',
             'overwrite_from_index',
             'overwrite_from_index_token',
+            'distribution_scope',
         } | set(additional_optional_params or [])
 
         validate_request_params(
@@ -697,6 +703,15 @@ class RequestIndexImageMixin:
                     'The "overwrite_from_index" parameter is required when'
                     ' the "overwrite_from_index_token" parameter is used'
                 )
+
+        distribution_scope = request_kwargs.pop('distribution_scope', None)
+        if distribution_scope:
+            distribution_scope = distribution_scope.lower()
+            if distribution_scope not in ['prod', 'stage', 'dev']:
+                raise ValidationError(
+                    'The "distribution_scope" value must be one of "dev", "stage", or "prod"'
+                )
+            request_kwargs['distribution_scope'] = distribution_scope
 
         # Verify the user is authorized to use overwrite_from_index
         # current_user.is_authenticated is only ever False when auth is disabled
@@ -762,6 +777,7 @@ class RequestIndexImageMixin:
             'index_image': getattr(self.index_image, 'pull_specification', None),
             'organization': None,
             'removed_operators': [],
+            'distribution_scope': getattr(self, 'distribution_scope', None),
         }
 
     def get_index_image_mutable_keys(self):
@@ -839,7 +855,12 @@ class RequestAdd(Request, RequestIndexImageMixin):
 
         cls._from_json(
             request_kwargs,
-            additional_optional_params=['from_index', 'organization', 'bundles'],
+            additional_optional_params=[
+                'from_index',
+                'organization',
+                'bundles',
+                'distribution_scope',
+            ],
             batch=batch,
         )
 
--- a/iib/web/static/api_v1.yaml
+++ b/iib/web/static/api_v1.yaml
@@ -581,6 +581,12 @@ components:
             index_image:
               type: string
               example: 'quay.io/iib-stage/iib:5'
+            distribution_scope:
+              description: >
+                The scope of distribution for the index created by the request.
+                This determined what level of protection the addition or removal had.
+              type: string
+              example: 'prod'
             organization:
               type: string
               example: openshift-operators
@@ -648,6 +654,12 @@ components:
             index images that IIB does not have access to. The format of the token is in the format
             "user:password".
           example: token
+          distribution_scope:
+            description: >
+                The scope of distribution for the index created by the request.
+                This will determine what level of protection the addition will have.
+            type: string
+            example: 'prod'
       required:
         - binary_image
         - bundles
@@ -686,6 +698,12 @@ components:
             non-privileged users to use overwrite_from_index. The format of the token must be in
             the format "user:password".
           example: token
+        distribution_scope:
+          description: >
+                The scope of distribution for the index created by the request.
+                This will determine what level of protection the removal will have.
+          type: string
+          example: 'prod'
       required:
         - binary_image
         - from_index
--- a/iib/workers/tasks/build.py
+++ b/iib/workers/tasks/build.py
@@ -655,6 +655,7 @@ def _prepare_request_for_build(
     overwrite_from_index_token=None,
     add_arches=None,
     bundles=None,
+    distribution_scope=None,
 ):
     """
     Prepare the request for the index image build.
@@ -677,8 +678,11 @@ def _prepare_request_for_build(
         currently built for; if ``from_index`` is ``None``, then this is used as the list of arches
         to build the index image for
     :param list bundles: the list of bundles to create the bundle mapping on the request
-    :return: a dictionary with the keys: arches, binary_image_resolved, from_index_resolved, and
-        ocp_version.
+    :param str distribution_scope: the scope for distribution of the index image, defaults to
+        ``None``.
+    :return: a dictionary with the keys: arches, binary_image_resolved, from_index_resolved,
+        bundle_mapping, ocp_version, and distribution_scope.
+    :rtype: dict
     :raises IIBError: if the container image resolution fails or the architectures couldn't be
         detected.
     """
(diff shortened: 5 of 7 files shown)


shawn-hurley:
@yashvardhannanavati @mprahl 

I would assign you both for reviews but I don't think I have the ability to.

iib/workers/tasks/build.py
```diff
@@ -655,6 +655,7 @@ def _prepare_request_for_build(
     overwrite_from_index_token=None,
     add_arches=None,
     bundles=None,
+    distribution_scope=None,
```
yashvardhannanavati, line 658: Could you please add the docstring for this?

yashvardhannanavati (changes requested):
Looks great Shawn! Just one minor change

shawn-hurley:
@yashvardhannanavati 

Updated!

iib/web/models.py
```diff
@@ -698,6 +704,15 @@ def _from_json(
                     ' the "overwrite_from_index_token" parameter is used'
                 )
 
+        distribution_scope = request_kwargs.pop('distribution_scope', None)
+        if distribution_scope:
```
querti, line 708: Shouldn't this be set to PROD by default?

iib/workers/tasks/build.py
```diff
@@ -1431,3 +1466,43 @@ def _add_ocp_label_to_index(ocp_version, temp_dir, dockerfile_name):
...
+    :param str resolved_distribution_scope: the distribution_scope that the index is for.
+    :param str distribution_scope: the distribution_scope that has been requested the index is for.
+    :return: the valid distribution_scope
+    :rtype: str
+    :raises IIBError: if the resolved_distribution_scope is of lesser scope then distribution_scope
+    """
+    resolved_distribution_scope = resolved_distribution_scope.upper()
+    distribution_scope = (distribution_scope or 'PROD').upper()
+
+    if (resolved_distribution_scope == 'STAGE' and distribution_scope == 'PROD') or (
```
querti, line 1500: If there are plans to add more distribution scopes in the future, this should probably be rewritten.

iib/web/api_v1.py
```diff
@@ -54,6 +54,7 @@ def _get_rm_args(payload, request_id, overwrite_from_index):
         payload.get('add_arches'),
         overwrite_from_index,
         payload.get('overwrite_from_index_token'),
+        payload.get('distribution_scope'),
```
querti, line 57: I think the API args should be updated

querti (changes requested):

iib/web/models.py
shawn-hurley: No IMO we want to keep the request what the user sent us right?

iib/web/api_v1.py
shawn-hurley: Where would I do that? I don't see a place to do this, it appears to all be auto-generated?

iib/workers/tasks/build.py
shawn-hurley: I think I agree, do we think this can be a follow-on?

iib/web/api_v1.py
querti: It's defined in iib/web/static/api_v1.yaml

iib/web/models.py
querti: Hmm maybe you're right, I'm not sure. @yashvardhannanavati what do you think?

iib/workers/tasks/build.py
querti: You mean change this separately from this PR? I guess we can leave it to later.

iib/web/api_v1.py
shawn-hurley: Done

iib/workers/tasks/build.py
```diff
@@ -891,6 +904,7 @@ def handle_add_request(
     force_backport=False,
     overwrite_from_index=False,
     overwrite_from_index_token=None,
+    distribution_scope=None,
```
yashvardhannanavati, line 907: Could you please add a doc string here too?

iib/workers/tasks/build.py
```diff
@@ -677,8 +678,10 @@ def _prepare_request_for_build(
         currently built for; if ``from_index`` is ``None``, then this is used as the list of arches
         to build the index image for
     :param list bundles: the list of bundles to create the bundle mapping on the request
-    :return: a dictionary with the keys: arches, binary_image_resolved, from_index_resolved, and
-        ocp_version.
+    :param str distribution_scope: the scope for distribution of the index image, defaults to
+        'PROD'.
+    :return: a dictionary with the keys: arches, binary_image_resolved, from_index_resolved,
+        bundle_mapping, ocp_version, and distribution_scope.
```
yashvardhannanavati, line 684: Could you please add `:rtype: dict` here?

iib/web/models.py
yashvardhannanavati: I agree with Shawn's point

yashvardhannanavati:
Also, could you please squash your commits into one?

shawn-hurley:
@yashvardhannanavati @mprahl done and updated

iib/web/migrations/versions/bc29053265ba_add_distribution_scope.py
```diff
@@ -0,0 +1,38 @@
...
+
+# revision identifiers, used by Alembic.
+revision = 'bc29053265ba'
+down_revision = '2ab3d4558cb6'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
```
yashvardhannanavati, line 20: Could you please remove these auto generated comments from this file?

iib/web/migrations/versions/bc29053265ba_add_distribution_scope.py
```diff
@@ -0,0 +1,38 @@
...
+        batch_op.add_column(sa.Column('distribution_scope', sa.String(), nullable=True))
+
+    with op.batch_alter_table('request_rm', schema=None) as batch_op:
+        batch_op.add_column(sa.Column('distribution_scope', sa.String(), nullable=True))
+
+    # ### end Alembic commands ###
+
+
+def downgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
```
yashvardhannanavati, line 31: Same here

iib/web/static/api_v1.yaml
```diff
@@ -581,6 +581,9 @@ components:
             index_image:
               type: string
               example: 'quay.io/iib-stage/iib:5'
+            distribution_scope:
```
yashvardhannanavati, line 584: Could you please add this to RequestAdd and RequestRemove?

iib/workers/tasks/build.py
```diff
@@ -702,17 +706,26 @@ def _prepare_request_for_build(
             ocp_version = (
                 get_image_label(from_index_resolved, 'com.redhat.index.delivery.version') or 'v4.5'
             )
+            resolved_distribution_scope = (
+                get_image_label(from_index_resolved, 'com.redhat.index.delivery.distribution_scope')
+                or 'prod'
```
mprahl, line 711: You're storing this as upper-case in the database but the default is lower-case here. I realize that `_validate_distribution_scope`  sets it to upper-case again, but this just seems weird and unnecessary.

iib/workers/tasks/build.py
```diff
@@ -1431,3 +1469,43 @@ def _add_ocp_label_to_index(ocp_version, temp_dir, dockerfile_name):
...
+    Validate distribution scope is allowed to be updated.
+
+    :param str resolved_distribution_scope: the distribution_scope that the index is for.
+    :param str distribution_scope: the distribution_scope that has been requested the index is for.
+    :return: the valid distribution_scope
+    :rtype: str
+    :raises IIBError: if the resolved_distribution_scope is of lesser scope then distribution_scope
+    """
+    resolved_distribution_scope = resolved_distribution_scope.upper()
+    distribution_scope = (distribution_scope or resolved_distribution_scope).upper()
```
mprahl, line 1501: You're already fixing the value once to store it in the database, why not just pass the value stored in the database to `handle_add_request` that is already fixed rather than fix it again on the worker? It'd keep it consistent and then you can just trust the parameters passed to `handle_add_request`.

iib/workers/tasks/build.py
```diff
@@ -1431,3 +1469,43 @@ def _add_ocp_label_to_index(ocp_version, temp_dir, dockerfile_name):
...
+
+    if (resolved_distribution_scope == 'STAGE' and distribution_scope == 'PROD') or (
+        resolved_distribution_scope == 'DEV'
+        and (distribution_scope == 'STAGE' or distribution_scope == 'PROD')
+    ):
+        raise IIBError(
+            f'can not set "distribution_scope" to {distribution_scope} because from index is'
+            ' already set to {resolved_distribution_scope}'
```
yashvardhannanavati, line 1509: This should be an f string too, no?

iib/workers/tasks/build.py
```diff
@@ -1431,3 +1469,43 @@ def _add_ocp_label_to_index(ocp_version, temp_dir, dockerfile_name):
...
+    :rtype: str
+    :raises IIBError: if the resolved_distribution_scope is of lesser scope then distribution_scope
+    """
+    resolved_distribution_scope = resolved_distribution_scope.upper()
+    distribution_scope = (distribution_scope or resolved_distribution_scope).upper()
+
+    if (resolved_distribution_scope == 'STAGE' and distribution_scope == 'PROD') or (
+        resolved_distribution_scope == 'DEV'
+        and (distribution_scope == 'STAGE' or distribution_scope == 'PROD')
+    ):
```
mprahl, line 1506: This is hard to read and it'll make the if statement even more complicated if you add a distribution scope later on. How about this?
```suggestion
    distribution_scopes = ["DEV", "STAGE", "PROD"]
    # Make sure the request isn't regressing the distribution scope
    if distribution_scopes.index(distribution_scope) < distribution_scopes.index(resolved_distrbution_scope):
```

iib/web/models.py
```diff
@@ -698,6 +704,15 @@ def _from_json(
...
+            distribution_scope = distribution_scope.upper()
```
mprahl, line 709: Optional: I'd prefer lower-case to match the rest of the REST API.

iib/web/models.py
```diff
@@ -698,6 +704,15 @@ def _from_json(
...
+            if distribution_scope not in ['PROD', 'STAGE', 'DEV']:
```
mprahl, line 710: Somewhat optional: Rather than hard-code these distribution scopes, I'd rather see this be configurable in the event more distribution scopes are added or renamed.

iib/workers/tasks/build.py
```diff
@@ -916,6 +931,8 @@ def handle_add_request(
     :param str overwrite_from_index_token: the token used for overwriting the input
         ``from_index`` image. This is required for non-privileged users to use
         ``overwrite_from_index``. The format of the token must be in the format "user:password".
+    :param str distribution_scope: the scope for distribution of the index image, defaults to
+        'PROD'.
```
mprahl, line 935: This doesn't seem to default to `PROD` based on `distribution_scope=None,` in the function signature.

iib/workers/tasks/build.py
```diff
@@ -1038,7 +1066,13 @@ def handle_rm_request(
...
     _cleanup()
     prebuild_info = _prepare_request_for_build(
-        binary_image, request_id, from_index, overwrite_from_index_token, add_arches
+        binary_image,
+        request_id,
+        from_index,
+        overwrite_from_index_token,
+        add_arches,
+        None,
+        distribution_scope,
```
mprahl, line 1075:
```suggestion
        distribution_scope=distribution_scope,
```

iib/workers/tasks/build.py
```diff
@@ -1431,3 +1469,43 @@ def _add_ocp_label_to_index(ocp_version, temp_dir, dockerfile_name):
         log.debug(
             'Added the following line to %s: %s', dockerfile_name, label,
         )
+
+
+def _add_distribution_scope_label_to_index(distribution_scope, temp_dir, dockerfile_name):
```
mprahl, line 1474: This is basically a copy-paste of `_add_ocp_label_to_index`. Why not make a generic function called `add_label_to_index` that accepts the label name and value?

iib/web/models.py
shawn-hurley: I think that we want to make the entire thing configurable in a follow on IMO. 

What I mean is the label key, potential values and this bit here too, we want to make a configuration value IMO.

WDYT of that?

iib/workers/tasks/build.py
shawn-hurley: Maybe I am being silly, but if the default of what the distribution_scope will be is prod even through the default for the parameter as none? or is that not the way to think about it?

shawn-hurley:
@mprahl  @yashvardhannanavati updated based on feedback

iib/web/models.py
mprahl: I agree with the approach but I'd like to see it in this MR if possible. I'll leave it up to @yashvardhannanavati to decide on that though.

iib/workers/tasks/build.py
mprahl: It's just that the way it was worded makes it sound like the `distribution_scope` keyword argument defaults to `'PROD'` even though it's code below that does. If you rephrase it, then it's useful to have this information.

iib/web/models.py
```diff
@@ -698,6 +704,15 @@ def _from_json(
                     ' the "overwrite_from_index_token" parameter is used'
                 )
 
+        distribution_scope = request_kwargs.pop('distribution_scope', None)
+        if distribution_scope:
+            distribution_scope = distribution_scope.lower()
+            if distribution_scope not in ['prod', 'stage', 'dev']:
+                raise ValidationError(
+                    'The "distribution_scope" parameter must be one of "dev", "stage", or "prod"'
```
mprahl, line 712: Optional: This sounds a little more intuitive when talking about JSON.
```suggestion
                    'The "distribution_scope" value must be one of "dev", "stage", or "prod"'
```

iib/workers/tasks/build.py
```diff
@@ -677,8 +678,11 @@ def _prepare_request_for_build(
         currently built for; if ``from_index`` is ``None``, then this is used as the list of arches
         to build the index image for
     :param list bundles: the list of bundles to create the bundle mapping on the request
-    :return: a dictionary with the keys: arches, binary_image_resolved, from_index_resolved, and
-        ocp_version.
+    :param str distribution_scope: the scope for distribution of the index image, defaults to
+        None.
```
mprahl, line 682:
```suggestion
        ``None``.
```

iib/workers/tasks/build.py
```diff
@@ -916,6 +931,8 @@ def handle_add_request(
     :param str overwrite_from_index_token: the token used for overwriting the input
         ``from_index`` image. This is required for non-privileged users to use
         ``overwrite_from_index``. The format of the token must be in the format "user:password".
+    :param str distribution_scope: the scope for distribution of the index image, defaults to
+        'prod'.
```
mprahl, line 935: This doesn't default to `prod` as far as I'm aware based on the keyword argument default of `None` and `_validate_distribution_scope` just returns the resolved scope. If there is no resolved scope, then it defaults to `prod`.

iib/web/static/api_v1.yaml
```diff
@@ -648,6 +651,9 @@ components:
             index images that IIB does not have access to. The format of the token is in the format
             "user:password".
           example: token
+          distribution_scope:
```
mprahl, line 654: Could you please add a description?

iib/web/static/api_v1.yaml
```diff
@@ -686,6 +692,9 @@ components:
             non-privileged users to use overwrite_from_index. The format of the token must be in
             the format "user:password".
           example: token
+        distribution_scope:
```
mprahl, line 695: Could you please add a description?

iib/workers/tasks/build.py
```diff
@@ -1417,7 +1466,7 @@ def _adjust_csv_annotations(operator_csvs, package_name, organization):
         operator_csv.dump()
 
 
-def _add_ocp_label_to_index(ocp_version, temp_dir, dockerfile_name):
+def _add_label_to_index(label_key, label_value, temp_dir, dockerfile_name):
```
mprahl, line 1469: You need to update the docstring of this function to mention `label_key` and `label_value`.

iib/workers/tasks/build.py
```diff
@@ -1426,8 +1475,31 @@ def _add_ocp_label_to_index(ocp_version, temp_dir, dockerfile_name):
...
             'Added the following line to %s: %s', dockerfile_name, label,
         )
+
+
+def _validate_distribution_scope(resolved_distribution_scope, distribution_scope):
+    """
+    Validate distribution scope is allowed to be updated.
+
+    :param str resolved_distribution_scope: the distribution_scope that the index is for.
+    :param str distribution_scope: the distribution_scope that has been requested the index is for.
```
mprahl, line 1490:
```suggestion
    :param str distribution_scope: the distribution scope that has been requested for the index image.
```

iib/workers/tasks/build.py
```diff
@@ -1426,8 +1475,31 @@ def _add_ocp_label_to_index(ocp_version, temp_dir, dockerfile_name):
...
+    :return: the valid distribution_scope
```
mprahl, line 1491: We can use plain English here instead of variable names :smile: 
```suggestion
    :return: the valid distribution scope
```

iib/workers/tasks/build.py
```diff
@@ -1426,8 +1475,31 @@ def _add_ocp_label_to_index(ocp_version, temp_dir, dockerfile_name):
...
+    :rtype: str
+    :raises IIBError: if the resolved_distribution_scope is of lesser scope then distribution_scope
```
mprahl, line 1493: Typo and should use double back ticks around variable names in the RST docstring.
```suggestion
    :raises IIBError: if the ``resolved_distribution_scope`` is of lesser scope than ``distribution_scope``
```

iib/workers/tasks/build.py
```diff
@@ -1426,8 +1475,31 @@ def _add_ocp_label_to_index(ocp_version, temp_dir, dockerfile_name):
...
+    """
+    if not distribution_scope:
+        return resolved_distribution_scope
+
+    scopes = ["dev", "stage", "prod"]
+    # Make sure the request isn't regressing the distribution scope
+    if scopes.index(distribution_scope) > scopes.index(resolved_distribution_scope):
+        raise IIBError(
+            f'can not set "distribution_scope" to {distribution_scope} because from index is'
```
mprahl, line 1502: Typo and missed capitalization
```suggestion
            f'Cannot set "distribution_scope" to {distribution_scope} because from index is'
```

iib/web/models.py
twaugh: Would that involve building the plumbing to pass configuration through to the workers though? As well as designing what to capture in the configuration to define the semantics of each value.

mprahl: I think we can duplicate the configuration in the API and the workers since this happens infrequently.

iib/web/static/api_v1.yaml
```diff
@@ -581,6 +581,13 @@ components:
             index_image:
               type: string
               example: 'quay.io/iib-stage/iib:5'
+            distribution_scope:
+              description: >
+                The scope of distribution for the index creaded by the request.
```
mprahl, line 586:
```suggestion
                The scope of distribution for the index created by the request.
```

iib/web/static/api_v1.yaml
```diff
@@ -581,6 +581,13 @@ components:
...
+                Will determine what level of protection addition and removal will
```
mprahl, line 587:
```suggestion
                This will determine what level of protection the removal will
```

iib/web/static/api_v1.yaml
```diff
@@ -581,6 +581,13 @@ components:
...
+                have.
```
mprahl, line 588:
```suggestion
                This determined what level of protection the addition or removal had.
```

iib/web/static/api_v1.yaml
```diff
@@ -648,6 +655,13 @@ components:
             index images that IIB does not have access to. The format of the token is in the format
             "user:password".
           example: token
+          distribution_scope:
+            description: >
+                The scope of distribution for the index creaded by the request.
+                Will determine what level of protection addition and removal will
+                have.
```
mprahl, line 662:
```suggestion
                The scope of distribution for the index created by the request.
                This will determine what level of protection the addition will have.
```

iib/web/static/api_v1.yaml
```diff
@@ -686,6 +700,13 @@ components:
             non-privileged users to use overwrite_from_index. The format of the token must be in
             the format "user:password".
           example: token
+        distribution_scope:
+          description: >
+                The scope of distribution for the index creaded by the request.
+                Will determine what level of protection addition and removal will
+                have.
```
mprahl, line 707:
```suggestion
                The scope of distribution for the index created by the request.
                This will determine what level of protection removal will have.
```

yashvardhannanavati (approved):

mprahl (approved):

PR #165, opened 2020-09-29
Using opm force when index has distribution_scope of dev or stage

shawn-hurley:
fixes CLOUDDST-2752


mprahl (approved):

iib/workers/tasks/build.py
```diff
@@ -1000,6 +1011,7 @@ def handle_add_request(
             prebuild_info['binary_image_resolved'],
             from_index,
             overwrite_from_index_token,
+            (prebuild_info['distribution_scope'] in ['dev', 'stage']),
```
lcarva, line 1014: Can we add a test to verify that `_opm_index_add` is being called correctly given the distribution scope? If I change this line to just `False` no tests will fail.

shawn-hurley: Yeah this is an oversight on my part, sorry.

Fixing now.

yashvardhannanavati (approved):

lcarva (approved):

PR #166, opened 2020-09-30
Don't perform index image verification when overwrite_from_index is False

querti:
There is no risk of data loss when IIB doesn't overwrite the original index image. Thus, to prevent unnecessary failures, this check is disabled when overwrite_from_index is False.

More info in CLOUDDST-2783

### DIFF
--- a/iib/workers/tasks/build.py
+++ b/iib/workers/tasks/build.py
@@ -165,6 +165,7 @@ def _update_index_image_pull_spec(
     from_index=None,
     overwrite_from_index=False,
     overwrite_from_index_token=None,
+    resolved_prebuild_from_index=None,
 ):
     """
     Update the request with the modified index image.
@@ -181,11 +182,18 @@ def _update_index_image_pull_spec(
         index image.
     :param str overwrite_from_index_token: the token used for overwriting the input
         ``from_index`` image.
+    :param str resolved_prebuild_from_index: resolved index image before starting the build.
     :raises IIBError: if the manifest list couldn't be created and pushed
     """
     conf = get_worker_config()
     if from_index and overwrite_from_index:
-        _overwrite_from_index(request_id, output_pull_spec, from_index, overwrite_from_index_token)
+        _overwrite_from_index(
+            request_id,
+            output_pull_spec,
+            from_index,
+            resolved_prebuild_from_index,
+            overwrite_from_index_token,
+        )
         index_image = from_index
     elif conf['iib_index_image_output_registry']:
         index_image = output_pull_spec.replace(
@@ -613,7 +621,11 @@ def _opm_index_rm(base_dir, operators, binary_image, from_index, overwrite_from_
 
 
 def _overwrite_from_index(
-    request_id, output_pull_spec, from_index, overwrite_from_index_token=None
+    request_id,
+    output_pull_spec,
+    from_index,
+    resolved_prebuild_from_index,
+    overwrite_from_index_token=None,
 ):
     """
     Overwrite the ``from_index`` image.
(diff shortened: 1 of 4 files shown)


lcarva:
No risk of data loss but IIB may report incorrect from index resolved image data. @shawn-hurley tells me that we can now use a digest for the index image pull spec. If we do that we can safely remove this check for all cases. Can we try that instead?

iib/workers/tasks/build.py
```diff
@@ -1112,9 +1112,10 @@ def handle_rm_request(
             _build_image(temp_dir, 'index.Dockerfile', request_id, arch)
             _push_image(request_id, arch)
 
-    _verify_index_image(
-        prebuild_info['from_index_resolved'], from_index, overwrite_from_index_token
-    )
+    if overwrite_from_index:
+        _verify_index_image(
```
lcarva, line 1116: Maybe in the future, but it would be great if this was done as part of the `_overwrite_from_index` function.

lcarva (approved):

iib/workers/tasks/build.py
querti: Done

querti:
@yashvardhannanavati @shawn-hurley @mprahl check pls

iib/workers/tasks/build.py
```diff
@@ -617,10 +629,13 @@ def _overwrite_from_index(
     :param str output_pull_spec: the pull specification of the manifest list for the index image
         that IIB built.
     :param str from_index: the pull specification of the image to overwrite.
+    :param str resolved_prebuild_from_index: resolved index image before starting the build.
     :param str overwrite_from_index_token: the user supplied token to use when overwriting the
         ``from_index`` image. If this is not set, IIB's configured credentials will be used.
     :raises IIBError: if one of the skopeo commands fails.
```
mprahl, line 635: This needs to also reflect that `_verify_index_image` can raise an exception.

mprahl (approved):
This looks good from a quick glance

querti:
@mprahl @lcarva @yashvardhannanavati Can you merge this? I don't have the rights

midnightercz (approved):

mprahl:
One thing I just thought of is that the resolved from index is marked in the database at the beginning of the request, but if it changed during the request, then what is recorded as the resolved from index may be different than what was used.
https://github.com/release-engineering/iib/blob/99bbcf8cc08b53dee3c4bf0969805632400da472/iib/workers/tasks/build.py#L551-L553

Once the resolved from index pullspec can be used, then we can actually just remove the verification code altogether.

It looks like podman 1.9.2 allows us to specify a digest for the from index:
https://github.com/containers/podman/issues/5234

If you'd still like to pursue with this approach, perhaps you can update the resolved from index image in the database after the opm command.

querti:
> One thing I just thought of is that the resolved from index is marked in the database at the beginning of the request, but if it changed during the request, then what is recorded as the resolved from index may be different than what was used.
> 
> https://github.com/release-engineering/iib/blob/99bbcf8cc08b53dee3c4bf0969805632400da472/iib/workers/tasks/build.py#L551-L553
> 
> Once the resolved from index pullspec can be used, then we can actually just remove the verification code altogether.
> 
> It looks like podman 1.9.2 allows us to specify a digest for the from index:
> [containers/podman#5234](https://github.com/containers/podman/issues/5234)
> 
> If you'd still like to pursue with this approach, perhaps you can update the resolved from index image in the database after the opm command.

I've actually already filed a story for implementing what you're describing. Luiz and I have agreed that this PR is an okay short-term solution (as some teams seem affected by the verification failures). Also, I think that the verification will still be required for overwriting the index image in order to prevent data loss.

mprahl (approved):

yashvardhannanavati:
@querti could you please resolve conflicts and update the PR?

querti:
@yashvardhannanavati done.

PR #168, opened 2020-10-01
Fix backporting bug because of changed folder structure in OPM 1.14.2

yashvardhannanavati:


lcarva (approved):

yashvardhannanavati:
The travis job fails because of a `safety` issue. Skipping fixing it for now since this is a critical fix and needs to get in

PR #171, opened 2020-10-07
Added distribution scope for merge_index_image end point

midnightercz:


iib/workers/tasks/build_merge_index_image.py
```diff
@@ -177,6 +177,7 @@ def handle_merge_request(
     target_index=None,
     overwrite_target_index=False,
     overwrite_target_index_token=None,
+    distribution_scope=None,
```
lcarva, line 180: Will there be a future PR for passing the `distribution_scope` to this task from `api_v1.py`?

mprahl, line 180: Could you please update the doc string?

mprahl: I had the same question.

yashvardhannanavati:
@midnightercz Could you please squash your commits into a single one and rebase this?

midnightercz:
> @midnightercz Could you please squash your commits into a single one and rebase this?
I thought that's why there's "squash and merge" button but it looks like we don't have it enabled for this repo

iib/workers/tasks/build_merge_index_image.py
```diff
@@ -250,6 +254,13 @@ def handle_merge_request(
                 overwrite_target_index_token,
             )
 
+        _add_label_to_index(
```
yashvardhannanavati, line 257: For `Add` request, we call this function in the handle_add_request  because that's where we add the ocp_version label too. Can I please suggest that we move this call to `_add_bundles_missing_in_source` because that's where we add the ocp_version label. The other option is to move the `ocp_version` call here in this function. I would leave it up to you. The latter would be ideal but I would be okay if you move this too to keep things consistent. :)

yashvardhannanavati:
Just one comment. Otherwise, looks good to me.

iib/web/models.py
```diff
@@ -1198,6 +1195,15 @@ def from_json(cls, kwargs, batch=None):
 
         request_kwargs['binary_image'] = Image.get_or_create(pull_specification=binary_image)
 
+        distribution_scope = request_kwargs.pop('distribution_scope', None)
```
shawn-hurley, line 1198: nit: This logic is in two places, we may want to refactor at some point to its own validation function?

tests/test_web/test_api_v1.py
```diff
@@ -1639,11 +1631,12 @@ def test_merge_index_image_overwrite_token_redacted(
 
     rv = client.post('/api/v1/builds/merge-index-image', json=data, environ_base=auth_env)
     rv_json = rv.json
+    # print(rv_json)
     assert rv.status_code == 201
     mock_merge.apply_async.assert_called_once()
     # Second to last element in args is the overwrite_from_index parameter
```
shawn-hurley, line 1637: nit: update the comment?

iib/web/models.py
midnightercz: other validation code for distribution scope is in iib/workers to refactor that, share module will be needed which will then have more then just this validation code and I don't want to include that change in this pull review

iib/web/migrations/versions/983a81fe5e98_added_distribution_scope_attribute_for_.py
```diff
@@ -0,0 +1,30 @@
...
+
+# revision identifiers, used by Alembic.
+revision = '983a81fe5e98'
+down_revision = '4c9db41195ec'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
```
yashvardhannanavati, line 20: Could you please remove these comments?>

iib/web/migrations/versions/983a81fe5e98_added_distribution_scope_attribute_for_.py
```diff
@@ -0,0 +1,30 @@
...
+Revision ID: 983a81fe5e98
+Revises: 4c9db41195ec
+Create Date: 2020-10-08 13:25:25.662595
+
+"""
+from alembic import op
+import sqlalchemy as sa
+
+
+# revision identifiers, used by Alembic.
```
yashvardhannanavati, line 12: This one too

iib/web/models.py
```diff
@@ -1039,7 +1035,7 @@ def from_json(cls, kwargs, batch=None):
         request_kwargs = deepcopy(kwargs)
 
         validate_request_params(
-            request_kwargs, required_params={'from_bundle_image'}, optional_params={'organization'},
+            request_kwargs, required_params={'from_bundle_image'}, optional_params={'organization'}
```
yashvardhannanavati, line 1038: Are these changes automatically done by Python black?

yashvardhannanavati (approved):

PR #174, opened 2020-10-09
Add IIB_BINARY_IMAGE_CONFIG to IIB and make binary_image param optional

yashvardhannanavati:
This commit introduces a new config variable called IIB_BINARY_IMAGE_CONFIG
which when specified, doesn't require the user to specify binary_image param
while submitting requests to IIB.

Refers to CLOUDDST-2761

### DIFF
--- a/iib/web/models.py
+++ b/iib/web/models.py
@@ -629,7 +629,7 @@ class RequestIndexImageMixin:
     @declared_attr
     def binary_image_id(cls):
         """Return the ID of the image that the opm binary comes from."""
-        return db.Column(db.Integer, db.ForeignKey('image.id'), nullable=False)
+        return db.Column(db.Integer, db.ForeignKey('image.id'))
 
     @declared_attr
     def binary_image_resolved_id(cls):
@@ -696,9 +696,10 @@ class RequestIndexImageMixin:
             be created automatically.
         """
         # Validate all required parameters are present
-        required_params = {'binary_image'} | set(additional_required_params or [])
+        required_params = set(additional_required_params or [])
         optional_params = {
             'add_arches',
+            'binary_image',
             'overwrite_from_index',
             'overwrite_from_index_token',
             'distribution_scope',
@@ -752,11 +753,14 @@ class RequestIndexImageMixin:
         Architecture.validate_architecture_json(add_arches)
 
         # Validate binary_image is correctly provided
-        binary_image = request_kwargs.pop('binary_image')
-        if not isinstance(binary_image, str):
+        binary_image = request_kwargs.pop('binary_image', None)
+        if binary_image is not None and not isinstance(binary_image, str):
             raise ValidationError('"binary_image" must be a string')
+        elif not binary_image and not current_app.config['IIB_BINARY_IMAGE_CONFIG']:
+            raise ValidationError('The "binary_image" value must be a non-empty string')
 
-        request_kwargs['binary_image'] = Image.get_or_create(pull_specification=binary_image)
+        if binary_image:
+            request_kwargs['binary_image'] = Image.get_or_create(pull_specification=binary_image)
 
         if 'from_index' in request_kwargs:
             if not isinstance(request_kwargs['from_index'], str):
@@ -790,7 +794,7 @@ class RequestIndexImageMixin:
         :rtype: dict
         """
         return {
-            'binary_image': self.binary_image.pull_specification,
+            'binary_image': getattr(self.binary_image, 'pull_specification', None),
             'binary_image_resolved': getattr(
                 self.binary_image_resolved, 'pull_specification', None
             ),
@@ -812,6 +816,7 @@ class RequestIndexImageMixin:
         :rtype: set
         """
         return {
+            'binary_image',
             'binary_image_resolved',
             'from_bundle_image_resolved',
             'from_index_resolved',
@@ -850,16 +855,10 @@ class RequestAdd(Request, RequestIndexImageMixin):
                 '"bundles" should be either an empty array or an array of non-empty strings'
             )
 
-        # Check if no bundles `from_index and `binary_image` are specified
-        # if no bundles and and no from index then a empty index will be created
-        # if no binary image and just a from_index then we are not updating anything and it would
-        # be a no-op
-        if not request_kwargs.get('bundles') and (
-            not request_kwargs.get('from_index') or not request_kwargs.get('binary_image')
-        ):
-            raise ValidationError(
-                '"from_index" and "binary_image" must be specified if no bundles are specified'
-            )
+        # Check if no bundles and `from_index is specified
+        # if no bundles and no from index then an empty index will be created which is a no-op
+        if not (request_kwargs.get('bundles') or request_kwargs.get('from_index')):
+            raise ValidationError('"from_index" must be specified if no bundles are specified')
 
         for param in ('cnr_token', 'organization'):
             if param not in request_kwargs:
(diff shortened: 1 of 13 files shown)


iib/web/models.py
```diff
@@ -752,11 +753,16 @@ def _from_json(
         Architecture.validate_architecture_json(add_arches)
 
         # Validate binary_image is correctly provided
-        binary_image = request_kwargs.pop('binary_image')
-        if not isinstance(binary_image, str):
+        binary_image = request_kwargs.pop('binary_image', None)
+        if binary_image and not isinstance(binary_image, str):
```
mprahl, line 757: This will catch falsy values of the wrong type
```suggestion
        if binary_image is not None and not isinstance(binary_image, str):
```

iib/web/models.py
```diff
@@ -752,11 +753,16 @@ def _from_json(
...
             raise ValidationError('"binary_image" must be a string')
+        elif not binary_image and not current_app.config['IIB_BINARY_IMAGE_CONFIG']:
+            raise ValidationError(
+                '"binary_image" is required when IIB_BINARY_IMAGE_CONFIG is empty'
```
mprahl, line 761: This error message is exposing an internal detail to the user. I would just say it is required.
```suggestion
                '"binary_image" is required when IIB_BINARY_IMAGE_CONFIG is empty'
```

iib/web/models.py
```diff
@@ -852,16 +859,10 @@ def from_json(cls, kwargs, batch=None):
...
-        # if no binary image and just a from_index then we are not updating anything and it would
-        # be a no-op
-        if not request_kwargs.get('bundles') and (
-            not request_kwargs.get('from_index') or not request_kwargs.get('binary_image')
-        ):
-            raise ValidationError(
-                '"from_index" and "binary_image" must be specified if no bundles are specified'
-            )
+        # Check if no bundles and `from_index is specified
+        # if no bundles and and no from index then a empty index will be created which is a no-op
```
mprahl, line 863:
```suggestion
        # if no bundles and no from index then an empty index will be created which is a no-op
```

iib/web/models.py
```diff
@@ -852,16 +859,10 @@ def from_json(cls, kwargs, batch=None):
...
+        if not request_kwargs.get('bundles') and not request_kwargs.get('from_index'):
+            raise ValidationError('"from_index" must be specified if no bundles are specified')
```
mprahl, line 865: This seems like a change in behavior. Don't we still want to allow creating empty index images?

iib/web/models.py
```diff
@@ -1193,10 +1194,15 @@ def from_json(cls, kwargs, batch=None):
 
         # Validate binary_image is correctly provided
         binary_image = request_kwargs.pop('binary_image', None)
-        if not isinstance(binary_image, str):
+        if binary_image and not isinstance(binary_image, str):
```
mprahl, line 1197: Same as above
```suggestion
        if binary_image is not None and not isinstance(binary_image, str):
```

iib/workers/tasks/build.py
```diff
@@ -699,16 +699,39 @@ def get_index_image_info(overwrite_from_index_token, from_index=None, default_oc
...
+    :param dict binary_image_config: the dict of config required to identify the appropriate
+        ``binary_image`` to use.
+    :return: pull specification of the binary_image to be used for this build.
+    :rtype: str
+    :raises IIBError: when the config value for the ocp_version and distribution_scope is missing.
+    """
+    binary_image = binary_image_config.get(distribution_scope, {}).get(ocp_version, None)
+    if not binary_image:
+        raise IIBError(
+            'No binary image found in IIB_BINARY_IMAGE_CONFIG for'
```
mprahl, line 717: Can you make this more user friendly? The user doesn't care about `IIB_BINARY_IMAGE_CONFIG` being a thing. They just need to know IIB couldn't automatically determine the appropriate binary image to use.

iib/web/models.py
yashvardhannanavati: We were not allowing to create an empty index image in the previous behavior too, hence this change to keep that behavior same. We do not want a request with no `bundles` and no `from_index`. Wasn't that the case previously?

iib/workers/tasks/build.py
```diff
@@ -781,6 +803,19 @@ def _prepare_request_for_build(
         from_index_info['resolved_distribution_scope'], distribution_scope
     )
 
+    if not binary_image:
+        binary_image_ocp_version = from_index_info['ocp_version']
+        binary_image_distribution_scope = distribution_scope
```
mprahl, line 808: What is the purpose of this line?

mprahl:
I'm a bit torn. I think I would prefer if the config lived on the worker and then the API just had a config of `IIB_BINARY_IMAGE_REQUIRED` to determine if the binary image is required by the user.

This makes the architectural separation cleaner since the API doesn't actually care about the values in the configuration.

yashvardhannanavati:
As discussed offline, there's a precedent to this. the IIB_GREENWAVE_CONFIG and we try to keep the worker config specific to celery. Hence, I'm going to keep the implementation the same. Thanks Matt for the review :)

iib/workers/tasks/build.py
yashvardhannanavati: You're right. I assumed a different implementation when `distribution_scope` was included for merge-index-image. But looking at Jindrich's PR, this is not required.

iib/workers/tasks/build.py
yashvardhannanavati: Done

iib/web/models.py
```diff
@@ -852,16 +857,10 @@ def from_json(cls, kwargs, batch=None):
...
-        # if no binary image and just a from_index then we are not updating anything and it would
-        # be a no-op
-        if not request_kwargs.get('bundles') and (
-            not request_kwargs.get('from_index') or not request_kwargs.get('binary_image')
-        ):
-            raise ValidationError(
-                '"from_index" and "binary_image" must be specified if no bundles are specified'
-            )
+        # Check if no bundles and `from_index is specified
+        # if no bundles and and no from index then an empty index will be created which is a no-op
```
mprahl, line 861: You missed this one :stuck_out_tongue: 
```suggestion
        # if no bundles and no from index then an empty index will be created which is a no-op
```

iib/web/models.py
mprahl: You are right. I misread the code.

iib/web/models.py
```diff
@@ -852,16 +857,10 @@ def from_json(cls, kwargs, batch=None):
...
+        if not request_kwargs.get('bundles') and not request_kwargs.get('from_index'):
```
mprahl, line 862: Optional: It's more concise but I'll leave it up to you for readability
```suggestion
        if not (request_kwargs.get('bundles') and request_kwargs.get('from_index')):
```

mprahl (approved):

iib/web/models.py
yashvardhannanavati: Ah, this should be `or` :p

That broke the build lol. I should have checked

iib/web/models.py
yashvardhannanavati: lol, I forgot the migration file too :p

Thanks for pointing it out :)

yashvardhannanavati:
I verified it works on my local :)

midnightercz:
Code looks correct but it introduces another backward incompatibility. I would prefer solution were we simply allow binary image to be None with preserving the current API

midnightercz (approved):

iib/web/models.py
mprahl: Ah correct

PR #178, opened 2020-10-13
Install podman 1.9.3

querti:
Since podman RPM is not available on Centos 8, the source RPM for RHEL 8 is rebuilt by the builder container. One of its build dependencies is not available either so the same process is done for it as well.


mprahl:
@querti this is awesome. I'm fine with assuming x86_64 for now. Could you please mention this bug in the commit message?
https://bugs.centos.org/view.php?id=17786

It's not that podman isn't available, it's just that it's too old. There's some sort of issue where the module stream was not updated in CentOS 8. I'm trying to get attention for it on #centos.

If it's not too much trouble, it might be nice to also build runc with the correct version.

querti:
@mprahl The newest runc version available on RHEL 8 seems to be 1.0.0-66, while only 1.0.0-65 is available on Centos 8. I encountered some errors when rebuilding 1.0.0-66 even after resolving the dependencies. Since podman doesn't explicitly define any runc version as a requirement, I tried installing 1.0.0-65 via dnf and all the operations seemed to work without any issues. Is that an acceptable solution?

docker/Dockerfile-workers
```diff
@@ -1,12 +1,28 @@
 FROM centos:8 AS builder
 
+RUN dnf -y install dnf-plugins-core \
```
mprahl, line 3: Since this is CentOS, you can just install EPEL with `dnf install -y epel-release`.

docker/Dockerfile-workers
```diff
@@ -1,12 +1,28 @@
...
+    && dnf -y install https://dl.fedoraproject.org/pub/epel/epel-release-latest-8.noarch.rpm \
```
mprahl, line 4: Since this is CentOS, you can just install EPEL with `dnf install -y epel-release`.

docker/Dockerfile-workers
```diff
@@ -16,6 +32,13 @@ RUN git clone https://github.com/operator-framework/operator-registry $OPM_SRC &
     (cd $OPM_SRC && git checkout $OPM_TAG && make build) && \
     cp $OPM_SRC/bin/opm /usr/bin/opm
 
+COPY ./docker/podman-1.9.3-2.module+el8.2.1+6867+366c07d6.src.rpm /tmp/podman-1.9.3-2.module+el8.2.1+6867+366c07d6.src.rpm
```
mprahl, line 35: Could you also add a comment above here with a link to the bug?

mprahl:
> @mprahl The newest runc version available on RHEL 8 seems to be 1.0.0-66, while only 1.0.0-65 is available on Centos 8. I encountered some errors when rebuilding 1.0.0-66 even after resolving the dependencies. Since podman doesn't explicitly define any runc version as a requirement, I tried installing 1.0.0-65 via dnf and all the operations seemed to work without any issues. Is that an acceptable solution?

Yes, that is totally fine. Thanks!

docker/Dockerfile-workers
querti: done

docker/Dockerfile-workers
querti: done

mprahl (approved):

querti:
@lcarva @yashvardhannanavati check pls

yashvardhannanavati (approved):

midnightercz (approved):

PR #180, opened 2020-10-15
Check csvName when adding additional bundles

yashvardhannanavati:
In order to avoid adding overwritten bundles to the cleaned up
index image, compare the csvName of the bundles in the target_index
and the source_index.

Refers to CLOUDDST-3133


yashvardhannanavati:
@shawn-hurley @lcarva @mprahl Could you please review this? It's kind of an urgent fix

lcarva (approved):

PR #181, opened 2020-10-19
Change IIB to use index image specified by digest in its operations

querti:
Index image can now be specified by digest, which removes a possible discrepancy between image specified in task's metadata and the one actually being used.


querti:
@yashvardhannanavati @lcarva @mprahl @midnightercz @lipoja check pls

midnightercz (approved):

mprahl:
@querti Could you please document somewhere in the read me that podman 1.9.2 is required by IIB?

lcarva (approved):

PR #196, opened 2021-01-05
Fix worker installation

### LINKED ISSUE #193: Problem when installing worker
When workers image is installed i am getting following error
```
Error: No matching repo to modify: PowerTools.
ERROR: Service 'iib-worker' failed to build : The command '/bin/sh -c dnf -y install dnf-plugins-core     && dnf -y install epel-release     && dnf config-manager --set-enabled PowerTools' returned a non-zero code: 1
```
I fixed it by using (small letters of PowerTools)
```
dnf config-manager --set-enabled powertools
```

Will you fix it or should i make PR?

pbortlov:
Fixes #193

### DIFF
--- a/docker/Dockerfile-workers
+++ b/docker/Dockerfile-workers
@@ -2,7 +2,7 @@ FROM centos:8 AS builder
 
 RUN dnf -y install dnf-plugins-core \
     && dnf -y install epel-release \
-    && dnf config-manager --set-enabled PowerTools
+    && dnf config-manager --set-enabled powertools
 
 RUN dnf -y install \
     --setopt=deltarpm=0 \


docker/Dockerfile-workers
```diff
@@ -2,7 +2,7 @@ FROM centos:8 AS builder
 
 RUN dnf -y install dnf-plugins-core \
     && dnf -y install epel-release \
-    && dnf config-manager --set-enabled PowerTools
+    && dnf config-manager --set-enabled powertools
```
lipoja, line 5: I am not sure with this. 
In CentOS8 the correct repo name is PowerTool and not the lower case. 
Maybe we should ask  the reporter what distro is he using, because this change does not work for me.

yashvardhannanavati: I think it's a valid issue. I tried running locally and had to change to `powertools`

lipoja: Now I get this error too (after removing all container images). I've read somewhere that they used lowercase for this repo in CentOS 8.3 (I guess for 8.2 it should be PowerTools).
I would suggest to change also `FROM centos:8 AS builder` to `FROM centos:8.3 AS builder` in case we want to accept this patch

lipoja: It looks like there is not tag for `8.3` just for `8`

lipoja: I think we might solve it by adding changing it to `centos:latest` ?

lcarva: I'm not sure what's the value of using the `latest` vs the `8` tag.  CentOS is sort of discontinued, so the `8` should always point to the latest image. There will likely never be a `9` tag in the future.

It would be nice if we could specify a minor release, e.g. `8.3`, but this doesn't seem to exist.

IMO, this patch is good as is. 

@pbortlov, could you try to sort out why the travis build failed?

pbortlov: Done @lcarva

lcarva (approved):

docker/Dockerfile-workers
lipoja: OK, I agree with keeping it as it is

lipoja (approved):

PR #198, opened 2021-01-08
Add support for distribution_scope in patch request

yashvardhannanavati:
Refers to CLOUDDST-4297


yashvardhannanavati:
@pbortlov Added the change to upgrade cryptography as a separate commit:)

tests/test_web/test_api_v1.py
```diff
@@ -902,6 +902,7 @@ def test_patch_request_add_success(mock_smfsc, db, minimal_request_add, worker_a
         'state_reason': 'All done!',
         'index_image': 'index:image',
         'binary_image_resolved': 'binary-image@sha256:1234',
+        'distribution_scope': 'stage',
```
lipoja, line 905: `distribution_scope` is optional, right? Should not we have test for request when it is not set?

iib/web/api_v1.py
```diff
@@ -433,6 +433,9 @@ def patch_request(request_id):
             bundle_img = Image.get_or_create(bundle)
             bundle_img.operator = operator_img
 
+    if 'distribution_scope' in payload:
+        setattr(request, 'distribution_scope', payload['distribution_scope'])
```
mprahl, line 437: Why are you using `setattr`?
```suggestion
        request.distribution_scope = payload['distribution_scope']
```

iib/web/api_v1.py
```diff
@@ -433,6 +433,9 @@ def patch_request(request_id):
             bundle_img = Image.get_or_create(bundle)
             bundle_img.operator = operator_img
 
+    if 'distribution_scope' in payload:
```
mprahl, line 436: Don't you need to validate the value?

yashvardhannanavati: We already validate it in build.py. So I don't think we need to do it here as well.

mprahl: Sure, there is a precedent to validate all values on this API endpoint though. The main reason is that the API is separate from the workers, and trust should be limited.

midnightercz:
Overall looks good

iib/web/api_v1.py
yashvardhannanavati: you're right

iib/web/api_v1.py
yashvardhannanavati: So L377 should throw an error if it's a non-empty string. Are you suggesting to validate something else too?

tests/test_web/test_api_v1.py
yashvardhannanavati: good point. added! :)

mprahl (approved):

PR #203, opened 2021-01-20
Parse bundle version when propagating bundles in merge-index-image

yashvardhannanavati:
Refers to CLOUDDST-4452


yashvardhannanavati:
@pbortlov

iib/workers/tasks/build_merge_index_image.py
```diff
@@ -285,3 +287,63 @@ def handle_merge_request(
...
+            if bundle_version_label.startswith('='):
+                bundle_version = float(bundle_version.strip('='))
+                if bundle_version == float_valid_ocp_version:
+                    index_valid_bundles.append(bundle)
+            elif '-' in bundle_version_label:
+                first_version, second_version = [
+                    float(version) for version in bundle_version.split('-')
+                ]
+                if (
+                    first_version <= float_valid_ocp_version
```
lcarva, line 326: Maybe more pythonic this way?
```python
if first_version <= float_valid_ocp_version <= second_version:
  ...
```

iib/workers/tasks/build_merge_index_image.py
```diff
@@ -285,3 +287,63 @@ def handle_merge_request(
...
+    for bundle in bundle_paths:
+        try:
+            bundle_version_label = get_image_label(bundle, 'com.redhat.openshift.versions')
+            bundle_version = bundle_version_label.replace('v', '')
+            if bundle_version_label.startswith('='):
+                bundle_version = float(bundle_version.strip('='))
+                if bundle_version == float_valid_ocp_version:
+                    index_valid_bundles.append(bundle)
+            elif '-' in bundle_version_label:
+                first_version, second_version = [
```
lcarva, line 322: Nit: `min_version` and `max_version` are more descriptive names.

iib/workers/tasks/build_merge_index_image.py
```diff
@@ -285,3 +287,63 @@ def handle_merge_request(
...
+                if (
+                    first_version <= float_valid_ocp_version
+                    and float_valid_ocp_version <= second_version
+                ):
+                    index_valid_bundles.append(bundle)
+            elif "," in bundle_version_label:
+                versions = [float(version) for version in bundle_version.split(",")]
+                # This means the version is something like v4.6, v4.5 which is not valid
+                if versions != sorted(versions):
+                    invalid_version_bundles[bundle] = bundle_version_label
```
lcarva, line 334: For debugging purposes, we should log the reason here.

iib/workers/tasks/build_merge_index_image.py
```diff
@@ -285,3 +287,63 @@ def handle_merge_request(
...
+                    continue
+                if float_valid_ocp_version in versions or float_valid_ocp_version > versions[0]:
```
lcarva, line 336: Why not this?

```python
if float_valid_ocp_version >= versions[0]:
  ...
```

The items should be sorted at this point.

iib/workers/tasks/build_merge_index_image.py
```diff
@@ -285,3 +287,63 @@ def handle_merge_request(
...
+
+           |  "v4.5"   |   "=v4.6"    | "v4.5-v4.7" | "v4.5,v4.6"
+    ---------------------------------------------------------------
+    v4.5   | included  | NOT included |  included   |  included
+    ---------------------------------------------------------------
+    v4.6   | included  |   included   |  included   |  included
+    ---------------------------------------------------------------
+    v4.7   | included  | NOT included |  included   |  included
+    """
+    index_valid_bundles = []
```
lcarva, line 310: I think `valid` and `invalid` is not the correct terminology. `index_valid_bundles` contains a list of bundle images that are allowed to be included in the index image, while `invalid_version_bundles` contains a list of bundle images that contain unsupported metadata. There's a third category which are the bundle images that contain the correct metadata, but should not be added to the index image. 

IMO, this function would be better off if it only processed a single bundle at a time, and it only returned a boolean which indicates whether or not the bundle should be added to the index. If the bundle image contains unsupported metadata, log a warning and return False.

tests/test_workers/test_tasks/test_build_merge_index_image.py
```diff
@@ -227,6 +228,7 @@ def test_add_bundles_missing_in_source(
             'csvName': 'bundle5-5.0',
         },
     ]
+    mock_gil.side_effect = ['=v4.5', '=v4.6', 'v4.7', 'v4.5-v4.7', 'v4.5,v4.6']
```
lcarva, line 231: Could you add, maybe new test cases, that explicitly check the behavior of `filter_invalid_bundles`?

pbortlov:
@yashvardhannanavati could you add the JIrat ticket link here?

iib/workers/tasks/build_merge_index_image.py
pbortlov, line 336: maybe:
else: a message what was wrong?

iib/workers/tasks/build_merge_index_image.py
```diff
@@ -89,7 +90,12 @@ def _add_bundles_missing_in_source(
         ):
             missing_bundles.append(bundle)
 
-    missing_bundle_paths = [bundle['bundlePath'] for bundle in missing_bundles]
+    missing_bundle_paths = [
+        bundle['bundlePath']
+        for bundles in (missing_bundles, source_index_bundles)
```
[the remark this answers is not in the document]
yashvardhannanavati: Because, all the bundles present in the source_index might not be valid for the target_index. Hence, we will have to vet them all.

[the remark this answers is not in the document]
yashvardhannanavati: elegant. I like it. thanks!

iib/workers/tasks/build_merge_index_image.py
```diff
@@ -89,7 +90,12 @@ def _add_bundles_missing_in_source(
...
+        for bundle in bundles
```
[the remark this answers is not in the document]
yashvardhannanavati: It's actually the other way around. I get `bundles` from the tuple and then iterate over it and use `bundle['bundlePath']`. Anyways it's going to be replaced with `chain`

iib/workers/tasks/build_merge_index_image.py
```diff
@@ -99,11 +105,7 @@ def _add_bundles_missing_in_source(
         )
 
     _opm_index_add(
-        base_dir,
-        missing_bundle_paths,
-        binary_image,
-        source_from_index,
-        overwrite_target_index_token,
+        base_dir, missing_bundle_paths, binary_image, overwrite_target_index_token,
```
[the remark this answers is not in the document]
yashvardhannanavati: It's automatically formatted by `python-black`. https://black.readthedocs.io/en/stable/

iib/workers/tasks/build_merge_index_image.py
```diff
@@ -285,3 +287,63 @@ def handle_merge_request(
...
+    v4.7   | included  | NOT included |  included   |  included
+    """
+    index_valid_bundles = []
+    invalid_version_bundles = {}
+    float_valid_ocp_version = float(valid_ocp_version.replace('v', ''))
+    for bundle in bundle_paths:
+        try:
+            bundle_version_label = get_image_label(bundle, 'com.redhat.openshift.versions')
+            bundle_version = bundle_version_label.replace('v', '')
+            if bundle_version_label.startswith('='):
```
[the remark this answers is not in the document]
yashvardhannanavati: because if we use `bundle_version` here, something invalid such as `v=4.5` would be deemed as valid which would be wrong.

iib/workers/tasks/build_merge_index_image.py
```diff
@@ -89,7 +90,12 @@ def _add_bundles_missing_in_source(
         ):
             missing_bundles.append(bundle)
 
-    missing_bundle_paths = [bundle['bundlePath'] for bundle in missing_bundles]
+    missing_bundle_paths = [
+        bundle['bundlePath']
+        for bundles in (missing_bundles, source_index_bundles)
+        for bundle in bundles
+    ]
+    missing_bundle_paths = filter_invalid_bundles(missing_bundle_paths, ocp_version)
```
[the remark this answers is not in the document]
yashvardhannanavati: Sounds good. I'll make the change.

iib/workers/tasks/build_merge_index_image.py
```diff
@@ -285,3 +287,63 @@ def handle_merge_request(
...
+    invalid_version_bundles = {}
+    float_valid_ocp_version = float(valid_ocp_version.replace('v', ''))
+    for bundle in bundle_paths:
+        try:
+            bundle_version_label = get_image_label(bundle, 'com.redhat.openshift.versions')
+            bundle_version = bundle_version_label.replace('v', '')
+            if bundle_version_label.startswith('='):
+                bundle_version = float(bundle_version.strip('='))
+                if bundle_version == float_valid_ocp_version:
+                    index_valid_bundles.append(bundle)
```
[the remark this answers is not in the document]
yashvardhannanavati: the `bundle_version` not being the same as the `float_valid_ocp_version` would not exactly mean that the bundle version is invalid. It would simply mean that it's not valid for this particular target_index. It might be valid in general for some other ocp_version

iib/workers/tasks/build_merge_index_image.py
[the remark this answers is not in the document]
lcarva: I see. Thanks for the explanation.

iib/workers/tasks/build_merge_index_image.py
```diff
@@ -285,3 +287,63 @@ def handle_merge_request(
...
+            elif '-' in bundle_version_label:
```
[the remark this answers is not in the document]
yashvardhannanavati: Same reason as I mentioned above. So that an invalid version doesn't pass off as a correct version.

iib/workers/tasks/build_merge_index_image.py
```diff
@@ -285,3 +287,63 @@ def handle_merge_request(
...
+                first_version, second_version = [
+                    float(version) for version in bundle_version.split('-')
+                ]
+                if (
+                    first_version <= float_valid_ocp_version
+                    and float_valid_ocp_version <= second_version
+                ):
+                    index_valid_bundles.append(bundle)
```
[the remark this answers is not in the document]
yashvardhannanavati: This wouldn't conclude that it isn't valid. It would simply mean that this bundle version cannot be added to the current `target_index`. Out of the context of this target_index, the bundle version might actually be valid.

iib/workers/tasks/build_merge_index_image.py
```diff
@@ -285,3 +288,58 @@ def handle_merge_request(
...
+                return True
+        elif '-' in bundle_version_label:
+            min_version, max_version = [float(version) for version in bundle_version.split('-')]
+            if min_version <= float_valid_ocp_version <= max_version:
+                return True
+        elif "," in bundle_version_label:
+            versions = [float(version) for version in bundle_version.split(",")]
+            # This means the version is something like v4.6, v4.5 which is not valid
+            if versions != sorted(versions):
+                log.warning(
```
lcarva, line 328: Just raise a `ValueError` so it goes through the same path as the other invalid cases.

iib/workers/tasks/build_merge_index_image.py
```diff
@@ -285,3 +288,58 @@ def handle_merge_request(
...
+    v4.5   | included  | NOT included |  included   |  included
+    ---------------------------------------------------------------
+    v4.6   | included  |   included   |  included   |  included
+    ---------------------------------------------------------------
+    v4.7   | included  | NOT included |  included   |  included
+    """
+    try:
+        float_valid_ocp_version = float(valid_ocp_version.replace('v', ''))
+    except ValueError:
+        raise IIBError(f'Invalid ocp_version "{valid_ocp_version}" specified for target_index')
```
lcarva, line 313: Within the scope of this function `target_index` and `ocp_version` are unknown variables. Can we tweak the terminology to say something like:
> Invalid OCP version, "vYolo", specified in Index Image.

If knowing which Index Image is important, you can log a message before `is_bundle_version_valid` is called to give additional context.

lcarva (commented):
Just a couple of very minor points. LGTM otherwise!

iib/workers/tasks/build_merge_index_image.py
yashvardhannanavati: +1

iib/workers/tasks/build_merge_index_image.py
yashvardhannanavati: Done

iib/workers/tasks/build_merge_index_image.py
yashvardhannanavati: Also added a test to cover that.

lcarva (approved):

PR #205, opened 2021-01-22
Add support for `deprecation_list` in Add endpoint

yashvardhannanavati:
Refers to CLOUDDST-3835

### DIFF
--- a/iib/web/migrations/versions/7573241a5156_rename_bundle_deprecation_association_.py
+++ b/iib/web/migrations/versions/7573241a5156_rename_bundle_deprecation_association_.py
@@ -1,0 +1,48 @@
+"""Rename bundle deprecation association for merge index.
+
+Revision ID: 7573241a5156
+Revises: eec630370e68
+Create Date: 2021-02-09 13:50:01.905796
+
+"""
+from alembic import op
+
+
+# revision identifiers, used by Alembic.
+revision = '7573241a5156'
+down_revision = 'eec630370e68'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+
+    with op.batch_alter_table('bundle_deprecation', schema=None) as batch_op:
+        batch_op.drop_index('ix_bundle_deprecation_bundle_id')
+        batch_op.drop_index('ix_bundle_deprecation_merge_index_image_id')
+
+    op.rename_table('bundle_deprecation', 'request_merge_bundle_deprecation', schema=None)
+
+    with op.batch_alter_table('request_merge_bundle_deprecation', schema=None) as batch_op:
+        batch_op.create_index(
+            batch_op.f('ix_request_merge_bundle_deprecation_bundle_id'), ['bundle_id'], unique=False
+        )
+        batch_op.create_index(
+            batch_op.f('ix_request_merge_bundle_deprecation_merge_index_image_id'),
+            ['merge_index_image_id'],
+            unique=False,
+        )
+
+
+def downgrade():
+    with op.batch_alter_table('request_merge_bundle_deprecation', schema=None) as batch_op:
+        batch_op.drop_index(batch_op.f('ix_request_merge_bundle_deprecation_merge_index_image_id'))
+        batch_op.drop_index(batch_op.f('ix_request_merge_bundle_deprecation_bundle_id'))
+
+    op.rename_table('request_merge_bundle_deprecation', 'bundle_deprecation', schema=None)
+
+    with op.batch_alter_table('bundle_deprecation', schema=None) as batch_op:
+        batch_op.create_index(
+            'ix_bundle_deprecation_merge_index_image_id', ['merge_index_image_id'], unique=False
+        )
+        batch_op.create_index('ix_bundle_deprecation_bundle_id', ['bundle_id'], unique=False)
--- a/iib/web/migrations/versions/eec630370e68_support_deprecation_list_in_add_request_.py
+++ b/iib/web/migrations/versions/eec630370e68_support_deprecation_list_in_add_request_.py
@@ -1,0 +1,47 @@
+"""Support deprecation_list in Add request type.
+
+Revision ID: eec630370e68
+Revises: 60f89c046096
+Create Date: 2021-01-20 20:36:29.184275
+
+"""
+from alembic import op
+import sqlalchemy as sa
+
+
+# revision identifiers, used by Alembic.
+revision = 'eec630370e68'
+down_revision = '60f89c046096'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    op.create_table(
+        'request_add_bundle_deprecation',
+        sa.Column('request_add_id', sa.Integer(), autoincrement=False, nullable=False),
+        sa.Column('bundle_id', sa.Integer(), autoincrement=False, nullable=False),
+        sa.ForeignKeyConstraint(['bundle_id'], ['image.id'],),
+        sa.ForeignKeyConstraint(['request_add_id'], ['request_add.id'],),
+        sa.PrimaryKeyConstraint('request_add_id', 'bundle_id'),
+        sa.UniqueConstraint(
+            'request_add_id', 'bundle_id', name='request_add_bundle_deprecation_constraint'
+        ),
+    )
+    with op.batch_alter_table('request_add_bundle_deprecation', schema=None) as batch_op:
+        batch_op.create_index(
+            batch_op.f('ix_request_add_bundle_deprecation_bundle_id'), ['bundle_id'], unique=False
+        )
+        batch_op.create_index(
+            batch_op.f('ix_request_add_bundle_deprecation_request_add_id'),
+            ['request_add_id'],
+            unique=False,
+        )
+
+
+def downgrade():
+    with op.batch_alter_table('request_add_bundle_deprecation', schema=None) as batch_op:
+        batch_op.drop_index(batch_op.f('ix_request_add_bundle_deprecation_request_add_id'))
+        batch_op.drop_index(batch_op.f('ix_request_add_bundle_deprecation_bundle_id'))
+
+    op.drop_table('request_add_bundle_deprecation')
--- a/iib/web/models.py
+++ b/iib/web/models.py
@@ -106,6 +106,29 @@ class BundleDeprecation(db.Model):
     )
 
 
+class RequestAddBundleDeprecation(db.Model):
+    """An association table between add requests and bundle images which they deprecate."""
+
+    # A primary key is required by SQLAlchemy when using declaritive style tables, so a composite
+    # primary key is used on the two required columns
+    request_add_id = db.Column(
+        db.Integer,
+        db.ForeignKey('request_add.id'),
+        autoincrement=False,
+        index=True,
+        primary_key=True,
+    )
+    bundle_id = db.Column(
+        db.Integer, db.ForeignKey('image.id'), autoincrement=False, index=True, primary_key=True
+    )
+
+    __table_args__ = (
+        db.UniqueConstraint(
+            'request_add_id', 'bundle_id', name='request_add_bundle_deprecation_constraint'
+        ),
+    )
+
+
 class Architecture(db.Model):
     """An architecture associated with an image."""
 
@@ -800,6 +823,7 @@ class RequestIndexImageMixin:
             ),
             'bundle_mapping': {},
             'bundles': [],
+            'deprecation_list': [],
             'from_index': getattr(self.from_index, 'pull_specification', None),
             'from_index_resolved': getattr(self.from_index_resolved, 'pull_specification', None),
             'index_image': getattr(self.index_image, 'pull_specification', None),
@@ -832,6 +856,7 @@ class RequestAdd(Request, RequestIndexImageMixin):
 
     id = db.Column(db.Integer, db.ForeignKey('request.id'), autoincrement=False, primary_key=True)
     bundles = db.relationship('Image', secondary=RequestAddBundle.__table__)
+    deprecation_list = db.relationship('Image', secondary=RequestAddBundleDeprecation.__table__)
     organization = db.Column(db.String, nullable=True)
 
     omps_operator_version = db.Column(db.String, nullable=True)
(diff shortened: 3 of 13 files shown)


yashvardhannanavati:
Working on the API documentation changes.

iib/web/models.py
```diff
@@ -106,6 +106,29 @@ class BundleDeprecation(db.Model):
     )
 
 
+class AddRequestBundleDeprecation(db.Model):
```
pbortlov, line 109: Would be worth it to merge it with th class BundleDeprecation(db.Model)? or create a general one for deprecation?

iib/web/models.py
```diff
@@ -882,13 +908,15 @@ def from_json(cls, kwargs, batch=None):
...
                 'distribution_scope',
+                'deprecation_list',
             ],
             batch=batch,
         )
 
-        request_kwargs['bundles'] = [
-            Image.get_or_create(pull_specification=item) for item in bundles
-        ]
+        for key in ('bundles', 'deprecation_list'):
```
pbortlov, line 916: Why is at line #876 the same for cycle? Could we put it together?

iib/workers/tasks/build_merge_index_image.py
```diff
@@ -251,17 +200,21 @@ def handle_merge_request(
 
         set_request_state(request_id, 'in_progress', 'Deprecating bundles in the deprecation list')
         log.info('Deprecating bundles in the deprecation list')
-        intermediate_bundles = source_index_bundles + missing_bundles
-        deprecate_bundles = _get_bundles_from_deprecation_list(
+        intermediate_bundles = [
+            bundle['bundlePath']
+            for bundles in (source_index_bundles, missing_bundles)
```
pbortlov, line 205: is this from another open PR?

iib/workers/tasks/utils.py
```diff
@@ -23,6 +24,139 @@
 dogpile_cache_region = create_dogpile_region()
 
 
+def deprecate_bundles(
+    bundles, base_dir, binary_image, from_index, overwrite_target_index_token=None
```
pbortlov, line 28: can we add each attribute on separate line?

tests/test_web/test_api_v1.py
```diff
@@ -46,6 +46,7 @@ def test_get_build(app, auth_env, client, db):
         'binary_image_resolved': 'quay.io/namespace/binary_image@sha256:abcdef',
         'bundle_mapping': {},
         'bundles': ['quay.io/namespace/bundle:1.0-3'],
+        'deprecation_list': [],
```
pbortlov, line 49: I have a feeling that this test is not complete. If you use `get_build` function. You can get more request types that don't have the same structure of the response. Would it be useful to modify the name of the test that we know it tests only add request_type?

tests/test_workers/test_tasks/test_build.py
```diff
@@ -830,14 +625,13 @@ def test_skopeo_copy_fail_max_retries(mock_run_cmd):
 @mock.patch('iib.workers.tasks.build.get_legacy_support_packages')
 @mock.patch('iib.workers.tasks.build.validate_legacy_params_and_config')
 @mock.patch('iib.workers.tasks.build.gate_bundles')
-@mock.patch('iib.workers.tasks.build._get_resolved_bundles')
+@mock.patch('iib.workers.tasks.build.get_resolved_bundles')
 @mock.patch('iib.workers.tasks.build._add_label_to_index')
 @mock.patch('iib.workers.tasks.build._get_present_bundles')
-@mock.patch('iib.workers.tasks.build._get_missing_bundles')
+# @mock.patch('iib.workers.tasks.build._get_missing_bundles')
```
pbortlov, line 631: what is this commented?

tests/test_workers/test_tasks/test_build.py
```diff
@@ -815,6 +607,9 @@ def test_skopeo_copy_fail_max_retries(mock_run_cmd):
     'force_backport, binary_image', ((True, 'binary-image:latest'), (False, None))
 )
 @pytest.mark.parametrize('distribution_scope', ('dev', 'stage', 'prod'))
+@pytest.mark.parametrize('deprecate_bundles', (True, False))
```
pbortlov, line 610: Isn't the order of `imports` wrong? lines 657 and 658

iib/web/models.py
yashvardhannanavati: So the BundleDeprecation is an association table between the merge-index-image requests and bundle images. This one is an association between the Add requests and bundle images. I think we can rename BundleDeprecation to MergeIndexImageRequestBundleDeprecation to reduce confusion as a separate commit.

iib/web/models.py
yashvardhannanavati: I kept them separate because the #876 is for validating the input and this one creates entries into the database. So it made sense to complete all the validation before creating entries into the database.

iib/workers/tasks/build_merge_index_image.py
yashvardhannanavati: Nope, I changed the input for `get_bundles_from_deprecation_list` hence this change. the other PR has the change in `_add_bundles_missing_in_source`

yashvardhannanavati: I'll change it to use `chain` though

iib/workers/tasks/utils.py
yashvardhannanavati: Sorry, it's done by python-black auto-formatter :)

tests/test_web/test_api_v1.py
yashvardhannanavati: I think this test was written more to test the outcome of the `get_build` endpoint rather than specific response types. The other request types have their own tests. But these are my two cents. Please let me know what you think

tests/test_workers/test_tasks/test_build.py
yashvardhannanavati: good catch! I forgot to remove this one

tests/test_workers/test_tasks/test_build.py
yashvardhannanavati: why so? Could you please elaborate on this one? Am I missing something?

pbortlov:
Hi @yashvardhannanavati, I miss changes in `iib/web/static/api_v1.yaml`

tests/test_workers/test_tasks/test_build.py
pbortlov: sorry, my bad. Everything is correct

iib/workers/tasks/utils.py
pbortlov: no problem

iib/web/models.py
pbortlov: Can I ask you for comments there?

iib/web/models.py
pbortlov: it would make less confusion if you rename it. thank you

tests/test_web/test_api_v1.py
pbortlov: Sounds ok if the intent was this.

yashvardhannanavati:
@midnightercz @lipoja could you PTAL?

yashvardhannanavati:
@pbortlov yes, on it.

iib/web/models.py
pbortlov: @yashvardhannanavati Could you add a comment here?

pbortlov:
@yashvardhannanavati the documentation looks fine :+1:

iib/workers/tasks/build.py
```diff
@@ -1151,7 +1077,25 @@ def handle_add_request(
...
         )
 
+        present_bundles_pull_spec = [bundle['bundlePath'] for bundle in present_bundles]
+        deprecation_bundles = get_bundles_from_deprecation_list(
+            present_bundles_pull_spec + resolved_bundles, deprecation_list
+        )
+
         arches = prebuild_info['arches']
+        if deprecation_bundles:
+            arch = 'amd64' if 'amd64' in arches else arches[0]
```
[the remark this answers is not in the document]
yashvardhannanavati: `BundleDeprecation` uses bundle_id so keeping it consistent. If you feel strongly about it, I can change it

iib/web/models.py
```diff
@@ -106,6 +106,29 @@ class BundleDeprecation(db.Model):
...
+    # A primary key is required by SQLAlchemy when using declaritive style tables, so a composite
+    # primary key is used on the two required columns
+    request_add_id = db.Column(
+        db.Integer,
+        db.ForeignKey('request_add.id'),
+        autoincrement=False,
+        index=True,
+        primary_key=True,
+    )
+    bundle_id = db.Column(
```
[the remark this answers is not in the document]
yashvardhannanavati: BundleDeprecation uses bundle_id so keeping it consistent. If you feel strongly about it, I can change it

iib/workers/tasks/build.py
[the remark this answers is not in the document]
yashvardhannanavati: This is how the merge-index-image does it when deprecating bundles. So I used this approach. I can change it if you want. Doesn't make much of a difference :)

midnightercz (approved):

iib/workers/tasks/build.py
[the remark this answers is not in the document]
lcarva: I think it's important to be consistent because we're matching bundles by digest. If our API endpoints are not using the same logic, there's a risk that something that should be deprecated doesn't. Can we have a follow up PR/ticket for this? I don't think we need to address it as part of this PR.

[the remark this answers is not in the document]
yashvardhannanavati: ack, filed https://projects.engineering.redhat.com/browse/CLOUDDST-4901

pbortlov (approved):

iib/web/migrations/versions/7573241a5156_rename_bundle_deprecation_association_.py
```diff
@@ -0,0 +1,70 @@
...
+            batch_op.f('ix_request_merge_bundle_deprecation_merge_index_image_id'),
+            ['merge_index_image_id'],
+            unique=False,
+        )
+
+    with op.batch_alter_table('bundle_deprecation', schema=None) as batch_op:
+        batch_op.drop_index('ix_bundle_deprecation_bundle_id')
+        batch_op.drop_index('ix_bundle_deprecation_merge_index_image_id')
+
+    op.drop_table('bundle_deprecation')
```
lcarva, line 45: Hey, I didn't catch this at first. Is there data in this table that needs to be migrated to the new table?

iib/web/migrations/versions/7573241a5156_rename_bundle_deprecation_association_.py
```diff
@@ -0,0 +1,70 @@
...
+
+# revision identifiers, used by Alembic.
+revision = '7573241a5156'
+down_revision = 'eec630370e68'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    op.create_table(
```
lcarva, line 20: If you're just renaming the table, can you just use [rename_table](https://alembic.sqlalchemy.org/en/latest/ops.html#alembic.operations.Operations.rename_table) ?

iib/web/migrations/versions/7573241a5156_rename_bundle_deprecation_association_.py
```diff
@@ -0,0 +1,70 @@
...
+        batch_op.drop_index('ix_bundle_deprecation_bundle_id')
+        batch_op.drop_index('ix_bundle_deprecation_merge_index_image_id')
+
+    op.drop_table('bundle_deprecation')
+
+
+def downgrade():
+    op.create_table(
+        'bundle_deprecation',
+        sa.Column('merge_index_image_id', sa.INTEGER(), nullable=False),
```
lcarva, line 51: In the other `create_table` call, `sa.Integer()` is used. Should we be consistent here instead of using `sa.INTEGER()` ?

iib/workers/tasks/utils.py
```diff
@@ -23,6 +24,139 @@
...
+    """
+    Get the pull specification of the container image using its digest.
+
+    :param str pull_spec: the pull specification of the container image to resolve
+    :return: the resolved pull specification
+    :rtype: str
+    """
+    log.debug('Resolving %s', pull_spec)
+    name = _get_container_image_name(pull_spec)
+    skopeo_output = skopeo_inspect(f'docker://{pull_spec}', '--raw', return_json=False)
```
lcarva, line 144: huh odd, why do we pass `return_json=False` only to use `json.loads` on the next line?

iib/web/migrations/versions/eec630370e68_support_deprecation_list_in_add_request_.py
```diff
@@ -0,0 +1,47 @@
...
+
+
+# revision identifiers, used by Alembic.
+revision = 'eec630370e68'
+down_revision = '60f89c046096'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
```
lcarva, line 19: It would be super if we could add tests for these db migrations in https://github.com/release-engineering/iib/blob/master/tests/test_web/test_migrations.py.

iib/web/migrations/versions/7573241a5156_rename_bundle_deprecation_association_.py
yashvardhannanavati: hm, good catch. I believe we will have some data there.

yashvardhannanavati: I'm going to put the renaming part in a separate PR to unblock the previous change.

iib/workers/tasks/utils.py
yashvardhannanavati: /me assumes it's due to a skopeo quirk

iib/web/migrations/versions/7573241a5156_rename_bundle_deprecation_association_.py
yashvardhannanavati: On second thoughts, we have never used deprecation list yet, they were all empty. So I believe it's okay to not care about it.

lcarva (approved):

PR #213, opened 2021-02-17
Fix deprecate property in add request

lcarva:

### DIFF
--- a/iib/workers/tasks/utils.py
+++ b/iib/workers/tasks/utils.py
@@ -25,7 +25,12 @@ dogpile_cache_region = create_dogpile_region()
 
 
 def deprecate_bundles(
-    bundles, base_dir, binary_image, from_index, overwrite_target_index_token=None
+    bundles,
+    base_dir,
+    binary_image,
+    from_index,
+    overwrite_target_index_token=None,
+    container_tool=None,
 ):
     """
     Deprecate the specified bundles from the index image.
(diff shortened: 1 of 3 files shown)


iib/workers/tasks/utils.py
```diff
@@ -25,7 +25,12 @@
 
 
 def deprecate_bundles(
-    bundles, base_dir, binary_image, from_index, overwrite_target_index_token=None
+    bundles,
```
pbortlov, line 28: I would suggest doing mapping: bundles=bundles etc. It will help to be readable and it will ensure that no one will mix it. I say this because I found a bug today.

lcarva: I don't think we should do that because that would imply `bundles` is an optional parameter which is not correct.

pbortlov: I don't think I explain myself correctly. 
The optional argument here is `overwrite_target_index_token` only. I don't think mapping arguments by name will make them optional. 
I like Martin Basti idea about arguments from here https://github.com/release-engineering/iib/pull/214. 
https://www.vegardstikbakke.com/python-keyword-only/

lipoja: @pbortlov I am not sure if this is the correct place. I think you meant to have this comment on function call and not when function is defined. If you do it here it would make the parameter optional as @lcarva said. 
However we should discuss a way if we can force developers to specify parameters when they call a function.

pbortlov: Oh, yeah. @lipoja You are right. That's the definition, my bad.

pbortlov (approved):

mprahl (approved):

PR #214, opened 2021-02-18
Fix _opm_index_add map overwrite_from_index_token [CLOUDDST-5026]

pbortlov:
* return  `source_from_index` paramater to `_opm_index_add` function

### DIFF
--- a/iib/workers/tasks/build_merge_index_image.py
+++ b/iib/workers/tasks/build_merge_index_image.py
@@ -109,7 +109,10 @@ def _add_bundles_missing_in_source(
         )
 
     _opm_index_add(
-        base_dir, missing_bundle_paths, binary_image, overwrite_target_index_token,
+        base_dir,
+        missing_bundle_paths,
+        binary_image,
+        overwrite_from_index_token=overwrite_target_index_token,
     )
     _add_label_to_index(
         'com.redhat.index.delivery.version', ocp_version, base_dir, 'index.Dockerfile'
--- a/tests/test_workers/test_tasks/test_build_merge_index_image.py
+++ b/tests/test_workers/test_tasks/test_build_merge_index_image.py
@@ -262,7 +262,7 @@ def test_add_bundles_missing_in_source(
             'quay.io/bundle2@sha256:456132',
         ],
         'binary-image:4.5',
-        None,
+        overwrite_from_index_token=None,
     )
     assert mock_gil.call_count == 5
     assert mock_aolti.call_count == 2
@@ -441,7 +441,7 @@ def test_add_bundles_missing_in_source_none_missing(
         'some_dir',
         ['quay.io/bundle3@sha256:123456', 'quay.io/bundle4@sha256:123456'],
         'binary-image:4.5',
-        None,
+        overwrite_from_index_token=None,
     )
     assert mock_gil.call_count == 4
     assert mock_aolti.call_count == 2


pbortlov:
@lcarva @chandwanitulsi @querti @midnightercz @lipoja PTAL. I found out that after the new release the source_from_index was missing in _opm_index_add function. We found out that IIB QE has problems because the parameter `from_index=source_from_index` was missing and it took the `overwrite_from_index_token` as `from_index` since the paramaters weren't mapped correctly. Big thank you to @lipoja and Emily for helping me out to figure out.

iib/workers/tasks/build_merge_index_image.py
```diff
@@ -109,20 +113,32 @@ def _add_bundles_missing_in_source(
         )
 
     _opm_index_add(
-        base_dir, missing_bundle_paths, binary_image, overwrite_target_index_token,
+        base_dir=base_dir,
+        bundles=missing_bundle_paths,
+        binary_image=binary_image,
+        from_index=source_from_index,
```
pbortlov, line 119: The problem was here: from_index=source_from_index was removed in previous release.

lipoja (approved):

MartinBasti:
Can we separate fix and refactoring to different comits? Can you also merge test fix into commit that causes it?

lipoja:
@pbortlov I think that @MartinBasti is right. Can we separate the code into two commits.

pbortlov:
done @lipoja @MartinBasti

MartinBasti:
Because you already updated function calls to use kwarg names, do you think that it may be useful to do it on python level and enforce calling functions with named parameters because param list is too long and may lead to potentially same issue in future?

```python
def func(*, param1, param2, ...)

def method(self, *, param1, param2, ...)
```
https://www.vegardstikbakke.com/python-keyword-only/

pbortlov:
> Because you already updated function calls to use kwarg names, do you think that it may be useful to do it on python level and enforce calling functions with named parameters because param list is too long and may lead to potentially same issue in future?
> 
> ```python
> def func(*, param1, param2, ...)
> 
> def method(self, *, param1, param2, ...)
> ```
> 
> https://www.vegardstikbakke.com/python-keyword-only/
> 
> Edit: Not necessary in this PR if this has to be merged quickly as a bugfix

This should be a fix. I would suggest you talk to the hello-operator guild about this idea. :)

pbortlov:
I talked to @lipoja and we decided to have this PR only for the bug fix

lipoja (approved):

MartinBasti (approved):

querti (approved):

pbortlov:
Update: After I talked to @lcarva about this issue, I understood that removing `source_from_index` was intentional. Unfortunately, if the `overwrite_from_index_token` is not mapped in _opm_index_add() correctly it function can misunderstand if with `source_from_index`. I fixed this issue for now. 
I see that the same problems can happen with other functions in IIB. The idea from @lcarva is "we could have our own little policy around it, if it's more than 2 parameters, use the keyword". What do you think?

lipoja (approved):

MartinBasti:
> Update: After I talked to @lcarva about this issue, I understood that removing `source_from_index` was intentional. Unfortunately, if the `overwrite_from_index_token` is not mapped in _opm_index_add() correctly it function can misunderstand if with `source_from_index`. I fixed this issue for now.
> I see that the same problems can happen with other functions in IIB. The idea from @lcarva is "we could have our own little policy around it, if it's more than 2 parameters, use the keyword". What do you think?

Do you mean what I already proposed with `*` param syntax?

pbortlov:
> > Update: After I talked to @lcarva about this issue, I understood that removing `source_from_index` was intentional. Unfortunately, if the `overwrite_from_index_token` is not mapped in _opm_index_add() correctly it function can misunderstand if with `source_from_index`. I fixed this issue for now.
> > I see that the same problems can happen with other functions in IIB. The idea from @lcarva is "we could have our own little policy around it, if it's more than 2 parameters, use the keyword". What do you think?
> 
> Do you mean what I already proposed with `*` param syntax?

@lcarva Mention that we will run it first through the guild.

lcarva (approved):

PR #215, opened 2021-02-23
Stop setting content-encoding on AMQP messages

mprahl:
This was errantly set as the value should be like the
HTTP Content-Encoding header:

* http://docs.oasis-open.org/amqp/core/v1.0/os/amqp-core-messaging-v1.0-os.html#type-properties
* https://www.iana.org/assignments/http-parameters/http-parameters.xml

### DIFF
--- a/iib/web/messaging.py
+++ b/iib/web/messaging.py
@@ -244,7 +244,6 @@ def json_to_envelope(address, content, properties=None):
     """
     message = proton.Message(body=json.dumps(content), properties=properties)
     message.id = str(uuid.uuid4())
-    message.content_encoding = 'utf-8'
     message.content_type = 'application/json'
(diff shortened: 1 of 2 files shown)


lcarva:
@mprahl, how can we verify that this is now working as expected? Is there a current issue that we can reproduce?

mprahl:
@lcarva what we were previously doing wasn't causing any issues since no consumers used that field, but I figured I'd remove it since it was wrong.

lcarva (approved):

lcarva:
> @lcarva what we were previously doing wasn't causing any issues since no consumers used that field, but I figured I'd remove it since it was wrong.

Ack. I tested this out locally and didn't see any issues when processing a simple add request.

PR #216, opened 2021-02-25
Refresh after db.session.commit()

midnightercz:
In order to resolve qlalchemy.orm.exc.ObjectDeletedError which is raised
when accesing the object after db.session.commit, in specific cases
db.session.refresh() is called on db models

### DIFF
--- a/iib/web/api_v1.py
+++ b/iib/web/api_v1.py
@@ -313,6 +313,7 @@ def add_bundles():
     request = RequestAdd.from_json(payload)
     db.session.add(request)
     db.session.commit()
+    db.session.refresh(request)
     messaging.send_message_for_state_change(request, new_batch_msg=True)
 
     overwrite_from_index = _should_force_overwrite() or payload.get('overwrite_from_index')
@@ -438,6 +439,7 @@ def patch_request(request_id):
         request.distribution_scope = payload['distribution_scope']
 
     db.session.commit()
+    db.session.refresh(request)
 
     if state_updated:
         messaging.send_message_for_state_change(request)
@@ -468,6 +470,7 @@ def rm_operators():
     request = RequestRm.from_json(payload)
     db.session.add(request)
     db.session.commit()
+    db.session.refresh(request)
     messaging.send_message_for_state_change(request, new_batch_msg=True)
 
     overwrite_from_index = _should_force_overwrite() or payload.get('overwrite_from_index')
@@ -506,6 +509,7 @@ def regenerate_bundle():
     request = RequestRegenerateBundle.from_json(payload)
     db.session.add(request)
     db.session.commit()
+    db.session.refresh(request)
     messaging.send_message_for_state_change(request, new_batch_msg=True)
 
     error_callback = failed_request_callback.s(request.id)
@@ -554,6 +558,8 @@ def regenerate_bundle_batch():
         requests.append(request)
 
     db.session.commit()
+    for request in requests:
+        db.session.refresh(request)
     messaging.send_messages_for_new_batch_of_requests(requests)
 
     request_jsons = []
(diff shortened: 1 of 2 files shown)


lcarva:
The [docs](https://docs.sqlalchemy.org/en/13/orm/exceptions.html#sqlalchemy.orm.exc.ObjectDeletedError) for this exception state that this error is raised when a refresh operation failed. It also appears to indicate that the record was modified outside the ORMs control and may have a new primary key. How's doing another refresh going to fix the issue?

midnightercz:
> The [docs](https://docs.sqlalchemy.org/en/13/orm/exceptions.html#sqlalchemy.orm.exc.ObjectDeletedError) for this exception state that this error is raised when a refresh operation failed. It also appears to indicate that the record was modified outside the ORMs control and may have a new primary key. How's doing another refresh going to fix the issue?
I thought exception is raised by requesting attribute which is expired and therefore refreshing whole object explicitly could reload the object (with potential pk changes as well). Otherwise I have no idea what could help to prevent the problem

lcarva:
> > The [docs](https://docs.sqlalchemy.org/en/13/orm/exceptions.html#sqlalchemy.orm.exc.ObjectDeletedError) for this exception state that this error is raised when a refresh operation failed. It also appears to indicate that the record was modified outside the ORMs control and may have a new primary key. How's doing another refresh going to fix the issue?
> > I thought exception is raised by requesting attribute which is expired and therefore refreshing whole object explicitly could reload the object (with potential pk changes as well). Otherwise I have no idea what could help to prevent the problem

Your reasoning makes perfect sense to me. The docs are super vague on this. There are only two places on the sqlalchemy code that would raise this error. One of them appears to indicate that this may happen if the refresh operation timed out: https://github.com/sqlalchemy/sqlalchemy/blob/master/lib/sqlalchemy/orm/loading.py#L1411-L1414

I think refreshing again is a way to do a retry for such cases, is that right?

I wish there was a better way to do this, but I can't think of one.

lcarva (approved):

yashvardhannanavati:
@midnightercz could you please add it to the tests too?

mprahl:
I'm not sure that this will help any since refresh expires the object and then performs a query to get the database, but this happens implicitly after a commit and then accessing an attribute on the object. Read the expire documentation to see what I mean:
https://docs.sqlalchemy.org/en/13/orm/session_api.html#sqlalchemy.orm.session.Session.expire

midnightercz:
> @midnightercz could you please add it to the tests too?
if you mean add it after commit() calls in tests, I can do that but I didn't find it useful

midnightercz:
> Perhaps calling refresh in a loop until it succeeds is what you intended?

I think refresh always succeed. Problem happened in reading non-refreshed model

mprahl:
> > Perhaps calling refresh in a loop until it succeeds is what you intended?
> 
> I think refresh always succeed. Problem happened in reading non-refreshed model

From my understanding of the docs, the refresh was occurring implicitly when trying to access a column on the object after a commit. I would suggest turning on debug logs of SQLAlchemy and trying it in the Python REPL to see if that is correct. If it is, then this code would just lead to the object being refreshed sooner and thus, lead to the same problem that you are trying to solve.

midnightercz:
> From my understanding of the docs, the refresh was occurring implicitly when trying to access a column on the object after a commit. I would suggest turning on debug logs of SQLAlchemy and trying it in the Python REPL to see if that is correct. If it is, then this code would just lead to the object being refreshed sooner and thus, lead to the same problem that you are trying to solve.

So I did some testing and it looks like refresh() doesn't immediately do anything but it rather invalidates the model. I wasn't able to notice difference in called SQL commands with explicit refresh() and without it

PR #222, opened 2021-03-02
Fix index_image_resolved bug for  merge-index-image endpoint

yashvardhannanavati:


lipoja (approved):

tests/test_workers/test_tasks/test_build.py
```diff
@@ -120,11 +123,15 @@ def test_update_index_image_pull_spec(
         overwrite,
         overwrite_token,
         resolved_prebuild_from_index=resolved_from_index,
+        add_or_rm=add_or_rm,
```
lipoja, line 126: I am just thinking if there should be also test when `add_or_rm` is not set at all (testing default behavior).

yashvardhannanavati: Added a branch to test the default behavior :)

lipoja (approved):

pbortlov:
@yashvardhannanavati could you add a description?

PR #223, opened 2021-03-09
Fix skopeo verification of bundle image to retry when mediattype is none

zanssa:
Sometimes, skopeo inspect cmd succeeds but returns mediaType as none.
This is to retry the cmd for 3 times when this happens to ensure mediaType is retrieved and the bundle is resolved.

CLOUDDST-4590

### DIFF
--- a/iib/workers/tasks/utils.py
+++ b/iib/workers/tasks/utils.py
@@ -102,7 +102,9 @@ def get_resolved_bundles(bundles):
     log.info('Resolving bundles %s', ', '.join(bundles))
     resolved_bundles = set()
     for bundle_pull_spec in bundles:
-        skopeo_raw = skopeo_inspect(f'docker://{bundle_pull_spec}', '--raw')
+        skopeo_raw = skopeo_inspect(
+            f'docker://{bundle_pull_spec}', '--raw', require_media_type=True
+        )
         if (
             skopeo_raw.get('mediaType')
             == 'application/vnd.docker.distribution.manifest.list.v2+json'
@@ -299,15 +301,18 @@ def set_registry_token(token, container_image):
 @dogpile_cache(
     dogpile_region=dogpile_cache_region, should_use_cache_fn=skopeo_inspect_should_use_cache
 )
-def skopeo_inspect(*args, return_json=True):
+def skopeo_inspect(*args, return_json=True, require_media_type=False):
     """
     Wrap the ``skopeo inspect`` command.
 
     :param args: any arguments to pass to ``skopeo inspect``
     :param bool return_json: if ``True``, the output will be parsed as JSON and returned
+    :param bool require_media_type: if ``True``, ``mediaType`` will be checked in the output
+        and it will be ignored when ``return_json`` is ``False``
     :return: a dictionary of the JSON output from the skopeo inspect command
     :rtype: dict
-    :raises IIBError: if the command fails
+    :raises IIBError: if the command fails and if ``mediaType`` is not found in the output while
+        ``require_media_type`` is ``True``
     """
     exc_msg = None
     for arg in args:
@@ -318,10 +323,14 @@ def skopeo_inspect(*args, return_json=True):
     skopeo_timeout = get_worker_config().iib_skopeo_timeout
     cmd = ['skopeo', '--command-timeout', skopeo_timeout, 'inspect'] + list(args)
     output = run_cmd(cmd, exc_msg=exc_msg)
-    if return_json:
-        return json.loads(output)
+    if not return_json:
+        return output
 
-    return output
+    json_output = json.loads(output)
+
+    if require_media_type and not json_output.get('mediaType'):
+        raise IIBError('mediaType not found')
+    return json_output
 
 
 @retry(wait_on=IIBError, logger=log)
--- a/tests/test_workers/test_tasks/test_utils.py
+++ b/tests/test_workers/test_tasks/test_utils.py
@@ -473,6 +473,14 @@ def test_get_resolved_bundles_failure(mock_si):
         utils.get_resolved_bundles(['some_bundle@some_sha'])
 
 
+@mock.patch('iib.workers.tasks.utils.run_cmd')
+def test_rasise_exception_on_none_mediatype_skopeo_inspect(mock_run_cmd):
+    mock_run_cmd.return_value = '{"Name": "some-image"}'
+    image = 'docker://some-image:latest'
+    with pytest.raises(IIBError, match='mediaType not found'):
+        utils.skopeo_inspect(image, '--raw', require_media_type=True)
+
+
 @pytest.mark.parametrize(
     'pull_spec, expected',
     (


zanssa:
working on adding unit test to verify the change!

iib/workers/tasks/utils.py
```diff
@@ -326,7 +319,12 @@ def skopeo_inspect(*args, return_json=True):
     cmd = ['skopeo', '--command-timeout', skopeo_timeout, 'inspect'] + list(args)
     output = run_cmd(cmd, exc_msg=exc_msg)
     if return_json:
-        return json.loads(output)
+        output_json = json.loads(output)
+        if output_json.get('mediaType') == None:
```
lipoja, line 323: Is `mediaType` field in every output of `skopeo inspect`?

TIP: `None` type is better compare using `is` operator `if output_json.get('mediaType') is None`.

lipoja: I guess `mediaType` does not have to be present in output of `skopeo inspect`.
Therefore the above code will not work properly, because get() will always return `None` when `mediaType` is not in output.

lipoja: And other question is: should we focus only on case when we return json? Shouldn't the retry be always when `mediaType` is `None`?

iib/workers/tasks/utils.py
```diff
@@ -319,7 +319,12 @@ def skopeo_inspect(*args, return_json=True):
...
```
pbortlov, line 323: I think `None` doesn't have to be there.

iib/workers/tasks/utils.py
```diff
@@ -319,7 +319,12 @@ def skopeo_inspect(*args, return_json=True):
...
+            error_msg = f'mediaType for {arg} is None'
```
pbortlov, line 324: Can be 324 and 325 merged together?

iib/workers/tasks/utils.py
pbortlov, line 323: I think you can get mediaType only if you use `--raw` otherwise you can find it. So one option is to check that skopeo is called with `--raw` or call it again with --raw if it is not called correctly

pbortlov: the problem is not here but probably where it was reported https://github.com/release-engineering/iib/blob/master/iib/workers/tasks/utils.py#L89.

zanssa: @lipoja @pbortlov 

`mediaType` will be in the result when passing `--raw` flag in the `skopeo inspect` command, and  it's passed to the command in the `get_resolved_bundles` method. Anyway, I made some modifications as I realized I overlooked some details.

zanssa:
Added `validate_mt` boolean in `skopeo inspect` method to verify mediaType when it is needed.

As mentioned here, `skopeo inspect` returns mediaType when `--raw` option is passed in the command. In our code, the option is passed in skopeo_inspect cmd in two methods which are `get_resolved_image` to check the `schemaVersion` and `get_resolved_bundles` to check `mediaType` & `schemaVersion`. That being said, we only need to validate this in `get_resolved_bundles` method.

zanssa:
@midnightercz, PTAL 

I changed where it validates `mediaType`. So, instead of verifying that mediaType isn't missing in get_resolved_bundles method as we talked in gchat, I have it in the skopeo_inspect method. The reason that made me change where it should be validated is that if I have it in get_resolved_bundles and it raises an IIBError for one bundle, it will redo or retry skopeo inspect for all the bundles in the list, which we don't want to. Recognized this while chatting with @lcarva last week.

iib/workers/tasks/utils.py
```diff
@@ -318,10 +318,13 @@ def skopeo_inspect(*args, return_json=True):
...
     cmd = ['skopeo', '--command-timeout', skopeo_timeout, 'inspect'] + list(args)
     output = run_cmd(cmd, exc_msg=exc_msg)
-    if return_json:
+    if return_json and not validate_mt:
         return json.loads(output)
-
-    return output
+    elif return_json and validate_mt:
+        if json.loads(output).get('mediaType') is None:
+            raise IIBError('mediaType is None')
```
pbortlov, line 325: I miss `else: return json.loads(output)`

pbortlov (changes requested):

iib/workers/tasks/utils.py
```diff
@@ -299,7 +299,7 @@ def set_registry_token(token, container_image):
 @dogpile_cache(
     dogpile_region=dogpile_cache_region, should_use_cache_fn=skopeo_inspect_should_use_cache
 )
-def skopeo_inspect(*args, return_json=True):
+def skopeo_inspect(*args, return_json=True, validate_mt=False):
```
pbortlov, line 302: Would it be possible to check a list of args to define if there is `--raw` argument? This way you need to change only this function. WDYT?

zanssa: I think it is just simpler to have this boolean as return_json as we only need to check this in get_resolved_bundles and --raw option is only passed in this method and get_resolved_image.

tests/test_workers/test_tasks/test_utils.py
```diff
@@ -473,6 +473,18 @@ def test_get_resolved_bundles_failure(mock_si):
...
 
 
+@mock.patch('iib.workers.tasks.utils.skopeo_inspect')
+def test_rasise_exception_on_none_mediatype_skopeo_inspect(mock_si):
+    skopeo_inspect_rv = {
+        'mediaType': 'None',
+    }
+    mock_si.return_value = skopeo_inspect_rv
+    with pytest.raises(IIBError):
+        utils.get_resolved_bundles(['some_bundle@some_sha'])
```
lcarva, line 483: The test in this file should be isolated to calling `skopeo_inspect` directly.

iib/workers/tasks/utils.py
```diff
@@ -102,7 +102,7 @@ def get_resolved_bundles(bundles):
     log.info('Resolving bundles %s', ', '.join(bundles))
     resolved_bundles = set()
     for bundle_pull_spec in bundles:
-        skopeo_raw = skopeo_inspect(f'docker://{bundle_pull_spec}', '--raw')
+        skopeo_raw = skopeo_inspect(f'docker://{bundle_pull_spec}', '--raw', validate_mt=True)
```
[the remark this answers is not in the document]
zanssa: huh. I was trying to avoid a long name. Changed to require_media_type

README.md
```diff
@@ -59,6 +59,11 @@ If you'd like to run a specific unit test, you can do the following:
 ```bash
 tox -e py37 tests/test_web/test_api_v1.py::test_add_bundle_invalid_param
 ```
+In case you get an error that says `fatal error: gssapi/gssapi.h: No such file` while trying to run the Unit Tests, try installing the `krb5-devel` package in your system
```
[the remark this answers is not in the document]
zanssa: Noted.

README.md
```diff
@@ -59,6 +59,11 @@ If you'd like to run a specific unit test, you can do the following:
...
+
+```bash
+sudo dnf install krb5-devel
```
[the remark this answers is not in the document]
zanssa: will create another pull request!

zanssa:
@yashvardhannanavati, PTAL :)

midnightercz (approved):

pbortlov (approved):

lcarva (approved):

lcarva:
Just waiting for commits to be squashed and PR rebased before merging

PR #227, opened 2021-03-17
Add support for private registry pull secrets

lkolacek:


iib/workers/tasks/build.py
```diff
@@ -952,6 +952,20 @@ def get_image_label(pull_spec, label):
     return get_image_labels(pull_spec).get(label)
 
 
+def create_dockerconfig(registry_auths):
+    home = os.path.expanduser('~')
+    dockerconfig_path = os.path.join(home, '.docker/config.json')
```
lkolacek, line 957: @lcarva do we want to replace it for https://github.com/release-engineering/iib/blob/master/iib/workers/config.py#L15 ??

iib/web/static/api_v1.yaml
```diff
@@ -752,6 +752,11 @@ components:
             index images that IIB does not have access to. The format of the token is in the format
             "user:password".
           example: token
+        registry_auths:
```
lkolacek, line 755: @lcarva which iib operations should have registry_auths added?  Add operator bundle, remove operator, anything else?

iib/workers/tasks/build.py
```diff
@@ -1004,6 +1021,10 @@ def handle_add_request(
         ``cnr_token`` or ``organization`` is not specified.
     """
     _cleanup()
+
+    if registry_auths:
+        create_dockerconfig(registry_auths)
```
lkolacek, line 1026: @lcarva looks like with dockerconfig added I was able to inspect image in registry.redhat.io in local dev environment  http://pastebin.test.redhat.com/948693

iib/web/static/api_v1.yaml
```diff
@@ -805,6 +810,11 @@ components:
             non-privileged users to use overwrite_from_index. The format of the token must be in
             the format "user:password".
           example: token
+        registry_auths:
+          type: string
```
lkolacek, line 814: @lcarva I used string type for this parameter. On web it's displayed like this: 
```
{
"add_arches": [
"amd64",
"s390x"
],
"binary_image": "quay.io/operator-framework/upstream-registry-builder:v1.5.9",
"bundles": [
"quay.io/user/openshift-bundle-example-container:v1"
],
"cnr_token": "token",
"deprecation_list": [
"quay.io/user/openshift-bundle-example-container@sha256:29e4531e4625653"
],
"force_backport": false,
"from_index": "quay.io/iib-stage/iib:4",
"organization": "openshift-operators",
"overwrite_from_index": false,
"overwrite_from_index_token": "token",
"registry_auths": "{\"auths\": {\"registry.example.io\": {\"auth\": \"my-secret-token-base-64-encoded\"}}}",
"distribution_scope": "prod"
}
```
If correct registry and secret is provided in this format ^ . It works well. Let me know if yo uhave better idea how to deal with passing json.

lkolacek:
PTAL @yashvardhannanavati

iib/web/models.py
```diff
@@ -923,6 +925,7 @@ def from_json(cls, kwargs, batch=None):
                 'bundles',
                 'distribution_scope',
                 'deprecation_list',
+                'registry_auths',
```
lcarva, line 928: Is this needed? You're already specifying it in `RequestIndexImageMixin._from_json`. I think we only need to specify it in one place. If this is only to be used for `add` requests, and not `rm`, then I think it should only be listed here and not in `RequestIndexImageMixin._from_json`.

iib/web/static/api_v1.yaml
lcarva: That's a good question. It's needed in at least `regenerate-bundle`. Let's follow up with the CVP on this.

iib/web/static/api_v1.yaml
lcarva: I think it's odd to inline a json inside another json. Can we use a json object instead of string?

iib/workers/tasks/build.py
lcarva: Let's use the existing `set_registry_token` context manager: https://github.com/release-engineering/iib/blob/c8b668b95d41b9d4807a279a1b1154f2296930fe/iib/workers/tasks/utils.py

iib/web/static/api_v1.yaml
lcarva: Confirmed with them. Let's just add it to `regenerate-bundle`.

iib/workers/tasks/build.py
lkolacek: I added adjustment to `set_registry_token` then, hope this is close to what you had in mind.

lkolacek: If it looks usable it's probably worth to rename that method.

iib/web/models.py
```diff
@@ -1092,6 +1092,8 @@ def from_json(cls, kwargs, batch=None):
         if not isinstance(from_bundle_image, str):
             raise ValidationError('"from_bundle_image" must be a string')
 
+        # Remove attributes that are not stored in the database
+        request_kwargs.pop('registry_auths', None)
```
lcarva, line 1096: We should validate that if `registry_auths` is used, it conforms to the expected format.

iib/workers/tasks/build.py
```diff
@@ -1237,19 +1237,24 @@ def handle_rm_request(
...
-def handle_regenerate_bundle_request(from_bundle_image, organization, request_id):
+def handle_regenerate_bundle_request(from_bundle_image, organization,
+                                     request_id, registry_auths=None):
     """
     Coordinate the work needed to regenerate the operator bundle image.
 
     :param str from_bundle_image: the pull specification of the bundle image to be regenerated.
     :param str organization: the name of the organization the bundle should be regenerated for.
     :param int request_id: the ID of the IIB build request
+    :param str registry_auths: Provide the dockerconfig.json for authentication to private registries,
```
lcarva, line 1248: `dict` not a `str`, right?

iib/workers/tasks/utils.py
```diff
@@ -240,7 +240,7 @@ def reset_docker_config():
 
 
 @contextmanager
-def set_registry_token(token, container_image):
+def set_registry_token(token, container_image, registry_auths=None):
```
lcarva, line 243: Looking at the different conditionals in this function now, would it make sense to have a new `set_registry_auths` function that gets used by `set_registry_token`?

```python
def set_registry_token(token, container_image):
    if not token:
        ...
    if not container_image:
        ...
    registry = ImageName.parse(container_image).registry
    encoded_token = base64.b64encode(token.encode('utf-8')).decode('utf-8')
    registry_auths = {'auths': {registry: encoded_token}}
    with set_registry_auths(registry_auths):
      yield
    return


def set_registry_auths(registry_auths):
    if not registry_auths:
        ...
    docker_config_path = os.path.join(os.path.expanduser('~'), '.docker', 'config.json')
    try:                                                                                            
        log.debug('Removing the Docker config symlink at %s', docker_config_path)                      
        try:                                                                                        
            os.remove(docker_config_path)                                                           
        except FileNotFoundError:                                                                   
            log.debug('The Docker config symlink at %s does not exist', docker_config_path)         
                                                                                                    
        conf = get_worker_config()                                                                  
        if os.path.exists(conf.iib_docker_config_template):                                         
            with open(conf.iib_docker_config_template, 'r') as f:                                   
                docker_config = json.load(f)                                                        
        else:                                                                                       
            docker_config = {}                                                                      
                                                                
        registries = list(registry_auths.get('auths', {}).keys())
        log.debug('Setting the override token for the registries %s in the Docker config', registries)  

        docker_config.setdefault('auths', {})
        docker_config['auths'].update(registry_auths.get('auths', {}))                              
        with open(docker_config_path, 'w') as f:                                                    
            json.dump(docker_config, f)                                                             
                                                                                                    
        yield                                                                                       
    finally:                                                                                        
        reset_docker_config()
```

Beware that I have not tested the code above.

iib/web/models.py
lkolacek: I added check if it's a dict, not sure if we need a more sophisticated check here.

lkolacek: Do you want check if there is 'auths' on top level and then registries?

iib/workers/tasks/build.py
lkolacek: Yes, good catch

iib/workers/tasks/utils.py
lkolacek: Nice! It makes perfect sense

iib/web/models.py
lcarva: The `.dockerconfig.json` format allows for things other than auth information to be set. We should prevent those from being used. I think the most straightforward way is to raise a validation error if those are set. So, I propose you check the following:

1. `registry_auths` is of type dict
2. `regsitry_auths` contain a single key, `auths`
3. `regsitry_auths.auths` is a dict, it could be empty IMO
4. each _value_ in `auths` is a dict and must have one and only key named `auth`. The value of `auth` should be a string.

This will look verbose, but we don't really have another option. At some point we should revisit using a json schema to do this sort of thing.

iib/workers/tasks/utils.py
```diff
@@ -243,10 +243,8 @@ def reset_docker_config():
 def set_registry_token(token, container_image):
     """
     Configure authentication to the registry that ``container_image`` is from.
-
     This context manager will reset the authentication to the way it was after it exits. If
     ``token`` is falsy, this context manager will do nothing.
-
```
lcarva, line 249: Is there a particular reason why this empty line, and 2 above, were removed?

iib/web/static/api_v1.yaml
```diff
@@ -838,6 +838,14 @@ components:
             various modifications in the operator metadata embeded in the operator bundle
             image, e.g. change the container registry, or the operator package name.
           example: 'openshift-operators'
+        registry_auths:
+          type: object
+          description: >
+            Provide the dockerconfig.json for authentication to private registries
```
lcarva, line 844: Let's state here that non-auth information in the dockerconfig.json is not allowed.

lcarva (commented):
This is looking really good!

lcarva:
@lkolacek, checkout the output here: https://travis-ci.com/github/release-engineering/iib/jobs/495018377
Run `tox` locally to reproduce the linting errors.

iib/workers/tasks/utils.py
lkolacek: nope were removed accidently

iib/web/static/api_v1.yaml
lkolacek: Changed

iib/web/models.py
lkolacek: I added validation check. Looks terrible and verbose. Please take a look if it looks usable @lcarva . I suppose unit test is required. :)

iib/web/models.py
```diff
@@ -1398,3 +1407,28 @@ def validate_request_params(request_params, required_params, optional_params):
...
+
+    :param dict registry_auths: User provided dockerconfig for authentication
+      to private registries
+    :raises ValidationError: if registry_auths are not in valid format
+    """
+    auths = 'auths'
+    if not isinstance(registry_auths, dict):
+        raise ValidationError('"registry_auths" must be a dict')
+    if list(registry_auths.keys()) != [auths]:
+        raise ValidationError('"registry_auths" must contain single key "auths"')
```
lcarva, line 1426: Let's use the variable `auths` in the error message too.

iib/web/models.py
```diff
@@ -1398,3 +1407,28 @@ def validate_request_params(request_params, required_params, optional_params):
...
+    if not registry_auths[auths] or not isinstance(registry_auths[auths], dict):
+        raise ValidationError('"registry_auths.auths" must be a non-empty dict')
```
lcarva, line 1428: Let's use the variable `auths` in the error message too.

iib/web/models.py
lcarva: It doesn't look that bad :)

tests/test_workers/test_tasks/test_utils.py
```diff
@@ -106,6 +106,73 @@ def test_set_registry_token(
...
 
+@pytest.mark.parametrize('config_exists', (True, False))
+@pytest.mark.parametrize('template_exists', (True, False))
+@mock.patch('os.path.expanduser')
+@mock.patch('os.remove')
+@mock.patch('os.path.exists')
+@mock.patch('iib.workers.tasks.utils.open')
+@mock.patch('iib.workers.tasks.utils.json.dump')
+@mock.patch('iib.workers.tasks.utils.reset_docker_config')
+def test_set_registry_auths(
```
lcarva, line 117: Let's add a test where we validate that the auth info from `registry_auths` overwrites whatever is in the config.

lkolacek: I added 1 extra registry which should be overwritten with provided registry_auths. Does it look sufficient? https://github.com/release-engineering/iib/pull/227/files#diff-ad30f2cc32db90c116232cb76cc707702b22936e22f90e0fd5f5d215aa984ec9R134

iib/web/models.py
lkolacek: Variable added yo err msg.

iib/web/models.py
lkolacek: Variable added yo err msg.

lcarva (approved):

iib/web/models.py
```diff
@@ -1398,3 +1407,28 @@ def validate_request_params(request_params, required_params, optional_params):
...
+    """
+    auths = 'auths'
+    if not isinstance(registry_auths, dict):
+        raise ValidationError('"registry_auths" must be a dict')
+    if list(registry_auths.keys()) != [auths]:
+        raise ValidationError(f'"registry_auths" must contain single key "{auths}"')
+    if not registry_auths[auths] or not isinstance(registry_auths[auths], dict):
+        raise ValidationError(f'"registry_auths.{auths}" must be a non-empty dict')
+    for reg, auth_dict in registry_auths[auths].items():
+        err_msg = f'{reg} in registry_auths has auth value in incorrect format'
```
yashvardhannanavati, line 1430: How about changing this to `{reg} in registry_auths has auth value in an incorrect format. It should be a base64 encoded token`?

yashvardhannanavati (approved):

iib/web/models.py
lcarva: It's more complex than that. It's well described in the API docs. How about something like `{reg} in registry_auths has auth value in an incorrect format. See the API docs for details on the expected format` ?

lkolacek: I added second option `See the API docs for details on the` ..

lkolacek:
@yashvardhannanavati  @lcarva Once your happy with changes, please merge this PR. I don't have permissions.

yashvardhannanavati:
I'm so sorry. My github UI was outdated and I accidentally hit the merge button.

yashvardhannanavati:
@lkolacek Could you please address Paya's comments in a separate PR? Or would you like me to revert the change? 
@pbortlov If your comments have major changes, I can revert the commit

lkolacek:
@yashvardhannanavati  np I'll address comments in separate PR

iib/workers/tasks/build.py
```diff
@@ -1237,68 +1238,76 @@ def handle_rm_request(
...
-    log.debug('Set to regenerate the bundle image for the following arches: %s', arches_str)
-
-    payload = {
-        'from_bundle_image_resolved': from_bundle_image_resolved,
-        'state': 'in_progress',
-        'state_reason': f'Regenerating the bundle image for the following arches: {arches_str}',
-    }
-    exc_msg = 'Failed setting the resolved "from_bundle_image" on the request'
-    update_request(request_id, payload, exc_msg=exc_msg)
+    with set_registry_auths(registry_auths):
```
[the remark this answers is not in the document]
lkolacek: set_registry_auths has contextmanager decorator and it applies changes to dockerconfig.json only when registry_auths were provided. See https://github.com/release-engineering/iib/pull/227/files#diff-1bece9798c37acfe1ca705657d12580325f65a8edd965d6c3d004c7fa3237b40R290  and https://docs.python.org/3/library/contextlib.html

PR #229, opened 2021-04-06
Create Empty Index db model [CLOUDDST-4920]

pbortlov:
* add database model RequestCreateEmptyIndex

https://projects.engineering.redhat.com/browse/CLOUDDST-4920

### DIFF
--- a/iib/web/migrations/versions/e16a8cd2e028_add_create_empty_index.py
+++ b/iib/web/migrations/versions/e16a8cd2e028_add_create_empty_index.py
@@ -1,0 +1,43 @@
+"""Add RequestCreateEmptyIndex model.
+
+Revision ID: e16a8cd2e028
+Revises: 9d60d35786c1
+Create Date: 2021-04-29 12:04:28.272171
+
+"""
+from alembic import op
+import sqlalchemy as sa
+
+
+# revision identifiers, used by Alembic.
+revision = 'e16a8cd2e028'
+down_revision = '9d60d35786c1'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    op.create_table(
+        'request_create_empty_index',
+        sa.Column('id', sa.Integer(), autoincrement=False, nullable=False),
+        sa.Column('from_index_id', sa.Integer(), nullable=False),
+        sa.Column('binary_image_id', sa.Integer(), nullable=True),
+        sa.Column('labels', sa.Text(), nullable=True),
+        sa.Column('binary_image_resolved_id', sa.Integer(), nullable=True),
+        sa.Column('from_index_resolved_id', sa.Integer(), nullable=True),
+        sa.Column('index_image_id', sa.Integer(), nullable=True),
+        sa.Column('index_image_resolved_id', sa.Integer(), nullable=True),
+        sa.Column('distribution_scope', sa.String(), nullable=True),
+        sa.ForeignKeyConstraint(['binary_image_id'], ['image.id'],),
+        sa.ForeignKeyConstraint(['binary_image_resolved_id'], ['image.id'],),
+        sa.ForeignKeyConstraint(['from_index_id'], ['image.id'],),
+        sa.ForeignKeyConstraint(['from_index_resolved_id'], ['image.id'],),
+        sa.ForeignKeyConstraint(['id'], ['request.id'],),
+        sa.ForeignKeyConstraint(['index_image_id'], ['image.id'],),
+        sa.ForeignKeyConstraint(['index_image_resolved_id'], ['image.id'],),
+        sa.PrimaryKeyConstraint('id'),
+    )
+
+
+def downgrade():
+    op.drop_table('request_create_empty_index')
--- a/iib/web/models.py
+++ b/iib/web/models.py
@@ -70,6 +70,7 @@ class RequestTypeMapping(BaseEnum):
     rm = 2
     regenerate_bundle = 3
     merge_index_image = 4
+    create_empty_index = 5
 
     @classmethod
     def pretty(cls, num):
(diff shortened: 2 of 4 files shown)


iib/web/models.py
```diff
@@ -911,6 +912,7 @@ def from_json(cls, kwargs, batch=None):
         if not isinstance(request_kwargs.get('force_backport', False), bool):
             raise ValidationError('"force_backport" must be a boolean')
 
+        # Why you check crn_token and force_backport above and remove it here?
```
pbortlov, line 915: I went through the add bundles but I don't understand why it is used this way @yashvardhannanavati

iib/web/models.py
```diff
@@ -1398,3 +1400,106 @@ def validate_request_params(request_params, required_params, optional_params):
...
+                not item or not isinstance(item, str) for item in labels
+        ):
+            raise ValidationError('"labels" should be either an empty array or an array of non-empty strings')
+
+
+        # Remove attributes that are not stored in the database
+        request_kwargs.pop('cnr_token', None)
+        request_kwargs.pop('force_backport', None)
+
+        # Should I remove all things which don't belong here?
```
pbortlov, line 1449: I am not sure what variables we need and I should remove

iib/web/models.py
```diff
@@ -1398,3 +1400,106 @@ def validate_request_params(request_params, required_params, optional_params):
...
+
+        labels = request_kwargs.get('labels', [])
+        if not isinstance(labels, list) or any(
+                not item or not isinstance(item, str) for item in labels
+        ):
+            raise ValidationError('"labels" should be either an empty array or an array of non-empty strings')
+
+
+        # Remove attributes that are not stored in the database
+        request_kwargs.pop('cnr_token', None)
```
pbortlov, line 1446: What is it used for?

iib/web/models.py
```diff
@@ -1398,3 +1400,106 @@ def validate_request_params(request_params, required_params, optional_params):
...
+        # Should I remove all things which don't belong here?
+        # How can I know which doesn't belong
+        request_kwargs.pop('add_arches', None)
+        request_kwargs.pop('overwrite_from_index', None)
+        request_kwargs.pop('overwrite_from_index_token', None)
+        request_kwargs.pop('distribution_scope', None)
+
+        cls._from_json(
+            request_kwargs,
+            additional_optional_params=[
```
pbortlov, line 1458: I am not sure if theses args are additional_optional or additional_required

iib/web/models.py
```diff
@@ -1398,3 +1400,106 @@ def validate_request_params(request_params, required_params, optional_params):
...
+                'labels'
+            ],
+            batch=batch,
+        )
+
+        request = cls(**request_kwargs)
+        request.add_state('in_progress', 'The request was initiated')
+        return request
+
+    def to_json(self, verbose=True):
```
pbortlov, line 1469: These are copy-paste without change

iib/web/models.py
```diff
@@ -1398,3 +1400,106 @@ def validate_request_params(request_params, required_params, optional_params):
...
+                rv['bundle_mapping'].setdefault(bundle.operator.name, []).append(
+                    bundle.pull_specification
+                )
+            rv['bundles'].append(bundle.pull_specification)
+
+        rv['deprecation_list'] = [bundle.pull_specification for bundle in self.deprecation_list]
+
+        return rv
+
+    def get_mutable_keys(self):
```
pbortlov, line 1495: These are copy-paste without change

iib/web/models.py
```diff
@@ -895,7 +896,7 @@ def from_json(cls, kwargs, batch=None):
                 raise ValidationError(
                     f'"{key}" should be either an empty array or an array of non-empty strings'
                 )
-
+        # TODO check this out
         # Check if no bundles and `from_index is specified
         # if no bundles and no from index then an empty index will be created which is a no-op
```
pbortlov, line 901: The comment doesn't make sense. For creation of empty index we still need from_index

iib/web/models.py
lipoja: We do not want to store sensitive information (token in this case)

iib/web/models.py
lipoja: Sensitive information should not be stored in DB directly. 
Not sure about `force_backport`.

pbortlov: thanks

iib/web/models.py
```diff
@@ -1435,3 +1435,104 @@ def validate_registry_auths(registry_auths):
...
+        """
+        rv = super().to_json(verbose=verbose)
+        rv.update(self.get_common_index_image_json())
+
+        # rv.pop('deprecation_list')
+        # rv.pop('remove_operators')
+        # rv.pop('bundles')
+        # rv.pop('bundle_mapping')
+        # rv.pop('organization')
+        # rv.pop('distribution_scope')
```
lipoja, line 1520: Let's remove these lines since they are not used.

pbortlov: removed

midnightercz (approved):

iib/web/models.py
```diff
@@ -895,7 +896,6 @@ def from_json(cls, kwargs, batch=None):
                 raise ValidationError(
                     f'"{key}" should be either an empty array or an array of non-empty strings'
                 )
-
```
yashvardhannanavati, line 898: removed by accident?

iib/web/models.py
```diff
@@ -789,7 +790,7 @@ def _from_json(
 
         # Validate binary_image is correctly provided
         binary_image = request_kwargs.pop('binary_image', None)
-        if binary_image is not None and not isinstance(binary_image, str):
+        if binary_image and not isinstance(binary_image, str):
```
yashvardhannanavati, line 793: This will not throw an error if binary_image is passed as `False`. Could you please change it back to what it was?

iib/web/models.py
```diff
@@ -1435,3 +1435,97 @@ def validate_registry_auths(registry_auths):
...
+        """
+        Handle JSON requests for the create-empty-index API endpoint.
+
+        :param dict kwargs: the JSON payload of the request.
+        :param Batch batch: the batch to specify with the request.
+        """
+        request_kwargs = deepcopy(kwargs)
+
+        if not request_kwargs.get('from_index'):
+            raise ValidationError('"from_index" must be specified')
```
yashvardhannanavati, line 1483: `"from_index" must be specified"` => `"from_index" must be a non-empty string`

iib/web/models.py
```diff
@@ -1435,3 +1435,97 @@ def validate_registry_auths(registry_auths):
...
+
+        if request_kwargs.get('from_index') and not isinstance(
```
yashvardhannanavati, line 1485: `if request_kwargs.get('from_index') is not `None` and not ...` ? This way you can get rid of #1482

yashvardhannanavati: On second thoughts, you don't really need this validation as the base class `from_json` will check for validity of `from_index`.

You can simply change line 1482 to check if `from_index` is not None to make sure your code doesn't falter on line 748.

Also, you might have to add a check here to make sure the user doesn't provide `add_arches`, `overwrite_from_index` and `overwrite_from_index_token`. Otherwise, IIB will process them as optional params and consider them valid because of the base class `from_json` function.

iib/web/models.py
```diff
@@ -1435,3 +1435,97 @@ def validate_registry_auths(registry_auths):
...
+        """
+        Return the set of keys representing the attributes that can be modified.
+
+        :return: a set of key names
+        :rtype: set
+        """
+        rv = super().get_mutable_keys()
+        rv.update(self.get_index_image_mutable_keys())
+        rv.update('labels')
+        rv.update('from_index')
```
yashvardhannanavati, line 1529: line 1527 will already add `index_image` and `binary_image`. You can remove it from here

yashvardhannanavati (changes requested):

iib/web/models.py
pbortlov: yes

iib/web/models.py
pbortlov: done

iib/web/models.py
pbortlov: done

iib/web/models.py
pbortlov: Should it be an error or I pop it?

pbortlov: done

lipoja (approved):

iib/web/models.py
```diff
@@ -1435,3 +1436,101 @@ def validate_registry_auths(registry_auths):
...
+        if not isinstance(labels, dict):
+            raise ValidationError('The value of "labels" must be a JSON object')
+
+        for arg in (
+            'add_arches',
+            'overwrite_from_index',
+            'overwrite_from_index_token',
+            'distribution_scope',
+        ):
+            if request_kwargs.get(arg) is not None:
```
yashvardhannanavati, line 1496: `if arg in request_kwargs:`? We want to throw an error irrespective of the value right.

yashvardhannanavati (changes requested):

iib/web/migrations/versions/e16a8cd2e028_add_create_empty_index.py
```diff
@@ -0,0 +1,43 @@
...
+        'request_create_empty_index',
+        sa.Column('id', sa.Integer(), autoincrement=False, nullable=False),
+        sa.Column('from_index_id', sa.Integer(), nullable=False),
+        sa.Column('binary_image_id', sa.Integer(), nullable=True),
+        sa.Column('labels', sa.Text(), nullable=True),
+        sa.Column('binary_image_resolved_id', sa.Integer(), nullable=True),
+        sa.Column('from_index_resolved_id', sa.Integer(), nullable=True),
+        sa.Column('index_image_id', sa.Integer(), nullable=True),
+        sa.Column('index_image_resolved_id', sa.Integer(), nullable=True),
+        sa.Column('distribution_scope', sa.String(), nullable=True),
```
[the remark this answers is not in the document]
pbortlov: This one is created by RequestIndexImageMixin. Should we really remove it?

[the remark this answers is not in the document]
pbortlov: @yashvardhannanavati ping

iib/web/models.py
```diff
@@ -1435,3 +1436,115 @@ def validate_registry_auths(registry_auths):
...
+class RequestCreateEmptyIndex(Request, RequestIndexImageMixin):
+    """An "create-empty-index" image build request."""
+
+    __tablename__ = 'request_create_empty_index'
+
+    id = db.Column(db.Integer, db.ForeignKey('request.id'), autoincrement=False, primary_key=True)
+    from_index_id = db.Column(db.Integer, db.ForeignKey('image.id'), nullable=False)
+    from_index = db.relationship('Image', foreign_keys=[from_index_id], uselist=False)
+    binary_image_id = db.Column(db.Integer, db.ForeignKey('image.id'), nullable=True)
+    binary_image = db.relationship('Image', foreign_keys=[binary_image_id], uselist=False)
```
yashvardhannanavati, line 1450: You don't have to define from_index and binary_image here right since you're inheriting the RequestIndexImageMixin class

iib/web/models.py
```diff
@@ -1435,3 +1436,115 @@ def validate_registry_auths(registry_auths):
...
+        """
+        request_kwargs = deepcopy(kwargs)
+
+        if request_kwargs.get('from_index') is None:
+            raise ValidationError('"from_index" must be a non-empty string')
+
+        labels = request_kwargs.pop('labels', {})
+        if not isinstance(labels, dict):
+            raise ValidationError('The value of "labels" must be a JSON object')
+        for label in labels:
```
yashvardhannanavati, line 1489: You should check if both key and value are strings.

iib/web/migrations/versions/e16a8cd2e028_add_create_empty_index.py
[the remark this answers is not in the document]
yashvardhannanavati: Ah, ok. In that case, we can probably keep it. But if it's specified in labels, let's make sure we update this column too.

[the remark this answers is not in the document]
yashvardhannanavati: Ah, I see. In that case, let's keep it. Also, let's make sure the distribution_scope value is updated when `labels` dict has that label

iib/web/models.py
```diff
@@ -1435,3 +1436,107 @@ def validate_registry_auths(registry_auths):
...
+            if not (isinstance(key, str) or isinstance(value, str)):
+                raise ValidationError(f'The key and value of {key} must be a string')
+
+        for arg in (
+            'add_arches',
+            'overwrite_from_index',
+            'overwrite_from_index_token',
+        ):
+            if arg in request_kwargs:
+                raise ValidationError(f'The {arg} is not allowed to use.')
```
yashvardhannanavati, line 1495: `The {arg} is not allowed to use.` => `The "{arg}" arg is invalid for the create-empty-index endpoint.`

yashvardhannanavati (approved):
One minor comment. Other than that, it looks good. Let's wait for the API PR to be ready before merging this? In-case we make changes to the schema again.

pbortlov:
CLosing in favor of #250

PR #230, opened 2021-04-08
Filter out unique bundles from grpcurl listBundles response

yashvardhannanavati:
Refers to CLOUDDST-5783


pbortlov:
@yashvardhannanavati Could you create the branch in your fork?

iib/workers/tasks/build.py
```diff
@@ -426,11 +427,22 @@ def _get_present_bundles(from_index, base_dir):
 
     # If no data is returned there are not bundles present
     if not bundles:
-        return []
+        return [], []
```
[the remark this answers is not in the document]
yashvardhannanavati: To avoid re-doing the same computation later when we want just the pull_specs of the labels. I figured we can create the list when we are filtering unique bundles and then use it in the rest of the code rather than doing the same computation over and over. It'll save a few milliseconds in every build.

iib/workers/tasks/build.py
```diff
@@ -1050,7 +1062,9 @@ def handle_add_request(
             set_request_state(request_id, 'in_progress', msg)
 
             with set_registry_token(overwrite_from_index_token, from_index_resolved):
-                present_bundles = _get_present_bundles(from_index_resolved, temp_dir)
+                present_bundles, present_bundles_pull_spec = _get_present_bundles(
```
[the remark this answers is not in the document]
yashvardhannanavati: We aren't pulling the pull_specs here if that's what you are suggesting. Could you please elaborate a bit on this comment?

yashvardhannanavati:
> @yashvardhannanavati Could you create the branch in your fork?

You're right. I should be using a fork instead of creating a branch. I'll keep that in mind from my next PR if that's okay.

iib/workers/tasks/build.py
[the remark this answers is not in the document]
pbortlov: the `present_bundles_pull_spec` is part of `with` statement but then it is used on line #1105 where it is on a different indentation level. In the same time `present_bundles` line #1057 is defined but it doesn't have to be. WHat approach are we taking here?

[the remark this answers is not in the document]
pbortlov: this explained me it a little bit https://stackoverflow.com/questions/44437694/using-python-variable-outside-with-statement

iib/workers/tasks/build.py
```diff
@@ -426,11 +427,22 @@ def _get_present_bundles(from_index, base_dir):
 
     # If no data is returned there are not bundles present
     if not bundles:
-        return []
+        return [], []
 
     # Transform returned data to parsable json
+    present_bundles = []
```
pbortlov, line 433: This is already defined line 436

pbortlov:
Q: line 1073 why do we copy values from filtered bundles to resolved `resolved_bundles = filtered_bundles`. Can i cost us problems?

iib/workers/tasks/build.py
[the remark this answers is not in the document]
pbortlov: thanks

pbortlov (approved):

iib/workers/tasks/build.py
```diff
@@ -426,11 +427,21 @@ def _get_present_bundles(from_index, base_dir):
...
-    return present_bundles
+
+    for bundle in present_bundles:
+        bundle_path = bundle['bundlePath']
+        if bundle_path in unique_present_bundles_pull_spec:
+            continue
+        unique_present_bundles.append(bundle)
+        unique_present_bundles_pull_spec.append(bundle_path)
+
+    return unique_present_bundles, unique_present_bundles_pull_spec
```
querti, line 444: Hmm, I personally find it confusing that this function returns two structures where one is the subset of the other. My preference would be to keep this part as it was. I wonder what others think.

midnightercz (approved):

iib/workers/tasks/build.py
yashvardhannanavati: The only reason to do this was to avoid re-doing the same computation later when we want just the pull_specs of the labels. I figured we can create the list when we are filtering unique bundles and then use it in the rest of the code rather than doing the same computation over and over.

yashvardhannanavati: So can I please suggest that we keep it this way

querti: Yeah, I understand the reasoning, but I feel like we are sacrificing design and readability to gain minuscule amount of performance. But I don't have strong feelings about it, I'm okay with keeping it as-is.

yashvardhannanavati: I personally think it's not too bad of a trade off since it's just this function. I'll keep it as-is for now. :)

PR #234, opened 2021-04-14
Separate out regenerate bundle image logic from build.py

yashvardhannanavati:
Refers to CLOUDDST-5856


midnightercz (approved):

iib/workers/tasks/build_regenerate_bundle.py
```diff
@@ -0,0 +1,356 @@
...
+        annotations_yaml['annotations'][package_label],
+    )
+
+    return (
+        annotations_yaml['annotations'][package_label],
+        {package_label: annotations_yaml['annotations'][package_label]},
+    )
+
+
+def _adjust_operator_bundle(manifests_path, metadata_path, organization=None, pinned_by_iib=False):
```
pbortlov, line 215: Do we want to get rid of `_` at the beginning of the name of the function?

iib/workers/tasks/build_regenerate_bundle.py
```diff
@@ -0,0 +1,356 @@
...
+    payload = {
+        'arches': list(arches),
+        'bundle_image': output_pull_spec,
+        'state': 'complete',
+        'state_reason': 'The request completed successfully',
+    }
+    update_request(request_id, payload, exc_msg='Failed setting the bundle image on the request')
+
+
+def _apply_package_name_suffix(metadata_path, organization=None):
```
pbortlov, line 139: Do we want to get rid of `_` at the beginning of the name of the function?

iib/workers/tasks/build_regenerate_bundle.py
```diff
@@ -0,0 +1,356 @@
...
+
+        operator_csv.dump()
+
+    if organization:
+        _adjust_csv_annotations(operator_manifest.files, package_name, organization)
+
+    return labels
+
+
+def _adjust_csv_annotations(operator_csvs, package_name, organization):
```
pbortlov, line 330: Do we want to get rid of `_` at the beginning of the name of the function?

pbortlov (commented):
Just minor changes, otherwise it looks great

iib/workers/tasks/build_regenerate_bundle.py
yashvardhannanavati: I don't think so because it's still a private function being used only in the same module.

iib/workers/tasks/build_regenerate_bundle.py
yashvardhannanavati: I don't think so because it's still a private function being used only in the same module.

iib/workers/tasks/build_regenerate_bundle.py
yashvardhannanavati: I don't think so because it's still a private function being used only in the same module.

PR #235, opened 2021-04-14
Ensure image is pullable before copying content

lkolacek:
CLOUDBLD-5140

Signed-off-by: Ladislav Kolacek <lkolacek@redhat.com>

### DIFF
--- a/iib/workers/tasks/build.py
+++ b/iib/workers/tasks/build.py
@@ -29,6 +29,7 @@ from iib.workers.tasks.utils import (
     get_image_labels,
     get_resolved_bundles,
     get_resolved_image,
+    podman_pull,
     request_logger,
     reset_docker_config,
     retry,
@@ -1264,6 +1265,9 @@ def _copy_files_from_image(image, src_path, dest_path):
     :param str src_path: the full path within the container image to copy from.
     :param str dest_path: the full path on the local host to copy into.
     """
+    # Check that image is pullable
+    podman_pull(image)
+
     # One way to copy a file from the image is to create a container from its filesystem
     # so the contents can be read. To create a container, podman always requires that a
     # command for the container is set. In this method, however, the command is not needed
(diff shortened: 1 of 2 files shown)


lkolacek:
PTAL @lcarva @pbortlov @yashvardhannanavati

lkolacek:
@lcarva Looks like all other operations (podman, skopeo) are covered with retries. I can add also `podman_pull` prior `_copy_files_from_image` but that does not seem to be necessary.  

From the ticket CLOUDBLD-5140
```
Analyze past failures to better understand where failures with the registry usually happen.
```
Is there a way how to list all failures that happened in IIB? So far I checked the code where we interact with registries. 

Thanks!

iib/workers/tasks/build.py
```diff
@@ -1338,6 +1338,7 @@ def handle_regenerate_bundle_request(
     update_request(request_id, payload, exc_msg='Failed setting the bundle image on the request')
 
 
+@retry(wait_on=IIBError, logger=log)
```
lcarva, line 1341: Is this too high level?

If the image cannot be pulled, the `podman create` command may fail. But it may also fail for completely unrelated reasons. 

What if right before running `podman create` we run `podman pull` inside this function? That will help us provide a better error message to the users. Wdyt?

lcarva:
> @lcarva Looks like all other operations (podman, skopeo) are covered with retries. I can add also `podman_pull` prior `_copy_files_from_image` but that does not seem to be necessary.
> 
> From the ticket CLOUDBLD-5140
> 
> ```
> Analyze past failures to better understand where failures with the registry usually happen.
> ```
> 
> Is there a way how to list all failures that happened in IIB? So far I checked the code where we interact with registries.
> 
> Thanks!

@lkolacek, you'd have to paginate through the list of failed requests by using the API.

iib/workers/tasks/build.py
lkolacek: OK, availability of registry will be checked with pull then.

lcarva (approved):

PR #240, opened 2021-04-16
Add distribution_scope and versions labels after deprecating the bundles

yashvardhannanavati:
Refers to CLOUDDST-5944

### DIFF
--- a/tests/test_workers/test_tasks/test_build.py
+++ b/tests/test_workers/test_tasks/test_build.py
@@ -807,6 +807,130 @@ def test_handle_add_request(
         mock_dep_b.assert_not_called()
 
 
+@mock.patch('iib.workers.tasks.build.deprecate_bundles')
+@mock.patch('iib.workers.tasks.utils.get_resolved_bundles')
+@mock.patch('iib.workers.tasks.build._cleanup')
+@mock.patch('iib.workers.tasks.build._verify_labels')
+@mock.patch('iib.workers.tasks.build._prepare_request_for_build')
+@mock.patch('iib.workers.tasks.build._update_index_image_build_state')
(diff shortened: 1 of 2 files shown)


lcarva:
LGTM!

It would be nice to add a unit test for this.

yashvardhannanavati:
Added a unit test to mock the label setting behavior :)

tests/test_workers/test_tasks/test_build.py
```diff
@@ -807,6 +807,130 @@ def side_effect(_, temp_dir, *args, **kwargs):
...
+    mock_capml.return_value = output_pull_spec
+    mock_gpb.return_value = [{'bundlePath': 'random_bundle@sha'}], ['random_bundle@sha']
+    bundles = ['some-bundle:2.3-1', 'some-deprecation-bundle:1.1-1']
+    cnr_token = 'token'
+    organization = 'org'
+    greenwave_config = {'some_key': 'other_value'}
+    mock_ugrb.return_value = ['random_bundle@sha', 'some-deprecation-bundle@sha']
+    deprecation_list = ['random_bundle@sha', 'some-deprecation-bundle@sha']
+    # Assume default labels are set on the index
+    os.environ['LABEL_SET'] = 'default_labels_set'
```
lcarva, line 877: Why use `os.environ` here? 

You can create a variable here and it will be available in the functions below. You just have to make sure to use a mutable data type, e.g. `dict`:
```python
label_state = {'LABEL_SET': 'default_labels_set'}
```

Then you can update `label_state['LABEL_SET']` as much as you want. It will always change the same `dict` object.

yashvardhannanavati: you're right. A mutable data type can be used indeed.

lcarva (approved):

PR #241, opened 2021-04-20
Change format of iib_organization_customizations worker config variable

yashvardhannanavati:
iib_organization_customizations now has the ability to specify the
order in which the customizations must be applied to the bundle
image.

Refers to CLOUDDST-5691

### DIFF
--- a/iib/workers/config.py
+++ b/iib/workers/config.py
@@ -70,20 +70,26 @@ class DevelopmentConfig(Config):
     iib_api_url = 'http://iib-api:8080/api/v1/'
     iib_log_level = 'DEBUG'
     iib_organization_customizations = {
-        'company-marketplace': {
-            'csv_annotations': {
-                'marketplace.company.io/remote-workflow': (
-                    'https://marketplace.company.com/en-us/operators/{package_name}/pricing'
-                ),
-                'marketplace.company.io/support-workflow': (
-                    'https://marketplace.company.com/en-us/operators/{package_name}/support'
-                ),
+        'company-marketplace': [
+            {
+                'type': 'csv_annotations',
+                'annotations': {
+                    'marketplace.company.io/remote-workflow': (
+                        'https://marketplace.company.com/en-us/operators/{package_name}/pricing'
+                    ),
+                    'marketplace.company.io/support-workflow': (
+                        'https://marketplace.company.com/en-us/operators/{package_name}/support'
+                    ),
+                },
             },
-            'package_name_suffix': '-cmp',
-            'registry_replacements': {
-                'registry.access.company.com': 'registry.marketplace.company.com/cm',
+            {'type': 'package_name_suffix', 'suffix': '-cmp'},
+            {
+                'type': 'registry_replacements',
+                'replacements': {
+                    'registry.access.company.com': 'registry.marketplace.company.com/cm',
+                },
             },
-        }
+        ]
     }
     iib_registry = 'registry:8443'
     iib_request_logs_dir = '/var/log/iib/requests'
@@ -150,48 +156,7 @@ def validate_celery_config(conf, **kwargs):
     if not isinstance(conf['iib_required_labels'], dict):
         raise ConfigError('iib_required_labels must be a dictionary')
 
-    if not isinstance(conf['iib_organization_customizations'], dict):
-        raise ConfigError('iib_organization_customizations must be a dictionary')
-
-    for org, org_config in conf['iib_organization_customizations'].items():
-        if not isinstance(org, str):
-            raise ConfigError('The keys in iib_organization_customizations must be strings')
-
-        if not isinstance(org_config, dict):
-            raise ConfigError('The values in iib_organization_customizations must be dictionaries')
-
-        invalid_keys = org_config.keys() - {
-            'csv_annotations',
-            'package_name_suffix',
-            'registry_replacements',
-        }
-        if invalid_keys:
-            raise ConfigError(
-                'The following keys set on iib_organization_customizations are '
-                f'invalid: {", ".join(sorted(invalid_keys))}'
-            )
-
-        for key in ('csv_annotations', 'registry_replacements'):
-            value = org_config.get(key)
-            if not value:
-                continue
-
-            for k, v in value.items():
-                if not isinstance(k, str):
-                    raise ConfigError(
-                        f'The keys in iib_organization_customizations.{org}.{key} must be strings'
-                    )
-
-                if not isinstance(v, str):
-                    raise ConfigError(
-                        f'The values in iib_organization_customizations.{org}.{key} must be strings'
-                    )
-
-        if not isinstance(org_config.get('package_name_suffix', ''), str):
-            raise ConfigError(
-                f'The value of iib_organization_customizations.{org}.package_name_suffix '
-                'must be a string'
-            )
+    _validate_iib_org_customizations(conf['iib_organization_customizations'])
 
     iib_request_logs_dir = conf.get('iib_request_logs_dir')
     if iib_request_logs_dir:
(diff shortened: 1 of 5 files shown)


yashvardhannanavati:
@midnightercz @lcarva @mprahl Could you PTAL?

iib/workers/config.py
```diff
@@ -153,45 +159,66 @@ def validate_celery_config(conf, **kwargs):
...
+                    f'Invalid customization in iib_organization_customizations {customization}'
+                )
+
+            invalid_customization_keys = (
+                customization.keys() - valid_customizations[customization_type] - {'type'}
             )
+            if invalid_customization_keys:
+                raise ConfigError(
+                    f'The keys {invalid_customization_keys} in iib_organization_customizations'
+                    f'.{org}[{org_config.index(customization)}] are invalid.'
```
mprahl, line 194: You can just use JSON notation and use the dot. It'd be a cleaner error message. Right now you are mixing both.

iib/workers/config.py
```diff
@@ -153,45 +159,66 @@ def validate_celery_config(conf, **kwargs):
     if not isinstance(conf['iib_organization_customizations'], dict):
```
mprahl, line 159: Could you move this check to a new private function? That would make it easier to read.

mprahl, line 159: Could you move this check to a new private function? That would make it easier to read.

iib/workers/tasks/build_regenerate_bundle.py
```diff
@@ -269,21 +285,84 @@ def _adjust_operator_bundle(manifests_path, metadata_path, organization=None, pi
...
+    """
+    annotations_yaml_path = os.path.join(metadata_path, 'annotations.yaml')
+    if not os.path.exists(annotations_yaml_path):
+        raise IIBError('metadata/annotations.yaml does not exist in the bundle')
+
+    with open(annotations_yaml_path, 'r') as f:
+        try:
+            annotations_yaml = yaml.load(f)
+        except ruamel.yaml.YAMLError:
+            error = 'metadata/annotations/yaml is not valid YAML'
```
mprahl, line 330:
```suggestion
            error = 'metadata/annotations.yaml is not valid YAML'
```

iib/workers/tasks/build_regenerate_bundle.py
```diff
@@ -152,37 +152,15 @@ def _apply_package_name_suffix(metadata_path, organization=None):
...
-
+    annotations_yaml = _get_package_annotations(metadata_path)
     package_label = 'operators.operatorframework.io.bundle.package.v1'
-    package_annotation = annotations_yaml.get('annotations', {}).get(package_label)
-    if not package_annotation:
-        raise IIBError(f'{package_label} is not set in metadata/annotations.yaml')
-
-    if not isinstance(package_annotation, str):
-        raise IIBError(f'The value of {package_label} in metadata/annotations.yaml is not a string')
+    package_annotation = annotations_yaml['annotations'][package_label]
```
mprahl, line 157: I don't see this behavior in the docstring. Could you please add it?

Without an explanation, it's confusing that `organization` can be `None`.

iib/workers/tasks/build_regenerate_bundle.py
```diff
@@ -235,17 +213,55 @@ def _adjust_operator_bundle(manifests_path, metadata_path, organization=None, pi
...
+    package_name = annotations_yaml['annotations'][
+        'operators.operatorframework.io.bundle.package.v1'
+    ]
+    labels = {}
+    if organization_customizations:
+        # Perform the customizations in order
+        for customization in organization_customizations:
+            customization_type = customization['type']
+            if customization_type == 'package_name_suffix':
+                package_name, labels = _apply_package_name_suffix(metadata_path, organization)
```
mprahl, line 236: Could you refactor `_apply_package_name_suffix`, `_get_bundle_metadata`, and `_adjust_csv_annotations` to accept `customization` rather than `organization`? Then each private function doesn't need to be aware of the `iib_organization_customizations` configuration and I think that'd be cleaner.

iib/workers/tasks/build_regenerate_bundle.py
```diff
@@ -235,17 +213,55 @@ def _adjust_operator_bundle(manifests_path, metadata_path, organization=None, pi
...
-    operator_csvs = []
+    conf = get_worker_config()
+    organization_customizations = conf['iib_organization_customizations'].get(organization, [])
+
+    annotations_yaml = _get_package_annotations(metadata_path)
+    package_name = annotations_yaml['annotations'][
+        'operators.operatorframework.io.bundle.package.v1'
+    ]
+    labels = {}
+    if organization_customizations:
```
mprahl, line 231: Optional: You could default `organization_customizations` to something like this and then you wouldn't need the if/else conditional.
```
[
    {'type': 'package_name_suffix'},
    {'type': 'registry_replacements'},
    {'type': 'csv_annotations'},
]
```

mprahl (changes requested):

MartinBasti:
with commit `Fixed after rebase` does it mean that all previous commits in PR break code?

iib/workers/tasks/build_regenerate_bundle.py
yashvardhannanavati: Yes, this is on my radar. I plan to make this change in my next PR when I introduce a new customization. Would it be okay if I did it in my next PR?

mprahl: I'd rather do it now, but I won't block my approval on it.

mprahl (approved):

yashvardhannanavati:
@mprahl Refactored the functions :)

iib/workers/tasks/build_regenerate_bundle.py
```diff
@@ -269,21 +281,62 @@ def _adjust_operator_bundle(manifests_path, metadata_path, organization=None, pi
...
+    if not package_annotation:
+        raise IIBError(f'{package_label} is not set in metadata/annotations.yaml')
+
+    if not isinstance(package_annotation, str):
+        raise IIBError(f'The value of {package_label} in metadata/annotations.yaml is not a string')
+
+    return annotations_yaml
+
+
+def _resolve_image_pull_specs(bundle_metadata, labels, pinned_by_iib, registry_replacements=None):
```
mprahl, line 326: Since you allow `None` for `registry_replacements`, could you please add this at the beginning of the function?
```python
if registry_replacements is None:
    registry_replacements = {}
```

tests/test_workers/test_tasks/test_build_regenerate_bundle.py
```diff
@@ -523,18 +685,16 @@ def test_adjust_csv_annotations(tmpdir):
...
 
-@mock.patch('iib.workers.tasks.build_regenerate_bundle.yaml.dump')
-def test_adjust_csv_annotations_no_customizations(mock_yaml_dump, tmpdir):
-    manifests_dir = tmpdir.mkdir('manifests')
-    manifests_dir.join('backup.crd.yaml').write(
-        'apiVersion: apiextensions.k8s.io/v1beta1\nkind: CustomResourceDefinition'
-    )
-    csv = manifests_dir.join('mig-operator.v1.1.1.clusterserviceversion.yaml')
-    csv.write('apiVersion: operators.coreos.com/v1alpha1\nkind: ClusterServiceVersion')
+# @mock.patch('iib.workers.tasks.build_regenerate_bundle.yaml.dump')
```
mprahl, line 688: Could you please fix this?

mprahl (changes requested):
It's looking good! Just a couple of small comments.

iib/workers/tasks/build_regenerate_bundle.py
yashvardhannanavati: It's only being called from one place and an empty dict will always be passed. Maybe I'll not have this optional at all.

tests/test_workers/test_tasks/test_build_regenerate_bundle.py
yashvardhannanavati: good catch! you still got it ;)

mprahl (approved):

PR #243, opened 2021-04-30
Add image_name_from_labels customization

yashvardhannanavati:
Refers to CLOUDDST-5692

### DIFF
--- a/README.md
+++ b/README.md
@@ -260,14 +260,19 @@ The custom configuration options for the Celery workers are listed below:
   and `registry_replacements`. The order of the dictionaries in the list will determine the order
   of customizations applied to the bundle.
 
-  * The `csv_annotations` value is a dictionary where each key is the annotation to set on the
-    ClusterServiceVersion files, and the value is a Python template string of the value to be set.
-    IIB only substitutes `{package_name}` in the template string.
-  * The `package_name_suffix` value is a string of a suffix to add to the package name of the
-    operator.
-  * The `registry_replacements` value is a dictionary where the keys are the old registries to
-    replace and the values are the registries to replace the old registries with. This replaces the
-    registry in all the ClusterServiceVersion files.
+  * The `csv_annotations` customization type is a dictionary where the key `annotations` value is
+    a dictionary where each key is the annotation to set on the ClusterServiceVersion files, and
+    the value is a Python template string of the value to be set. IIB only substitutes
+    `{package_name}` in the template string.
+  * The `package_name_suffix` customization type is a dictionary where the key `suffix` value is
+    a string of a suffix to add to the package name of the operator.
+  * The `registry_replacements` customization type is a dictionary where the key `replacements`
+    value is a dictionary where the keys are the old registries to replace and the values
+    are the registries to replace the old registries with. This replaces the registry in all
+    the ClusterServiceVersion files.
+  * The `image_name_from_labels` customization type is a dictionary where the key `template`
+    value is a string which specifies a combination of label names in curly braces which will be
+    substituted with the actual label values from the bundle image.
 
   Here is an example that ties this all together:
 
@@ -292,6 +297,7 @@ The custom configuration options for the Celery workers are listed below:
                     'registry.access.company.com': 'registry.marketplace.company.com/cm',
                 },
             },
+            {'type': 'image_name_from_labels', 'template': '{name}-{version}-final'},
         ]
     }
   ```
--- a/iib/workers/tasks/build_regenerate_bundle.py
+++ b/iib/workers/tasks/build_regenerate_bundle.py
@@ -20,6 +20,7 @@ from iib.workers.tasks.build import (
 from iib.workers.config import get_worker_config
 from iib.workers.tasks.celery import app
 from iib.workers.tasks.utils import (
+    get_image_labels,
     get_resolved_image,
     podman_pull,
     request_logger,
@@ -217,6 +218,7 @@ def _adjust_operator_bundle(manifests_path, metadata_path, organization=None, pi
         organization_customizations = [
             {'type': 'package_name_suffix'},
             {'type': 'registry_replacements'},
+            {'type': 'image_name_from_labels'},
             {'type': 'csv_annotations'},
         ]
 
@@ -245,11 +247,18 @@ def _adjust_operator_bundle(manifests_path, metadata_path, organization=None, pi
             if org_csv_annotations:
                 log.info('Applying csv annotations for organization %s', organization)
                 _adjust_csv_annotations(operator_manifest.files, package_name, org_csv_annotations)
+        elif customization_type == 'image_name_from_labels':
+            org_image_name_template = customization.get('template', '')
+            if org_image_name_template:
+                bundle_metadata = _get_bundle_metadata(
+                    operator_manifest, pinned_by_iib, perform_sanity_checks=False
+                )
+                _replace_image_name_from_labels(bundle_metadata, org_image_name_template)
 
     return labels
 
 
-def _get_bundle_metadata(operator_manifest, pinned_by_iib):
+def _get_bundle_metadata(operator_manifest, pinned_by_iib, perform_sanity_checks=True):
     """
     Get bundle metadata i.e. CSV's and all relatedImages pull specifications.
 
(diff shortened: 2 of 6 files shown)


midnightercz (changes requested):

iib/workers/tasks/build_regenerate_bundle.py
```diff
@@ -363,14 +376,65 @@ def _resolve_image_pull_specs(bundle_metadata, labels, pinned_by_iib, registry_r
...
+    replacement_labels = re.findall(label_regex, replacement_template)
+
+    for pullspec in bundle_metadata['found_pullspecs']:
+        new_pullspec = ImageName.parse(pullspec.to_str())
+        pullspec_labels = get_image_labels(pullspec.to_str())
+        modified_namespace_repo = replacement_template
+
+        for label in replacement_labels:
+            label_value = pullspec_labels.get(label)
+            if label_value:
```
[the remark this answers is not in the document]
yashvardhannanavati: No specific reason. I was already using regex to find the label names in the string so used it to substitute too. If you feel strongly against it, I can change it.

midnightercz (approved):

iib/workers/tasks/build_regenerate_bundle.py
```diff
@@ -234,52 +237,62 @@ def _adjust_operator_bundle(manifests_path, metadata_path, organization=None, pi
...
-def _get_bundle_metadata(operator_manifest, pinned_by_iib):
+def _get_bundle_metadata(operator_manifest, pinned_by_iib, pinning_metadata=False):
     """
     Get bundle metadata i.e. CSV's and all relatedImages pull specifications.
 
     :param operator_manifest.operator.OperatorManifest operator_manifest: the operator manifest
         object.
     :param bool pinned_by_iib: whether or not the bundle image has already been processed by
         IIB to perform image pinning of related images.
+    :param bool pinning_metadata: boolean specifying if metadata should be filtered for pinning
```
lcarva, line 266: After reading the code, it makes sense what the purpose of this parameter is. Can we consider renaming it something like `ignore_sanity_checks` ? I think that better describes its intent.

```
:param bool ignore_sanity_check: boolean specifying if sanity checks should be skipped.
```

I don't think that's great either, but maybe slightly better?

iib/workers/tasks/build_regenerate_bundle.py
```diff
@@ -363,14 +376,65 @@ def _resolve_image_pull_specs(bundle_metadata, labels, pinned_by_iib, registry_r
...
+        if invalid_labels:
+            raise IIBError(
+                f'Pull spec {pullspec.to_str()} is missing the following label(s)'
+                f' required in the image_name_from_labels {replacement_template}:'
+                f' {", ".join(invalid_labels)}'
+            )
+        new_pullspec.namespace, new_pullspec.repo = modified_namespace_repo.split('/', 1)
+        replacement_pullspecs[pullspec] = new_pullspec
+
+    _replace_csv_pullspecs(bundle_metadata, replacement_pullspecs)
```
lcarva, line 416: Why not `set_related_images` ?

iib/workers/tasks/build_regenerate_bundle.py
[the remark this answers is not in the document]
lcarva: Why do we need regex in the first place? We should be able to create a dict that maps labels to their values, then simply call `my_string.format(**labels)`

iib/workers/tasks/build_regenerate_bundle.py
```diff
@@ -213,6 +215,7 @@ def _adjust_operator_bundle(manifests_path, metadata_path, organization=None, pi
         organization_customizations = [
```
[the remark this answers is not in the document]
yashvardhannanavati: None of them will be applied except for `resolve_pull_specs` which is under `registry_replacements` (pre-existing code) and has to be done irrespective of the customization org. This is to avoid a huge else block where we call a bunch of functions and avoid code duplication.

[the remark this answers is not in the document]
yashvardhannanavati: More specifically, this => https://github.com/release-engineering/iib/commit/30525746d11bfe833ce78b3516e2960ed17514da#diff-1f7f19fb78fb596298715dcd50d317bee99dc1e246c61918856ee9c716f0eff8R242

iib/workers/tasks/build_regenerate_bundle.py
yashvardhannanavati: Because we already `set_related_images` when resolving the pull_specs. If we do it again here, operator-manifest adds an another entry instead of replacing the existing one. Do you think this should be changed in the operator-manifest library itself?

iib/workers/tasks/build_regenerate_bundle.py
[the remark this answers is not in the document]
yashvardhannanavati: Ah, yes. I did think about using `.format`, but it would throw a keyError in-case there are invalid labels present and that too only for the first invalid key. In which case, we will still have to catch the exception and still use regex to find the invalid labels. So I resorted to using regex throughout. Sorry if I missed something? 

```
>>> x = '{name}-version-{version}-{something}'
>>> y = {'name': 'some_name', 'other_key':'something'}
>>> x.format(**y)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
KeyError: 'version'
```

README.md
```diff
@@ -260,14 +260,19 @@ The custom configuration options for the Celery workers are listed below:
...
+    `{package_name}` in the template string.
+  * The `package_name_suffix` customization type is a dictionary where the key `suffix` value is
+    a string of a suffix to add to the package name of the operator.
+  * The `registry_replacements` customization type is a dictionary where the key `replacements`
+    value is a dictionary where the keys are the old registries to replace and the values
+    are the registries to replace the old registries with. This replaces the registry in all
+    the ClusterServiceVersion files.
+  * The `image_name_from_labels` customization type is a dictionary where the key `template`
+    value is a string which specifies a combination of label names in curly braces which will be
+    substituted with the actual label values from the bundle image.
```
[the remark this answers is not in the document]
yashvardhannanavati: Thinking outside of our use-case, couldn't it ideally be any label that's present on the bundle image? Hence, I do not think we should list any specific variables here. thoughts?

iib/workers/tasks/build_regenerate_bundle.py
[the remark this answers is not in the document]
lcarva: You can catch `KeyError` and provide a more helpful message which could include which labels are available and the template that IIB is trying to use. That should help debugging.

I just don't think it's worth adding the regex complexity for this.

README.md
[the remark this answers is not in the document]
lcarva: You're right.

iib/workers/tasks/build_regenerate_bundle.py
lcarva: Ok that makes sense. Maybe add a comment?

iib/workers/tasks/build_regenerate_bundle.py
```diff
@@ -238,52 +240,62 @@ def _adjust_operator_bundle(manifests_path, metadata_path, organization=None, pi
...
-            log.info('Skipping pinning because related images have already been pinned by IIB')
-        elif operator_csv.has_related_images():
-            csv_file_name = os.path.basename(operator_csv.path)
-            if operator_csv.has_related_image_envs():
-                raise IIBError(
-                    f'The ClusterServiceVersion file {csv_file_name} has entries in '
-                    'spec.relatedImages and one or more containers have RELATED_IMAGE_* '
-                    'environment variables set. This is not allowed for bundles regenerated with '
-                    'IIB.'
+        if ignore_sanity_checks:
```
lcarva, line 277: This seems to be the inverse of ignore. We should either rename this flag or invert the check.

lcarva: If you invert the check, also invert the calls to `_get_bundle_metadata`, of course.

iib/workers/tasks/build_regenerate_bundle.py
```diff
@@ -367,14 +379,55 @@ def _resolve_image_pull_specs(bundle_metadata, labels, pinned_by_iib, registry_r
...
+        pullspec_labels = get_image_labels(pullspec.to_str())
+        try:
+            modified_namespace_repo = replacement_template.format(**pullspec_labels)
+        except KeyError:
+            raise IIBError(
+                f'Pull spec {pullspec.to_str()} is missing one or more label(s)'
+                f' required in the image_name_from_labels {replacement_template}.'
+                f' Available labels: {", ".join(list(pullspec_labels.keys()))}'
+            )
+        new_pullspec.namespace, new_pullspec.repo = modified_namespace_repo.split('/', 1)
```
lcarva, line 406: What if the `modified_namespace_repo` does not contain `/` ?

lcarva, line 406: We also need to take into account flattening of repo names. For example, `spam/eggs/bacon` is invalid in some registries, i.e. quay.io. It should be turned into `spam/eggs-bacon`.

yashvardhannanavati: In order to flatten the repo name, you will have to use the enclose_repo customization in #245. So I think we can ignore flattening here. Since the customization only promises to replace the image name with labels.

yashvardhannanavati: Good point

iib/workers/tasks/build_regenerate_bundle.py
yashvardhannanavati: doh, my bad. Yes, I'll change it

iib/workers/tasks/build_regenerate_bundle.py
yashvardhannanavati: I can see how the user will be stranded if the name label has `spam/eggs/bacon` and then they specify `spam` as `namespace` and `enclose_glue` as `-`. It would become `spam/spam-eggs-bacon` which is invalid. In that case, we can alter the other PR to have namespace as optional in which case, the current namespace will be used and or the name, `/`s will be replaced with the enclose glue. thoughts?

lcarva: This should be handled in #245, I agree.

lcarva: Note from meeting:

* If there are two items after split, then the first is namespace, the second is repo. (as is in code right now)
* If there is only one item after split, then namespaces is `None`, and the only item is repo.

yashvardhannanavati: Done

iib/workers/tasks/build_regenerate_bundle.py
```diff
@@ -367,14 +379,63 @@ def _resolve_image_pull_specs(bundle_metadata, labels, pinned_by_iib, registry_r
...
+        try:
+            modified_namespace_repo = replacement_template.format(**pullspec_labels)
+        except KeyError:
+            raise IIBError(
+                f'Pull spec {pullspec.to_str()} is missing one or more label(s)'
+                f' required in the image_name_from_labels {replacement_template}.'
+                f' Available labels: {", ".join(list(pullspec_labels.keys()))}'
+            )
+
+        namespace_repo_list = modified_namespace_repo.split('/', 1)
```
lcarva, line 407: Ha! Cleaver :)

lcarva (approved):

PR #244, opened 2021-04-30
[Do not merge] Upgrade IIB cryptography package

lkolacek:
Closes CLOUDBLD-4575

Signed-off-by: Ladislav Kolacek <lkolacek@redhat.com>

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,7 @@
 ---
 language: python
 python:
-  - "3.6"
-  - "3.7"
+  - "3.8"
 install:
   - pip install tox
 script:
--- a/requirements-test.txt
+++ b/requirements-test.txt
@@ -121,21 +121,15 @@ coverage==5.5 \
     # via
     #   -r requirements-test.in
     #   pytest-cov
-cryptography==3.3.2 \
-    --hash=sha256:0d7b69674b738068fa6ffade5c962ecd14969690585aaca0a1b1fc9058938a72 \
-    --hash=sha256:1bd0ccb0a1ed775cd7e2144fe46df9dc03eefd722bbcf587b3e0616ea4a81eff \
-    --hash=sha256:3c284fc1e504e88e51c428db9c9274f2da9f73fdf5d7e13a36b8ecb039af6e6c \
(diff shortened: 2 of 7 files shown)


lkolacek:
Hi @MartinBasti  @yashvardhannanavati,

I added all deps for `cryptography` but I still can't make it work for `Py3.6`.
I also tried to update pip package as it was reccommended in cryptography documentation but nothing seems to work.

Do we really want to run and set the following? And if yes where should I put it?
```
curl https://sh.rustup.rs -sSf | bash -s -- -y
PATH="${HOME}/.cargo/bin:${PATH}"
```

lcarva:
Thanks for doing this! I think we should hold off merging this until IIB itself is running on python 3.8. I wonder if, to make the transition easier, we need a separate PR that simply adds python 3.8 (and maybe remove 3.7?) without upgrading any packages. We will want to ensure the tests work just fine on python 3.8 before upgrading the IIB environment to that version. Wdyt?

lkolacek:
@lcarva  separate PR for adding Py3.8 support and testing the current version of packages sounds good! I updated the description here: https://projects.engineering.redhat.com/browse/CLOUDBLD-5803

I also updated Summary of this PR to non to be accidentally merged.

pbortlov:
@lcarva @yashvardhannanavati @midnightercz @lipoja @querti Could we circle back here? It seems that we can't use the newest versions of libraries which we have to update because of security vulnerabilities. #269

MartinBasti:
> @lcarva @yashvardhannanavati @midnightercz @lipoja @querti Could we circle back here? It seems that we can't use the newest versions of libraries which we have to update because of security vulnerabilities. #269

I don't follow why not, can you elaborate more?

lkolacek:
+1 can you provide more info @pbortlov ? PRs do not seem to be affecting each other at first sight.

pbortlov:
@lkolacek we have a security vulnerability with urllib3 and if I bump the library to the secure version python3.6 is failing. So I was thinking to circle back here if we can open the discussion about python3.8

pbortlov:
@lkolacek @lcarva @yashvardhannanavati @release-engineering/exd-guild-hello-operator 
This PR can be closed because there is no need to upgrade to python 3.8 if you upgrade pip in the IIB. please see #279 #282

pbortlov:
This will be closed in favor of #282

PR #249, opened 2021-05-13
Remove the support of privileged users

emilyzheng:
For all users, overwriting index image is only allowed when overwrite_from_index/overwrite_target_index is True and overwrite_from_index_token/overwrite_target_index_token is set.

### DIFF
--- a/iib/workers/tasks/build.py
+++ b/iib/workers/tasks/build.py
@@ -456,8 +456,8 @@ def _opm_index_add(
     :param str from_index: the pull specification of the container image containing the index that
         the index image build will be based from.
     :param str overwrite_from_index_token: the token used for overwriting the input
-        ``from_index`` image. This is required for non-privileged users to use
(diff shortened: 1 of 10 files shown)


emilyzheng:
@yashvardhannanavati Could you help to add reviewers?

lipoja (approved):

pbortlov:
I am wondering if this branch is rebased with the newest upstream/master it seems that you are deleting functions from build.py which is not in the scope of this PR.

iib/workers/tasks/build.py
```diff
@@ -255,32 +255,6 @@ def _get_local_pull_spec(request_id, arch, include_transport=False):
     return pull_spec
 
 
-def _get_image_arches(pull_spec):
```
pbortlov, line 258: This is not in the scope of this PR

pbortlov (changes requested):

pbortlov (approved):

yashvardhannanavati (approved):

PR #258, opened 2021-06-09
New attribute 'build_tags' for index_image operations

midnightercz:
For index image operations, it's now possible to set build_tags
attribute to apply extra tags for intermediate index image

### DIFF
--- a/iib/web/models.py
+++ b/iib/web/models.py
@@ -241,6 +241,25 @@ class Operator(db.Model):
         return operator
 
 
+class BuildTag(db.Model):
+    """Extra tag associated with built index image."""
+
+    id = db.Column(db.Integer, primary_key=True)
+    name = db.Column(db.String, nullable=False, unique=False)
+
+
+class RequestBuildTag(db.Model):
+    """Association table for extra build tags and build request."""
+
+    request_id = db.Column(
+        db.Integer, db.ForeignKey('request.id'), index=True, nullable=False, primary_key=True
+    )
+    tag_id = db.Column(
+        db.Integer, db.ForeignKey('build_tag.id'), autoincrement=False, index=True, primary_key=True
+    )
+    __table_args__ = (db.UniqueConstraint('request_id', 'tag_id'),)
+
+
 class RequestRmOperator(db.Model):
     """An association table between rm requests and the operators they contain."""
 
@@ -305,6 +324,9 @@ class Request(db.Model):
         order_by='RequestState.updated',
     )
     user = db.relationship('User', back_populates='requests')
+    build_tags = db.relationship(
+        'BuildTag', order_by='BuildTag.name', secondary=RequestBuildTag.__table__
+    )
 
     __mapper_args__ = {
         'polymorphic_identity': RequestTypeMapping.__members__['generic'].value,
@@ -359,6 +381,21 @@ class Request(db.Model):
         db.session.add(request_state)
         db.session.flush()
         self.request_state_id = request_state.id
+
+    def add_build_tag(self, name):
+        """
+        Add a RequestBuildTag associated with the current request.
+
+        :param str name: tag name
+        """
+        bt = db.session.query(BuildTag).filter_by(name=name).first()
+        if not bt:
+            bt = BuildTag(name=name)
+            db.session.add(bt)
+            db.session.flush()
+
+        if bt not in self.build_tags:
+            self.build_tags.append(bt)
 
     def add_architecture(self, arch_name):
         """
@@ -625,6 +662,7 @@ def get_request_query_options(verbose=False):
         joinedload(RequestAdd.from_index_resolved),
         joinedload(RequestAdd.index_image),
         joinedload(RequestAdd.index_image_resolved),
+        joinedload(RequestAdd.build_tags),
         joinedload(RequestRegenerateBundle.bundle_image),
         joinedload(RequestRegenerateBundle.from_bundle_image),
         joinedload(RequestRegenerateBundle.from_bundle_image_resolved),
@@ -635,6 +673,8 @@ def get_request_query_options(verbose=False):
         joinedload(RequestRm.index_image),
         joinedload(RequestRm.index_image_resolved),
         joinedload(RequestRm.operators),
+        joinedload(RequestRm.build_tags),
+        joinedload(RequestMergeIndexImage.build_tags),
     ]
     if verbose:
         query_options.append(joinedload(Request.states))
@@ -739,6 +779,7 @@ class RequestIndexImageMixin:
             'overwrite_from_index',
             'overwrite_from_index_token',
             'distribution_scope',
+            'build_tags',
         } | set(additional_optional_params or [])
 
         validate_request_params(
@@ -842,6 +883,7 @@ class RequestIndexImageMixin:
             'organization': None,
             'removed_operators': [],
             'distribution_scope': self.distribution_scope,
+            'build_tags': [tag.name for tag in self.build_tags],
         }
 
     def get_index_image_mutable_keys(self):
@@ -922,6 +964,7 @@ class RequestAdd(Request, RequestIndexImageMixin):
                 'bundles',
                 'distribution_scope',
                 'deprecation_list',
+                'build_tags',
             ],
             batch=batch,
         )
@@ -930,8 +973,12 @@ class RequestAdd(Request, RequestIndexImageMixin):
             request_kwargs[key] = [
                 Image.get_or_create(pull_specification=item) for item in request_kwargs.get(key, [])
             ]
-
+        build_tags = request_kwargs.pop('build_tags', [])
         request = cls(**request_kwargs)
+
+        for bt in build_tags:
+            request.add_build_tag(bt)
+
         request.add_state('in_progress', 'The request was initiated')
         return request
 
@@ -1006,13 +1053,18 @@ class RequestRm(Request, RequestIndexImageMixin):
             raise ValidationError(f'"operators" should be a non-empty array of strings')
 
         cls._from_json(
-            request_kwargs, additional_required_params=['operators', 'from_index'], batch=batch
+            request_kwargs, additional_required_params=['operators', 'from_index'], batch=batch,
         )
 
         request_kwargs['operators'] = [Operator.get_or_create(name=item) for item in operators]
 
+        build_tags = request_kwargs.pop('build_tags', [])
         request = cls(**request_kwargs)
         request.add_state('in_progress', 'The request was initiated')
+
+        for bt in build_tags:
+            request.add_build_tag(bt)
+
         return request
 
     def to_json(self, verbose=True):
@@ -1064,6 +1116,7 @@ class RequestRegenerateBundle(Request):
     __mapper_args__ = {
         'polymorphic_identity': RequestTypeMapping.__members__['regenerate_bundle'].value
     }
+    build_tags = None
 
     @classmethod
     def from_json(cls, kwargs, batch=None):
@@ -1264,6 +1317,13 @@ class RequestMergeIndexImage(Request):
                 )
             request_kwargs['distribution_scope'] = distribution_scope
 
+        if not isinstance(request_kwargs.get('build_tags', []), list) or any(
+            not item or not isinstance(item, str) for item in request_kwargs.get('build_tags', [])
+        ):
+            raise ValidationError(
+                f'"build_tags" should be either an empty array or an array of non-empty strings'
+            )
+
         # current_user.is_authenticated is only ever False when auth is disabled
         if current_user.is_authenticated:
             request_kwargs['user'] = current_user
@@ -1274,6 +1334,12 @@ class RequestMergeIndexImage(Request):
         request_kwargs['batch'] = batch
 
         request = cls(**request_kwargs)
+
+        build_tags = request_kwargs.pop('build_tags', [])
+
+        for bt in build_tags:
+            request.add_build_tag(bt)
+
         request.add_state('in_progress', 'The request was initiated')
         return request
 
@@ -1301,6 +1367,7 @@ class RequestMergeIndexImage(Request):
             self.target_index_resolved, 'pull_specification', None
         )
         rv['distribution_scope'] = self.distribution_scope
+        rv['build_tags'] = [t.name for t in self.build_tags]
 
         return rv
 
@@ -1452,6 +1519,7 @@ class RequestCreateEmptyIndex(Request, RequestIndexImageMixin):
     __mapper_args__ = {
         'polymorphic_identity': RequestTypeMapping.__members__['create_empty_index'].value
     }
+    build_tags = None
 
     @property
     def labels(self):
@@ -1495,7 +1563,12 @@ class RequestCreateEmptyIndex(Request, RequestIndexImageMixin):
                 if not isinstance(value, str) or not isinstance(key, str):
                     raise ValidationError(f'The key and value of "{key}" must be a string')
 
-        for arg in ('add_arches', 'overwrite_from_index', 'overwrite_from_index_token'):
+        for arg in (
+            'add_arches',
+            'overwrite_from_index',
+            'overwrite_from_index_token',
+            'build_tags',
+        ):
             if arg in request_kwargs:
                 raise ValidationError(
                     f'The "{arg}" arg is invalid for the create-empty-index endpoint.'
@@ -1528,6 +1601,7 @@ class RequestCreateEmptyIndex(Request, RequestIndexImageMixin):
         rv.pop('organization')
         rv.pop('deprecation_list')
         rv.pop('removed_operators')
+        rv.pop('build_tags')
         rv['labels'] = self.labels
         return rv
 
--- a/iib/workers/tasks/build.py
+++ b/iib/workers/tasks/build.py
@@ -106,54 +106,69 @@ def _cleanup():
 
 
 @retry(exceptions=IIBError, tries=get_worker_config().iib_total_attempts, logger=log)
-def _create_and_push_manifest_list(request_id, arches):
+def _create_and_push_manifest_list(request_id, arches, build_tags):
     """
     Create and push the manifest list to the configured registry.
 
     :param int request_id: the ID of the IIB build request
     :param iter arches: an iterable of arches to create the manifest list for
+    :param build_tags: list of extra tag to use for intermediate index image
     :return: the pull specification of the manifest list
     :rtype: str
     :raises IIBError: if creating or pushing the manifest list fails
     """
     buildah_manifest_cmd = ['buildah', 'manifest']
-    output_pull_spec = get_rebuilt_image_pull_spec(request_id)
+    output_pull_specs = [get_rebuilt_image_pull_spec(request_id)]
+    _tags = [request_id]
+    if build_tags:
+        _tags += build_tags
+    conf = get_worker_config()
+    for tag in _tags:
+        for output_pull_spec in output_pull_specs:
+            output_pull_spec = conf['iib_image_push_template'].format(
+                registry=conf['iib_registry'], request_id=tag
+            )
+            try:
+                run_cmd(
+                    buildah_manifest_cmd + ['rm', output_pull_spec],
+                    exc_msg=f'Failed to remove local manifest list. {output_pull_spec} does not exist',
+                )
+            except IIBError as e:
+                error_msg = str(e)
+                if 'Manifest list not found locally.' not in error_msg:
+                    raise IIBError(f'Error removing local manifest list: {error_msg}')
+                log.debug('Manifest list cannot be removed. No manifest list %s found', output_pull_spec)
+            log.info('Creating the manifest list %s locally', output_pull_spec)
+            run_cmd(
+                buildah_manifest_cmd + ['create', output_pull_spec],
+                exc_msg=f'Failed to create the manifest list locally: {output_pull_spec}',
+            )
+            for arch in sorted(arches):
+                arch_pull_spec = _get_external_arch_pull_spec(
+                    request_id, arch, include_transport=True
+                )
+                log.debug(
+                    'Adding the manifest %s to the manifest list %s',
+                    arch_pull_spec,
+                    output_pull_spec,
+                )
+                run_cmd(
+                    buildah_manifest_cmd + ['add', output_pull_spec, arch_pull_spec],
+                    exc_msg=(
+                        f'Failed to add {arch_pull_spec} to the'
+                        f' local manifest list: {output_pull_spec}'
+                    ),
+                )
 
-    try:
-        run_cmd(
-            buildah_manifest_cmd + ['rm', output_pull_spec],
-            exc_msg=f'Failed to remove local manifest list. {output_pull_spec} does not exist',
-        )
-    except IIBError as e:
-        error_msg = str(e)
-        if 'Manifest list not found locally.' not in error_msg:
-            raise IIBError(f'Error removing local manifest list: {error_msg}')
-        log.debug('Manifest list cannot be removed. No manifest list %s found', output_pull_spec)
+            log.debug('Pushing manifest list %s', output_pull_spec)
+            run_cmd(
+                buildah_manifest_cmd
+                + ['push', '--all', output_pull_spec, f'docker://{output_pull_spec}'],
+                exc_msg=f'Failed to push the manifest list to {output_pull_spec}',
+            )
 
-    log.info('Creating the manifest list %s locally', output_pull_spec)
-    run_cmd(
-        buildah_manifest_cmd + ['create', output_pull_spec],
-        exc_msg=f'Failed to create the manifest list locally: {output_pull_spec}',
-    )
-    for arch in sorted(arches):
-        arch_pull_spec = _get_external_arch_pull_spec(request_id, arch, include_transport=True)
-        log.debug(
-            'Adding the manifest %s to the manifest list %s', arch_pull_spec, output_pull_spec
-        )
-        run_cmd(
-            buildah_manifest_cmd + ['add', output_pull_spec, arch_pull_spec],
-            exc_msg=(
-                f'Failed to add {arch_pull_spec} to the' f' local manifest list: {output_pull_spec}'
-            ),
-        )
-
-    log.debug('Pushing manifest list %s', output_pull_spec)
-    run_cmd(
-        buildah_manifest_cmd + ['push', '--all', output_pull_spec, f'docker://{output_pull_spec}'],
-        exc_msg=f'Failed to push the manifest list to {output_pull_spec}',
-    )
-
-    return output_pull_spec
+    # return 1st item as it holds production tag
+    return output_pull_specs[0]
 
 
 def _update_index_image_pull_spec(
@@ -800,6 +815,7 @@ def handle_add_request(
     greenwave_config=None,
     binary_image_config=None,
     deprecation_list=None,
+    build_tags=None,
 ):
     """
     Coordinate the the work needed to build the index image with the input bundles.
@@ -833,6 +849,7 @@ def handle_add_request(
         ``binary_image`` to use.
     :param list deprecation_list: list of deprecated bundles for the target index image. Defaults
         to ``None``.
+    :param list build_tags: List of tags which will be applied to intermediate index images.
     :raises IIBError: if the index image build fails.
     """
     _cleanup()
@@ -989,7 +1006,7 @@ def handle_add_request(
         )
 
     set_request_state(request_id, 'in_progress', 'Creating the manifest list')
-    output_pull_spec = _create_and_push_manifest_list(request_id, arches)
+    output_pull_spec = _create_and_push_manifest_list(request_id, arches, build_tags)
 
     _update_index_image_pull_spec(
         output_pull_spec,
@@ -1018,6 +1035,7 @@ def handle_rm_request(
     overwrite_from_index_token=None,
     distribution_scope=None,
     binary_image_config=None,
+    build_tags=None,
 ):
     """
     Coordinate the work needed to remove the input operators and rebuild the index image.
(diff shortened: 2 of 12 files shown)


pbortlov:
Hi @midnightercz, what index-image operations do you mean? Add bundles, Rm Operator, Merge index image?

midnightercz:
@pbortlov I meant add and remove. But it's true that it should be probably also added to merge index image as well

pbortlov:
Could we wait for #229 to be merged to add this variable to the new endpoint?

midnightercz:
@pbortlov yeah, we can wait

iib/web/models.py
```diff
@@ -997,6 +1030,14 @@ def from_json(cls, kwargs, batch=None):
         """
         request_kwargs = deepcopy(kwargs)
 
+        for key in ('build_tags',):
```
yashvardhannanavati, line 1033: I don't think you need this validation here right? Since you are calling the parent class `_from_json`?

iib/web/models.py
```diff
@@ -1013,6 +1054,17 @@ def from_json(cls, kwargs, batch=None):
 
         request = cls(**request_kwargs)
         request.add_state('in_progress', 'The request was initiated')
+        db.session.add(request)
```
yashvardhannanavati, line 1057: I would prefer having this logic in the `api_v1.yaml` for all the requests.

yashvardhannanavati: Looks like similar code to me for all requests. Maybe you can create a helper method in ` api_v1.yaml` ?

iib/web/models.py
```diff
@@ -1269,6 +1322,16 @@ def from_json(cls, kwargs, batch=None):
         request_kwargs['batch'] = batch
 
         request = cls(**request_kwargs)
+        db.session.add(request)
```
yashvardhannanavati, line 1325: Same as above

iib/workers/tasks/build.py
```diff
@@ -991,14 +1004,14 @@ def handle_add_request(
         )
 
     set_request_state(request_id, 'in_progress', 'Creating the manifest list')
-    output_pull_spec = _create_and_push_manifest_list(request_id, arches)
+    output_pull_specs = _create_and_push_manifest_list(request_id, arches, build_tags)
     if legacy_support_packages:
         export_legacy_packages(
-            legacy_support_packages, request_id, output_pull_spec, cnr_token, organization
+            legacy_support_packages, request_id, output_pull_specs[0], cnr_token, organization
```
yashvardhannanavati, line 1010: could you please add a comment here as to why we are using the first element of the list?

iib/workers/tasks/build_regenerate_bundle.py
```diff
@@ -118,7 +118,7 @@ def handle_regenerate_bundle_request(
                 _push_image(request_id, arch)
 
     set_request_state(request_id, 'in_progress', 'Creating the manifest list')
-    output_pull_spec = _create_and_push_manifest_list(request_id, arches)
+    output_pull_spec = _create_and_push_manifest_list(request_id, arches, [])
```
yashvardhannanavati, line 121: You are accepting `build_tags` from the user in the API and not using them at all. I think we shouldn't give an option at all, no?
Am I missing something?

yashvardhannanavati (changes requested):

iib/workers/tasks/build.py
```diff
@@ -106,59 +106,70 @@ def _cleanup():
...
+    output_pull_specs = [get_rebuilt_image_pull_spec(request_id)]
+    if build_tags:
+        conf = get_worker_config()
+        for tag in build_tags:
+            output_pull_specs.append(
+                conf['iib_image_push_template'].format(
+                    registry=conf['iib_registry'], request_id=tag
                 )
             )
-            for arch in sorted(arches):
```
pbortlov, line 132: Why is this removed?

iib/workers/tasks/build.py
```diff
@@ -106,59 +106,70 @@ def _cleanup():
...
-                arch_pull_spec = _get_external_arch_pull_spec(request_id, arch)
-                log.debug(
-                    'Adding the manifest %s to the manifest list %s',
-                    arch_pull_spec,
-                    output_pull_spec,
-                )
+
+    for output_pull_spec in output_pull_specs:
```
pbortlov, line 130: if I understand correctly, every pull spec will be appended with each arch?
Could you give me an example how the yaml file will look like in the end?

iib/workers/tasks/build_merge_index_image.py
```diff
@@ -135,7 +135,7 @@ def _add_bundles_missing_in_source(
     )
     _build_image(base_dir, 'index.Dockerfile', request_id, arch)
     _push_image(request_id, arch)
-    _create_and_push_manifest_list(request_id, [arch])
+    _create_and_push_manifest_list(request_id, [arch], [])
```
pbortlov, line 138: why is here an empty list?

tests/test_workers/test_tasks/test_build_regenerate_bundle.py
```diff
@@ -115,7 +115,7 @@ def test_handle_regenerate_bundle_request(
         ]
     )
 
-    mock_capml.assert_called_once_with(request_id, list(arches))
+    mock_capml.assert_called_once_with(request_id, list(arches), [])
```
pbortlov, line 118: I am missing tests for create_empty_index, merge_index_image

pbortlov (changes requested):

iib/web/models.py
midnightercz: but there's no validation for build_tags there

iib/web/models.py
midnightercz: I'm not sure I understand your request. Logic in yaml?

iib/workers/tasks/build.py
midnightercz: this was just moved bellow

iib/workers/tasks/build.py
midnightercz: that's correct. Yaml files will look very same as they were before this change

iib/workers/tasks/build_merge_index_image.py
midnightercz: because it's not needed here. Final call of _create_and_push_manifest_list is in handle_merge_request

iib/workers/tasks/build_regenerate_bundle.py
midnightercz: it's shared method, so if you don't want to use them, you need to pass there empty list. Build_tags are not accepted for regenerate_bundles. Or do you think we should support it as well?

tests/test_workers/test_tasks/test_build_regenerate_bundle.py
midnightercz: because there weren't any which works with real _create_and_push_manifest_list

iib/workers/tasks/build.py
pbortlov: thanks. I saw the example in tests

tests/test_workers/test_tasks/test_build_regenerate_bundle.py
pbortlov: oh, that's why. Thank you for explaining :)

iib/web/models.py
```diff
@@ -635,6 +655,7 @@ def get_request_query_options(verbose=False):
         joinedload(RequestRm.index_image),
         joinedload(RequestRm.index_image_resolved),
         joinedload(RequestRm.operators),
+        joinedload(RequestRm.build_tags),
```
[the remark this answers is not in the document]
pbortlov: I think so, because `get_request_query_options` is called by `get_builds`. Could you create tasks for both of them?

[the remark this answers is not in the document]
pbortlov: THanks for pointing this out @lipoja

iib/web/models.py
pbortlov: Do you need braces around the build_tags?

pbortlov: I will add it to the Request._from_json if you want to use it for RequestAdd, RequestRm, RequestCreateEmptyImage,

iib/web/models.py
```diff
@@ -1013,6 +1055,17 @@ def from_json(cls, kwargs, batch=None):
 
         request = cls(**request_kwargs)
         request.add_state('in_progress', 'The request was initiated')
+        db.session.add(request)
+        db.session.flush()
+
+        build_tags = []
+        if 'build_tags' in request_kwargs:
+            build_tags = request_kwargs.pop('build_tags')
```
pbortlov, line 1063: What about request_kwargs.pop('build_tags', []) instead of those 3 lines?
This logic can be done in Request._from_json() to avoid uplicates

iib/web/models.py
```diff
@@ -930,8 +955,17 @@ def from_json(cls, kwargs, batch=None):
             request_kwargs[key] = [
                 Image.get_or_create(pull_specification=item) for item in request_kwargs.get(key, [])
             ]
-
+        build_tags = []
```
pbortlov, line 958: What about request_kwargs.pop('build_tags', []) instead of those 3 lines?
This logic can be done in Request._from_json() to avoid uplicates

iib/web/models.py
```diff
@@ -886,7 +911,7 @@ def from_json(cls, kwargs, batch=None):
         """
         request_kwargs = deepcopy(kwargs)
 
-        for key in ('bundles', 'deprecation_list'):
+        for key in ('bundles', 'deprecation_list', 'build_tags'):
```
pbortlov, line 914: Let's check this in Request._from_json

iib/web/models.py
```diff
@@ -1269,6 +1323,16 @@ def from_json(cls, kwargs, batch=None):
         request_kwargs['batch'] = batch
 
         request = cls(**request_kwargs)
+        db.session.add(request)
+        db.session.flush()
+
+        build_tags = []
```
pbortlov, line 1329: Please add it to the Request

iib/web/models.py
```diff
@@ -1296,6 +1360,7 @@ def to_json(self, verbose=True):
             self.target_index_resolved, 'pull_specification', None
         )
         rv['distribution_scope'] = self.distribution_scope
+        rv['build_tags'] = [t.name for t in self.build_tags]
```
pbortlov, line 1363: Interestingly, you didn't add this line to the classes above RequestAdd and RequestRm. What about to add it to the Request class and this line will ne inherited for RequestAdd, RequestRm, RequestRegenerateBundle and RequestCreateEmptyImage.

iib/workers/tasks/build.py
```diff
@@ -1028,6 +1043,7 @@ def handle_rm_request(
     overwrite_from_index_token=None,
     distribution_scope=None,
     binary_image_config=None,
+    build_tags=[],
```
pbortlov, line 1046: Should it here None instead?

pbortlov (changes requested):

iib/web/models.py
midnightercz: yeah, this can be reduced into one line

iib/web/models.py
midnightercz: I'm not sure build_tags should be available for all requests. Therefore I added it per class

iib/web/models.py
midnightercz: I added it here only because of the inheritance you mentioned.

iib/workers/tasks/build.py
midnightercz: it doesn't matter if it's [] or None. But yeah, I should have been consistent

iib/web/models.py
yashvardhannanavati: Sorry I meant api_v1.py

iib/web/models.py
yashvardhannanavati: Since you have added it under additional_optional_params in `_from_json`, let's have this check in there if the user has provided `build_tags`. That would mean removing it from here.

iib/web/models.py
yashvardhannanavati: Same as above, moving this check from here to _from_json will help you get rid of this code duplication.

yashvardhannanavati: Simply pass `build_tags` in the call on line 1050 while calling `cls._from_json` under `additional_optional_params ` and that would do the trick.

yashvardhannanavati: Same can be done for all three classes `RequestAdd`, `RequestRm` and `RequestMergeIndex`

iib/web/models.py
yashvardhannanavati: So for all the three types of requests, let's have this piece of code in `api_v1.py` file instead.

iib/web/models.py
```diff
@@ -305,6 +314,7 @@ class Request(db.Model):
         order_by='RequestState.updated',
     )
     user = db.relationship('User', back_populates='requests')
+    build_tags = db.relationship('BuildTag', back_populates='request')
```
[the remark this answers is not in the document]
yashvardhannanavati: Ah, I see it the merge index isn't inheriting mixin class. In that case, let's add this in Mixin for `Add` and `Rm` and in the `RequestMergeIndex` class too. That will solve your problem

iib/web/models.py
yashvardhannanavati: Let's move this to `api_v1.py` and you will have to duplicate the validation here since you are not calling `cls._from_json` here since the Mixin class is not being inherited here.

iib/workers/tasks/build_regenerate_bundle.py
yashvardhannanavati: No, I agree they shouldn't be supported in regenerate-bundle.

yashvardhannanavati (changes requested):
@midnightercz This is looking good except for the changes I suggested in `models.py`. Please feel free to setup a call if my comments aren't clear.

iib/web/models.py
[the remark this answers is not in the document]
midnightercz: That's not the only reason. Another reason is back_populates attribute which points to Request model. I'm not sure if it's still going to work if I specify that in RequestRm but attribute itself won't be in Request. Also I don't think you can set multiple values in back_populates.
Therefore I simply set build_tags to None for classes where it's not relevant

iib/workers/tasks/build.py
```diff
@@ -110,59 +110,70 @@ def _cleanup():
...
 
 @retry(exceptions=IIBError, tries=3, logger=log)
-def _create_and_push_manifest_list(request_id, arches):
+def _create_and_push_manifest_list(request_id, arches, build_tags):
     """
     Create and push the manifest list to the configured registry.
 
     :param int request_id: the ID of the IIB build request
     :param iter arches: an iterable of arches to create the manifest list for
+    :param build_tags: list of extra tag to use for intermetdiate index image
```
pbortlov, line 119: Typo: 'intermetdiate'

pbortlov:
hi @midnightercz , could you please summarize where should be the build_tags? I already got confused about where the attribute is needed and why?

iib/web/models.py
```diff
@@ -1006,13 +1037,22 @@ def from_json(cls, kwargs, batch=None):
             raise ValidationError(f'"operators" should be a non-empty array of strings')
 
         cls._from_json(
-            request_kwargs, additional_required_params=['operators', 'from_index'], batch=batch
+            request_kwargs,
+            additional_required_params=['operators', 'from_index'],
+            additional_optional_params=['build_tags'],
```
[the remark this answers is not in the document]
yashvardhannanavati: It's not required because it's already present in `_from_json`

iib/web/models.py
```diff
@@ -930,8 +955,14 @@ def from_json(cls, kwargs, batch=None):
             request_kwargs[key] = [
                 Image.get_or_create(pull_specification=item) for item in request_kwargs.get(key, [])
             ]
-
+        build_tags = request_kwargs.pop('build_tags', [])
         request = cls(**request_kwargs)
+        db.session.add(request)
+        db.session.flush()
+
+        for bt in build_tags:
```
yashvardhannanavati, line 963: You have a function to do this. `add_build_tag`

iib/web/models.py
```diff
@@ -1006,13 +1037,22 @@ def from_json(cls, kwargs, batch=None):
...
+            additional_optional_params=['build_tags'],
+            batch=batch,
         )
 
         request_kwargs['operators'] = [Operator.get_or_create(name=item) for item in operators]
 
         request = cls(**request_kwargs)
         request.add_state('in_progress', 'The request was initiated')
+
+        build_tags = request_kwargs.pop('build_tags', [])
```
yashvardhannanavati, line 1051: Same, you have a function for this. `add_build_tag`

iib/web/models.py
```diff
@@ -1064,6 +1104,7 @@ class RequestRegenerateBundle(Request):
     __mapper_args__ = {
         'polymorphic_identity': RequestTypeMapping.__members__['regenerate_bundle'].value
     }
+    build_tags = None
```
yashvardhannanavati, line 1107: If it's present in `Request`, you can simply set `nullable=True` in Request and this won't be required.

iib/web/models.py
```diff
@@ -1269,6 +1317,12 @@ def from_json(cls, kwargs, batch=None):
         request_kwargs['batch'] = batch
 
         request = cls(**request_kwargs)
+
+        build_tags = request_kwargs.pop('build_tags', [])
+
+        for bt in build_tags:
```
yashvardhannanavati, line 1323: Same as above.

yashvardhannanavati (changes requested):

iib/web/models.py
midnightercz: yes, that's why I'm using it. Or I don't get your point

iib/web/models.py
midnightercz: hmm, I don't understand

iib/web/models.py
midnightercz: According to docs sqlalchemy.orm.relationship doesn't accept nullable attribute. Or do you mean something else?

iib/web/models.py
```diff
@@ -930,8 +955,14 @@ def from_json(cls, kwargs, batch=None):
             request_kwargs[key] = [
                 Image.get_or_create(pull_specification=item) for item in request_kwargs.get(key, [])
             ]
-
+        build_tags = request_kwargs.pop('build_tags', [])
```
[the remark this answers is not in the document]
midnightercz: fixed

iib/web/models.py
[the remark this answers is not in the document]
midnightercz: fixed

iib/workers/tasks/build.py
midnightercz: fixed

iib/web/models.py
yashvardhannanavati: Sorry I think I misunderstood something here.

iib/web/models.py
```diff
@@ -1006,13 +1035,21 @@ def from_json(cls, kwargs, batch=None):
             raise ValidationError(f'"operators" should be a non-empty array of strings')
 
         cls._from_json(
-            request_kwargs, additional_required_params=['operators', 'from_index'], batch=batch
+            request_kwargs,
+            additional_required_params=['operators', 'from_index'],
+            batch=batch,
```
yashvardhannanavati, line 1040: If you remove the last comma, it will go back to as it was before. I'm guessing this change was unintentional and Python black formatting.

iib/workers/tasks/build.py
```diff
@@ -106,59 +106,70 @@ def _cleanup():
 
 
 @retry(exceptions=IIBError, tries=3, logger=log)
-def _create_and_push_manifest_list(request_id, arches):
+def _create_and_push_manifest_list(request_id, arches, build_tags):
```
yashvardhannanavati, line 109: This will have to be refactored as a result of https://github.com/release-engineering/iib/pull/286

iib/workers/tasks/build.py
```diff
@@ -838,7 +850,13 @@ def handle_add_request(
         ``binary_image`` to use.
     :param list deprecation_list: list of deprecated bundles for the target index image. Defaults
         to ``None``.
+<<<<<<< HEAD
```
yashvardhannanavati, line 853: I think something got messed up in the rebase. Could you please fix it

yashvardhannanavati (changes requested):
Very close to merging. Could you please refactor the `_create_and_push_manifest_list` function and fix the rebase error.

iib/workers/tasks/build.py
midnightercz: sure

iib/workers/tasks/build.py
```diff
@@ -105,42 +105,54 @@ def _cleanup():
...
-    )
+        log.debug('Pushing manifest list %s', output_pull_spec)
+        run_cmd(
+            buildah_manifest_cmd
+            + ['push', '--all', output_pull_spec, f'docker://{output_pull_spec}'],
+            exc_msg=f'Failed to push the manifest list to {output_pull_spec}',
+        )
 
-    return output_pull_spec
+    return output_pull_spec[0]
```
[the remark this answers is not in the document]
midnightercz: that's correct

pbortlov (approved):

lipoja (approved):

iib/workers/tasks/build.py
```diff
@@ -105,42 +105,54 @@ def _cleanup():
...
+    output_pull_specs = [get_rebuilt_image_pull_spec(request_id)]
+    if build_tags:
+        conf = get_worker_config()
+        for tag in build_tags:
+            output_pull_specs.append(
+                conf['iib_image_push_template'].format(
+                    registry=conf['iib_registry'], request_id=tag
+                )
+            )
+    for output_pull_spec in output_pull_specs:
```
yashvardhannanavati, line 129: Why does this for loop create a new manifest list for every tag? Can't we simply put this for loop just before line 149? That way we simply push the same manifest list with different tags.

midnightercz: yeah, I can improve that

pbortlov (dismissed):

iib/web/models.py
```diff
@@ -241,6 +241,15 @@ def get_or_create(cls, name):
         return operator
 
 
+class BuildTag(db.Model):
```
yashvardhannanavati, line 244: I'm sorry I missed this earlier. But could you please use an association table like`RequestArchitecture` to be consistent. Something like `RequestBuildTag`

iib/web/models.py
```diff
@@ -360,6 +370,15 @@ def add_state(self, state, state_reason):
         db.session.flush()
         self.request_state_id = request_state.id
 
+    def add_build_tag(self, name):
```
yashvardhannanavati, line 373: Also, shouldn't we be checking if a build tag already exists before adding a new one associated with the request? Check out `add_architecture` for inspiration.

midnightercz: Yeah, we should. I thought sqlalchemy is smart enough to figure this out

iib/web/models.py
```diff
@@ -360,6 +370,20 @@ def add_state(self, state, state_reason):
...
+        Add a RequestBuildTag associated with the current request.
+
+        :param str name: tag name
+        """
+        bt = db.session.query(RequestBuildTag).filter_by(name=name).first()
+        if not bt:
+            bt = RequestBuildTag(name=name, request_id=self.id)
+            db.session.add(bt)
+            db.session.flush()
+        else:
```
yashvardhannanavati, line 384: This should probably be `if bt not in self.build_tags`

iib/web/models.py
```diff
@@ -241,6 +241,15 @@ def get_or_create(cls, name):
         return operator
 
 
+class RequestBuildTag(db.Model):
```
yashvardhannanavati, line 244: @midnightercz  I believe we talked about creating an association table here. `RequestBuildTag` should only be associating the two tables. Like `RequestArchitecture` on line 158. 

Another table called the `BuildTag` should house the `id` and `name` like the `Architecture` table. Could you please change it?

yashvardhannanavati (changes requested):

iib/web/models.py
midnightercz: no, it's not needed if you set request_id=self.id. You can check newly added tests.

iib/web/models.py
midnightercz: I get your point, but unlike architecture which can be chosen only from limited set of choices, build tag can be any string and most probably even will be unique for every request. Does it make sense then to create enumeration table and association table?

yashvardhannanavati: It would be nice to add a `__repr__` function too.

iib/web/models.py
```diff
@@ -241,13 +241,21 @@ def get_or_create(cls, name):
...
-class RequestBuildTag(db.Model):
+class BuildTag(db.Model):
     """Extra tag associated with built index image."""
 
     id = db.Column(db.Integer, primary_key=True)
     name = db.Column(db.String, nullable=False, unique=False)
+
+
+class RequestBuildTag(db.Model):
+    """Extra tag associated with built index image."""
```
pbortlov, line 252: This line is the same as on line no. 245. Is it really the same?

iib/workers/tasks/build.py
```diff
@@ -105,42 +105,54 @@ def _cleanup():
...
-    )
+        log.debug('Pushing manifest list %s', output_pull_spec)
+        run_cmd(
+            buildah_manifest_cmd
+            + ['push', '--all', output_pull_spec, f'docker://{output_pull_spec}'],
+            exc_msg=f'Failed to push the manifest list to {output_pull_spec}',
+        )
 
-    return output_pull_spec
+    return output_pull_specs[0]
```
pbortlov, line 155: Could you add a comment on why is taken the first item of the list?

iib/workers/tasks/build.py
```diff
@@ -971,7 +985,7 @@ def handle_add_request(
         )
 
     set_request_state(request_id, 'in_progress', 'Creating the manifest list')
-    output_pull_spec = _create_and_push_manifest_list(request_id, arches)
+    output_pull_spec = _create_and_push_manifest_list(request_id, arches, build_tags)
```
pbortlov, line 988: Is it singular or plural? Do we get the one pull spec or more? I guess one but better to ask

iib/workers/tasks/build_merge_index_image.py
```diff
@@ -272,9 +275,11 @@ def handle_merge_request(
             file_mode=(stat.S_IRUSR | stat.S_IWUSR | stat.S_IRGRP | stat.S_IWGRP),
         )
 
-    output_pull_spec = _create_and_push_manifest_list(request_id, prebuild_info['arches'])
+    output_pull_specs = _create_and_push_manifest_list(
```
pbortlov, line 278: Here is plural and in the file `build.py` is just `output_pull_spec` instead of `spec**s**`

pbortlov: it seems there is a discrepancy. THe function `_create_and_push_manifest_list` returns `return output_pull_specs[0]`. You are using it differently here.

iib/web/models.py
midnightercz: right, I'll fix the description

iib/workers/tasks/build.py
midnightercz: You mean in the comment or here? First item is taken as first item is production tag, next items in the list are extra build tags

pbortlov: Comment in code :)

iib/workers/tasks/build_create_empty_index.py
```diff
@@ -70,6 +70,7 @@ def handle_create_empty_index_request(
     :param dict labels: the dict of labels required to be added to a new index image
     :param dict binary_image_config: the dict of config required to identify the appropriate
         ``binary_image`` to use.
+    :param list build_tags: List of tags which will be applied to intermediate index images.
```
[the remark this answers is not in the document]
midnightercz: removed

tests/test_web/test_api_v1.py
```diff
@@ -1842,6 +1852,7 @@ def test_create_empty_index_success(
         'batch_annotations': None,
         'binary_image': binary_image,
         'binary_image_resolved': None,
+        'build_tags': [],
```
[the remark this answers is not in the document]
midnightercz: Yeah, it's fro generic request schema. This is why I didn't want to add it to generic _from_json which Yash insisted on.

yashvardhannanavati (dismissed):
We just need to confirm that this has been tested on the local dev env.

yashvardhannanavati:
@midnightercz Could you please rebase this?

yashvardhannanavati:
@midnightercz Could you please squash your commits :)

iib/workers/tasks/build.py
```diff
@@ -106,42 +106,60 @@ def _cleanup():
...
 
-    log.debug('Pushing manifest list %s', output_pull_spec)
-    run_cmd(
-        buildah_manifest_cmd + ['push', '--all', output_pull_spec, f'docker://{output_pull_spec}'],
-        exc_msg=f'Failed to push the manifest list to {output_pull_spec}',
-    )
+    conf = get_worker_config()
+    for tag in _tags:
+        for output_pull_spec in output_pull_specs:
+            output_pull_spec = conf['iib_image_push_template'].format(
```
pbortlov, line 129: Could you please add a comment here on why you iterate on  output_pull_spec and rewriting it?

yashvardhannanavati (dismissed):

pbortlov (approved):

PR #250, opened 2021-05-14
API endpoint and db model create-empty-index

pbortlov:
* add create-empty-index API endpoint
* add model RequestCreateEmptyIndex

PR #260, opened 2021-06-23
WIP adding OCP max version to bundles

gallettilance:
Not sure this is the right place for it but we need some regular job running this check in order to ensure this property is persisted (since many actors can rebuild the graph and this label is not in the bundles).


iib/workers/tasks/build.py
```diff
@@ -496,6 +497,65 @@ def _opm_index_add(
...
+    property["operatorbundle_path"] = bundle
+    cur.close()
+    # Insert property
+    con.execute(insert, (property["type"], property["value"], property["operatorbundle_version"], property["operaotorbundle_path"]))
+    con.close()
+
+
+def _sdk_is_valid(bundle):
+    cmd = [
+        'operator-sdk',
```
gallettilance, line 545: requires sdk > 1.8.0

iib/workers/tasks/build.py
```diff
@@ -496,6 +497,65 @@ def _opm_index_add(
...
+        bundle,
+        '--select-optional',
+        'name=community',
+        '--output=json-alpha1'
+    ]
+    try:
+        run_cmd(cmd, {}, exc_msg='operator-sdk command failed')
+        return True
+    except IIBError:
+        # TODO: check that the error is related to v1beta1 api removal
```
gallettilance, line 557: we can process the output which should look something like this:
```
ERRO[0001] Error: Value : (memcached-operator.v0.0.1) csv.Annotations not specified olm.maxOpenShiftVersion for an OCP version < 4.9. This annotation is required to prevent the user from upgrading their OCP cluster before they have installed a version of their operator which is compatible with 4.9. This bundle is using APIs which were deprecated and removed in v1.22. More info: https://kubernetes.io/docs/reference/using-api/deprecation-guide/#v1-22 which are no longer supported on 4.9. Migrate the API(s) for CRD: (["memcacheds.cache.example.com"]),ClusterRole: (["memcached-operator-metrics-reader"]), or use the annotation 
```

iib/workers/tasks/build.py
gallettilance: not sure where to add the sdk binary to iib

iib/workers/tasks/build.py
gallettilance: will tackle that tomorrow

iib/workers/tasks/build.py
```diff
@@ -496,6 +497,65 @@ def _opm_index_add(
...
+        raise IIBError("Error getting version for bundle")
+    property["operatorbundle_version"] = rows[0][0]
+    property["operatorbundle_path"] = bundle
+    cur.close()
+    # Insert property
+    con.execute(insert, (property["type"], property["value"], property["operatorbundle_version"], property["operaotorbundle_path"]))
+    con.close()
+
+
+def _sdk_is_valid(bundle):
```
camilamacedo86, line 543:
```suggestion
// Check if the bundle is using deprecated APIs and has not the max OCP version set = 4.8
def _requires_max_ocp_version(bundle):
```

iib/workers/tasks/build.py
```diff
@@ -496,6 +497,65 @@ def _opm_index_add(
...
+    # add properties to bundles in the index
+    if add_properties:
+        maxOpenShiftVersionProperty = {
+            "type": "olm.maxOpenShiftVersion",
+            "value": "4.8",
+            "operatorbundle_version": "",
+            "operatorbundle-path": "",
+        }
+        for bundle in bundle_str.split(','):
+            if _sdk_is_valid(bundle):
```
camilamacedo86, line 510:
```suggestion
            if _requires_max_ocp_version(bundle):
```

iib/workers/tasks/build.py
```diff
@@ -496,6 +497,65 @@ def _opm_index_add(
...
+                _add_property_to_index(base_dir, bundle, maxOpenShiftVersionProperty)
+
+
+@retry(attempts=2, wait_on=IIBError, logger=log)
+def _add_property_to_index(base_dir, bundle, property):
```
camilamacedo86, line 515:
```suggestion
// Add the property max ocp version. This function is required for we are able to avoid workflow issues. 
// We need to ensure that any bundle which has deprecated/removed API(s) in 1.22/ocp 4.9 will be using
// the property to not allow the users to upgrade the cluster to 4.9 before upgrade the operator installed
// to a version that is compatible with 4.9(migrated)
@retry(attempts=2, wait_on=IIBError, logger=log)
def _add_property_to_index(base_dir, bundle, property):
```

iib/workers/tasks/build.py
camilamacedo86: Yep. I think we can check if the errors[] has an string ` not specified olm.maxOpenShiftVersion` to return true here.

iib/workers/tasks/build.py
```diff
@@ -496,6 +497,65 @@ def _opm_index_add(
...
+        'operator-sdk',
+        'bundle',
+        'validate',
+        bundle,
+        '--select-optional',
+        'name=community',
+        '--output=json-alpha1'
+    ]
+    try:
+        run_cmd(cmd, {}, exc_msg='operator-sdk command failed')
```
camilamacedo86, line 554: Will not that return true for any error? Currently, we have no other checks on that. It will only check this scenario but I think we need to ensure if the error contains ` not specified olm.maxOpenShiftVersion`. Unfortunately, operator-framework/api does not allow us to return for example codes for the errors such as we have for HTTP request errors. That would be better in this case, however, we can still try to check the string.

iib/workers/tasks/build.py
camilamacedo86: nit: I think would be better we add a comment and a name that clarifies better its intention.

iib/workers/tasks/build.py
camilamacedo86: nit: I think would be better we add a comment and a name that clarifies better its intention.

iib/workers/tasks/build.py
camilamacedo86: nit: I think would be better we add a comment and a name that clarifies better its intention.

iib/workers/tasks/build.py
gallettilance: yes that is why I have the "TODO" comment :) planning to do that today

iib/workers/tasks/build.py
```diff
@@ -496,6 +497,65 @@ def _opm_index_add(
 
     with set_registry_token(overwrite_from_index_token, from_index):
         run_cmd(cmd, {'cwd': base_dir}, exc_msg='Failed to add the bundles to the index image')
+    
+    # add properties to bundles in the index
+    if add_properties:
```
bentito, line 502: How is `add_properties` ever not `false`? If it's set externally, maybe the comment should make clear how?

iib/workers/tasks/build.py
```diff
@@ -496,6 +497,65 @@ def _opm_index_add(
...
+            "operatorbundle-path": "",
+        }
+        for bundle in bundle_str.split(','):
+            if _sdk_is_valid(bundle):
+                _add_property_to_index(base_dir, bundle, maxOpenShiftVersionProperty)
+
+
+@retry(attempts=2, wait_on=IIBError, logger=log)
+def _add_property_to_index(base_dir, bundle, property):
+    import sqlite3
```
bentito, line 516: seems like this project organizes imports at top of file

iib/workers/tasks/build.py
bentito: oh nvm, I saw it was passed in. Just needs :param to doc it

iib/workers/tasks/build.py
```diff
@@ -442,6 +442,7 @@ def _opm_index_add(
     overwrite_from_index_token=None,
     overwrite_csv=False,
     container_tool=None,
+    add_properties=False,
```
bentito, line 445: This should get a `:param` entry

iib/workers/tasks/build.py
```diff
@@ -496,6 +497,65 @@ def _opm_index_add(
...
+            "value": "4.8",
+            "operatorbundle_version": "",
+            "operatorbundle-path": "",
+        }
+        for bundle in bundle_str.split(','):
+            if _sdk_is_valid(bundle):
+                _add_property_to_index(base_dir, bundle, maxOpenShiftVersionProperty)
+
+
+@retry(attempts=2, wait_on=IIBError, logger=log)
```
bentito, line 514: There is a config value for this, 2 is overriding that. Good reason to override?

iib/workers/tasks/build.py
bentito, line 516: This project seems to group imports at the tops of files

iib/workers/tasks/build.py
```diff
@@ -496,6 +497,65 @@ def _opm_index_add(
...
+    cur = con.cursor()
+    cur.execute("SELECT version FROM operatorbundle WHERE bundlepath=?", (bundle,))
+    rows = cur.fetchall()
+    if len(rows) != 1:
+        raise IIBError("Error getting version for bundle")
+    property["operatorbundle_version"] = rows[0][0]
+    property["operatorbundle_path"] = bundle
+    cur.close()
+    # Insert property
+    con.execute(insert, (property["type"], property["value"], property["operatorbundle_version"], property["operaotorbundle_path"]))
```
bentito, line 539: I think if you change to using `with`, like `with con:` you can then get rid of these clean up lines

iib/workers/tasks/build.py
```diff
@@ -496,6 +497,65 @@ def _opm_index_add(
...
+    con.close()
```
bentito, line 540: I think if you change to using `with`, like `with con:` you can then get rid of these clean up lines

iib/workers/tasks/build.py
gallettilance: the idea is just to make that an option in the api so that whatever job runs this command can set that as true.

iib/workers/tasks/build.py
```diff
@@ -8,6 +8,7 @@
 import time
 import tempfile
 import textwrap
+import sqlite3
```
yashvardhannanavati, line 11: Could you please re-order it alphabetically?

iib/workers/tasks/build.py
```diff
@@ -720,6 +721,53 @@ def _verify_index_image(
...
+        property["value"],
+        property["operatorbundle_name"],
+        property["operatorbundle_version"],
+        property["operatorbundle_path"]))
+    con.commit()
+    con.close()
+
+
+# Check if the bundle is using deprecated APIs AND does not have the max OCP version set to 4.8
+def _requires_max_ocp_version(bundle):
```
yashvardhannanavati, line 749: Could you please add the docstring here?

iib/workers/tasks/build.py
```diff
@@ -840,6 +888,36 @@ def handle_add_request(
...
+            ['grpcurl', '-plaintext', f'localhost:{port}', 'api.Registry/ListBundles'],
+            exc_msg='Failed to get bundle data from index image',
+        )
+        rpc_proc.kill()
+
+        # Transform returned data to parsable json
+        present_bundles = [json.loads(bundle) for bundle in re.split(r'(?<=})\n(?={)', rawBundles)]
+        bundles = list(filter(lambda b : b['bundlePath'] in resolved_bundles, present_bundles))
+        
+        for bndl in bundles:
```
yashvardhannanavati, line 910: Sorry for the knitpicking 
`for bndl in bundles` => `for bundle in bundles`

iib/workers/tasks/build.py
```diff
@@ -840,6 +888,36 @@ def handle_add_request(
             (prebuild_info['distribution_scope'] in ['dev', 'stage']),
         )
 
+        # Add the max ocp version property
+        # We need to ensure that any bundle which has deprecated/removed API(s) in 1.22/ocp 4.9
+        # will have this property to prevent users from upgrading clusters to 4.9 before upgrading
+        # the operator installed to a version that is compatible with 4.9
+
+        # Get the CSV name and version (not just the bundle path)
+        db_path = temp_dir + "/index.db"
```
yashvardhannanavati, line 897: Just a heads up #250 is going to create a helper function which can be used here and will avoid code duplication. :)

iib/workers/tasks/build.py
```diff
@@ -840,6 +888,36 @@ def handle_add_request(
...
+
+        rawBundles = run_cmd(
+            ['grpcurl', '-plaintext', f'localhost:{port}', 'api.Registry/ListBundles'],
+            exc_msg='Failed to get bundle data from index image',
+        )
+        rpc_proc.kill()
+
+        # Transform returned data to parsable json
+        present_bundles = [json.loads(bundle) for bundle in re.split(r'(?<=})\n(?={)', rawBundles)]
+        bundles = list(filter(lambda b : b['bundlePath'] in resolved_bundles, present_bundles))
```
yashvardhannanavati, line 908: This will overwrite the `bundles` argument. I don't think we want that.

iib/workers/tasks/build.py
yashvardhannanavati: On second thoughts, do you need this piece of code, based on our conversation earlier, we only need to run this on bundles that we are adding in a request. So you can simply use the `bundles` variable that's coming in to apply the property on.

yashvardhannanavati (changes requested):

iib/workers/tasks/build.py
gallettilance: in order to add the property I need the csv name and the version but bundles only gives me the pull spec. I figured instead of parsing the csv from the bundle I would just ask the db what these values are. What do you think?

yashvardhannanavati: Yeah, I think that makes sense. `_get_missing_bundles ` is already calling grpcurl maybe you can do everything in one call instead of calling it again?

gallettilance: it looks like the function is dependent on `_get_present_bundles` but this is being called before the new bundles (from the IIB request) are added to the index. I just need the name and version for the bundles from the IIB request that aren't in the index at that time unfortunately.

iib/workers/tasks/build.py
```diff
@@ -720,6 +721,61 @@ def _verify_index_image(
         )
 
 
+def _add_property_to_index(db_path, property):
+    """
+    Add a property to the index.
+
+    :param str base_dir: the base directory where the database and index.Dockerfile are created
```
querti, line 728: Should this be removed?

iib/workers/tasks/build.py
```diff
@@ -720,6 +721,61 @@ def _verify_index_image(
...
+    :param str bundle: the bundle path
```
querti, line 729: Wrong param name?

iib/workers/tasks/build.py
```diff
@@ -720,6 +721,61 @@ def _verify_index_image(
...
+            property["operatorbundle_name"],
+            property["operatorbundle_version"],
+            property["operatorbundle_path"],
+        ),
+    )
+    con.commit()
+    con.close()
+
+
+# Check if the bundle is using deprecated APIs AND does not have the max OCP version set to 4.8
```
querti, line 750: Adding a comment here is a bit strange. Maybe this info could be added to the docstring?

iib/workers/tasks/build.py
```diff
@@ -720,6 +721,61 @@ def _verify_index_image(
...
+def _requires_max_ocp_version(bundle):
+    """
+    Whether the bundle requires the olm.maxOpenShiftVersion property added to the db.
```
querti, line 753: Could you reword this to be imperative?

iib/workers/tasks/build.py
```diff
@@ -720,6 +721,61 @@ def _verify_index_image(
...
+        '--output=json-alpha1',
+        '-b',
+        'podman',
+    ]
+    result = run_cmd(cmd, strict=False)
+    if result:
+        output = json.loads(result)
+        if not output['passed']:
+            for msg in output['outputs']:
+                if "olm.maxOpenShiftVersion" in msg['message']:
```
querti, line 774: Maybe I'm misunderstanding this semantically. Is True returned when the bundle doesn't require this property to be added? Shouldn't it be the other way around?

iib/workers/tasks/build.py
```diff
@@ -720,6 +721,61 @@ def _verify_index_image(
...
+    con.commit()
+    con.close()
+
+
+# Check if the bundle is using deprecated APIs AND does not have the max OCP version set to 4.8
+def _requires_max_ocp_version(bundle):
+    """
+    Whether the bundle requires the olm.maxOpenShiftVersion property added to the db.
+
+    :param list bundles: a list of dicts representing bundles
```
querti, line 755: Is the method wrong or the docstring?

iib/workers/tasks/build.py
```diff
@@ -720,6 +721,61 @@ def _verify_index_image(
...
+        'validate',
+        bundle,
+        '--select-optional',
+        'name=community',
+        '--output=json-alpha1',
+        '-b',
+        'podman',
+    ]
+    result = run_cmd(cmd, strict=False)
+    if result:
```
querti, line 770: What does this mean? That the command failed? Is that ok?

iib/workers/tasks/build.py
```diff
@@ -840,6 +896,40 @@ def handle_add_request(
...
+        updated_present_bundles = [
+            json.loads(bundle) for bundle in re.split(r'(?<=})\n(?={)', rawBundles)
+        ]
+        updated_bundles = list(
+            filter(lambda b: b['bundlePath'] in resolved_bundles, updated_present_bundles)
+        )
+
+        for bundle in updated_bundles:
+            if _requires_max_ocp_version(bundle['bundlePath']):
+                maxOpenShiftVersionProperty = {
```
querti, line 924: The convention for Python variables are underscores, not camel case

iib/workers/tasks/build.py
```diff
@@ -840,6 +896,40 @@ def handle_add_request(
...
+                    "type": "olm.maxOpenShiftVersion",
+                    "value": "4.8",
+                    "operatorbundle_name": bundle["csvName"],
```
querti, line 927: I think IIB uses single quotes as a standard. Could you change your  " into ' ?

tests/test_workers/test_tasks/test_build.py
```diff
@@ -436,6 +438,8 @@ def test_handle_add_request(
     mock_cleanup,
     mock_ugrb,
     mock_dep_b,
+    mock_sir,
```
querti, line 441: Not sure if IIB has a requirement on this, but shouldn't unit tests be added for the two new functions?

querti (changes requested):

iib/workers/tasks/build.py
gallettilance: ah good to know! coming from golang :)

iib/workers/tasks/build.py
gallettilance: added some comments to clarify - let me know if that helps!

tests/test_workers/test_tasks/test_build.py
gallettilance: happy to add unit tests - will take a look at that today!

iib/workers/tasks/build.py
gallettilance: added some comments to clarify - hope it helps!

iib/workers/tasks/build.py
```diff
@@ -720,6 +721,66 @@ def _verify_index_image(
...
+    result = run_cmd(cmd, strict=False)
+    if result:
+        output = json.loads(result)
+        # check if the bundle validation failed
+        if not output['passed']:
+            # check if the failure is due to the presence of deprecated APIs
+            # and absence of the 'olm.maxOpenShiftVersion' property
+            # Note: there is no other error in the sdk that mentions this field
+            for msg in output['outputs']:
+                if 'olm.maxOpenShiftVersion' in msg['message']:
```
camilamacedo86, line 779: ihmo we need to improve it. 
The validator can return an error with that if we have the property set and its value is > 4.8 in this case. 
So, before inset the data in the database we need to check if that already exists and if yes, then update it.

camilamacedo86 (changes requested):
IHMO: we have a few nits to address
1 -  https://github.com/release-engineering/iib/pull/260/files#r663109951 
2 - We need to have a way to know what bundles we are sitting directly in the index in case we need to re-check, rollback or indeed verify if that was set but was lost as well at some point. Could we keep a yaml file with all bundle names that we are adding/updating the property?

iib/workers/tasks/build.py
gallettilance: I think what we had discussed was that we would not change what the operator author has explicitly set.

gallettilance: Note: this logic is only for new bundles being added not old bundles (the one-time script will take care of those)

camilamacedo86: We are adding it to be called in a method to do the build of the bundle. It might be called after as well. What happens if it is called twice (rebuild for e.g.)?  What happens if the user informed of this value in the CSV? This property is new but we are adding this information in the docs then, operator authors might start to use it and they can add invalid information and the error will be raised as well. 

In this way, IMHO we need to cover these scenarios. So, could we not only add automatically the values when we know that it does not have anything set in the index DB? If the user informed that will not be that in the index DB too? 

- Check the error 
- Before insert do a select in the index DB.
- If we found the property set already we need to skip and track this scenario.

iib/workers/tasks/build.py
```diff
@@ -720,6 +721,66 @@ def _verify_index_image(
...
+                    return True
+    return False
```
jianzhangbjz, line 781: If the `output['passed']` returns `true`, so this `return False` will be run, I guess it's not as expected, or am I missing something? Thanks!

gallettilance: If the bundle passes the set of tests then it does not require (or the operator-sdk cannot detect that it requires) the `olm.maxOpenShiftVersion` label

jianzhangbjz: Yes, thanks! Sorry, my misunderstanding.

README.md
```diff
@@ -360,6 +360,27 @@ them to the index image. If a Greenwave configuration is setup for your queue, I
 Greenwave to check if your bundle image builds have passed the tests in the Greenwave policy you
 have defined. The IIB request submitted to that queue will succeed only if the policy is satisfied.
 
+## Troubleshooting
```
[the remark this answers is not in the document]
gallettilance: will do!

docker/Dockerfile-workers
```diff
@@ -25,7 +25,7 @@ RUN dnf -y install \
     systemd-devel \
     && dnf clean all
 
-ENV OPM_SRC="/src"
+ENV OPM_SRC="/src/operator-registry"
```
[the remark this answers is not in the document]
gallettilance: I guess it's not needed it just seemed strange to clone the operator-registry repo in src/ instead of in a dedicated nested folder. I can revert that if you prefer.

iib/workers/tasks/build.py
```diff
@@ -720,6 +721,66 @@ def _verify_index_image(
...
+    cmd = [
+        'operator-sdk',
+        'bundle',
+        'validate',
+        bundle,
+        '--select-optional',
+        'name=community',
+        '--output=json-alpha1',
+        '-b',
+        'podman',
```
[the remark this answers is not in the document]
gallettilance: +1

docker/Dockerfile-workers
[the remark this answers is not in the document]
lcarva: Nah, it's fine. I was just curious. A separate commit might be nice :)

iib/workers/tasks/build.py
```diff
@@ -840,6 +901,42 @@ def handle_add_request(
...
         )
 
+        # Add the max ocp version property
+        # We need to ensure that any bundle which has deprecated/removed API(s) in 1.22/ocp 4.9
+        # will have this property to prevent users from upgrading clusters to 4.9 before upgrading
+        # the operator installed to a version that is compatible with 4.9
+
+        # Get the CSV name and version (not just the bundle path)
+        db_path = temp_dir + "/database/index.db"
+        port, rpc_proc = _serve_index_registry(db_path)
```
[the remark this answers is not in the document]
gallettilance: I'm not sure I follow. My reasoning is that we need the new bundles' `csvName` and `version` in order to add the properties to the index. Instead of untaring the bundle images and parsing the CSV of these bundles I thought I would query the database after they were added so I could get a nice json representation and extract these fields.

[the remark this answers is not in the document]
lcarva: I see the reason for the second call now. We need to serve the index again, because we want to query for the information of the newly added bundles.

iib/workers/tasks/build.py
```diff
@@ -840,6 +908,39 @@ def handle_add_request(
...
+
+        rawBundles = run_cmd(
+            ['grpcurl', '-plaintext', f'localhost:{port}', 'api.Registry/ListBundles'],
+            exc_msg='Failed to get bundle data from index image',
+        )
+        rpc_proc.kill()
+
+        # Get bundle json for bundles in the request
+        updated_bundles = list(
```
gallettilance, line 927: cc @lcarva

lcarva (approved):

iib/workers/tasks/build.py
```diff
@@ -720,6 +725,69 @@ def _verify_index_image(
...
+        output = json.loads(result)
+        # check if the bundle validation failed
+        if not output['passed']:
+            # check if the failure is due to the presence of deprecated APIs
+            # and absence of the 'olm.maxOpenShiftVersion' property
+            # Note: there is no other error in the sdk that mentions this field
+            for msg in output['outputs']:
+                if 'olm.maxOpenShiftVersion' in msg['message']:
+                    return True
+    return False
```
camilamacedo86, line 788: HI @gallettilance, 

Be aware that the message will be different for each case scenario, Ditto, so it is missing to work properly with. See that in the following cases we ought to not overwritten the info provided by the author:

1.- If the bundle has deprecated APIs and the property is informed and with a value informed which is >= 4.9 
```
 {
            "type": "error",
            "message": "Error: Value : (etcdoperator.v0.9.4) csv.Annotations.olm.properties with the key and value for olm.maxOpenShiftVersion has the OCP version value 4.9 which is \u003e= of 4.9. This bundle is using APIs which were deprecated and removed in v1.22. More info: https://kubernetes.io/docs/reference/using-api/deprecation-guide/#v1-22. Migrate the API(s) for CRD: ([\"etcdbackups.etcd.database.coreos.com\" \"etcdclusters.etcd.database.coreos.com\" \"etcdrestores.etcd.database.coreos.com\"]) or inform in this property an OCP version which is \u003c 4.9"
        }
```

2.- If the bundle has deprecated APIs and the property is informed and has an invalid value (attention to this scenario)```
  {
            "type": "error",
            "message": "Error: Value : (etcdoperator.v0.9.4) csv.Annotations.olm.properties has an invalid value.Unable to parse (invalid) using semver : Invalid character(s) found in major number \"invalid\""
        }
```

Comment (gallettilance) [reply iib/workers/tasks/build.py:788]: discussed offline: we will track these scenarios as bugs / enhancements for the coming sprint. cc @yashvardhannanavati

File: iib/workers/tasks/build.py [hunk 36]
```diff
@@ -840,6 +908,39 @@ def handle_add_request(
...
+        # We need to ensure that any bundle which has deprecated/removed API(s) in 1.22/ocp 4.9
+        # will have this property to prevent users from upgrading clusters to 4.9 before upgrading
+        # the operator installed to a version that is compatible with 4.9
+
+        # Get the CSV name and version (not just the bundle path)
+        db_path = temp_dir + "/database/index.db"
+        port, rpc_proc = _serve_index_registry(db_path)
+
+        rawBundles = run_cmd(
+            ['grpcurl', '-plaintext', f'localhost:{port}', 'api.Registry/ListBundles'],
```
midnightercz, line 921: I wonder if we should mention this dependency somewhere

iib/workers/tasks/build.py
```diff
@@ -840,6 +908,39 @@ def handle_add_request(
...
+        updated_bundles = list(
+            filter(lambda b: b['bundlePath'] in resolved_bundles, _get_bundle_json(rawBundles))
+        )
+
+        for bundle in updated_bundles:
+            if _requires_max_ocp_version(bundle['bundlePath']):
+                log.info('adding property for %s', bundle['bundlePath'])
+                max_openshift_version_property = {
+                    'type': 'olm.maxOpenShiftVersion',
+                    'value': '4.8',
```
midnightercz, line 936: can this be configurable? Or rather, can this be changed in the future? If yes, lets make it configurable

README.md
```diff
@@ -108,6 +108,27 @@ You may also run the development environment with
 branch as it has various fixes and new features required to run IIB. Set the environment variable
 `IIB_COMPOSE_ENGINE` to the path of the `podman-compose` script before running the `make` commands.
 
+### Troubleshooting
```
midnightercz, line 111: in the ideal world, this should be in separated PR. But it's just a minor

iib/workers/tasks/build.py
```diff
@@ -3,6 +3,7 @@
 import logging
 import os
 import re
+import sqlite3
```
midnightercz, line 6: this should be probably added to requirements.txt

midnightercz (changes requested):

iib/workers/tasks/build.py
gallettilance: on grpcurl?

iib/workers/tasks/build.py
gallettilance: There is no other valid configuration today than "4.8". If you are using APIs that will be removed in 4.9 then the max ocp version should be set to "4.8" and no other value. If that changes for some reason we should track it as a bug at that time IMO

iib/workers/tasks/build.py
gallettilance: +1 will do

iib/workers/tasks/build.py
midnightercz: yeah

iib/workers/tasks/build.py
```diff
@@ -720,6 +725,69 @@ def _verify_index_image(
...
+        'none',
+    ]
+    result = run_cmd(cmd, strict=False)
+    if result:
+        output = json.loads(result)
+        # check if the bundle validation failed
+        if not output['passed']:
+            # check if the failure is due to the presence of deprecated APIs
+            # and absence of the 'olm.maxOpenShiftVersion' property
+            # Note: there is no other error in the sdk that mentions this field
```
querti, line 784: I wonder, is this future-proof?

iib/workers/tasks/build.py
lipoja: @midnightercz Just FYI: We already depend on that - this was not newly introduced.
I do agree with you that we should have it mentioned somewhere. But I think we should do it for every non python dependency that we have (opm, podman, buildah, grpcurl, ...) - not part of this PR

iib/workers/tasks/build.py
```diff
@@ -840,6 +908,39 @@ def handle_add_request(
...
+        # Add the max ocp version property
+        # We need to ensure that any bundle which has deprecated/removed API(s) in 1.22/ocp 4.9
+        # will have this property to prevent users from upgrading clusters to 4.9 before upgrading
+        # the operator installed to a version that is compatible with 4.9
+
+        # Get the CSV name and version (not just the bundle path)
+        db_path = temp_dir + "/database/index.db"
+        port, rpc_proc = _serve_index_registry(db_path)
+
+        rawBundles = run_cmd(
```
querti, line 920: Could you use underscore convention here?

iib/workers/tasks/build.py
gallettilance: just realizing this is included in the standard library for python: https://stackoverflow.com/questions/19530974/how-can-i-add-the-sqlite3-module-to-python

iib/workers/tasks/build.py
```diff
@@ -720,6 +725,69 @@ def _verify_index_image(
...
+        'bundle',
+        'validate',
+        bundle,
+        '--select-optional',
+        'name=community',
+        '--output=json-alpha1',
+        '--image-builder',
+        'none',
+    ]
+    result = run_cmd(cmd, strict=False)
```
querti, line 777: Could you explain why the command is allowed to fail?

gallettilance: https://github.com/release-engineering/iib/pull/260#discussion_r664133524

iib/workers/tasks/build.py
midnightercz: ah, cool

midnightercz (approved):

iib/workers/tasks/build.py
querti: okay, but what if the operator-sdk command chokes on something and fails? Is that a tolerable state?

querti (commented):
LGTM, though I'm hesitant to give a +1 since I'm not overly familiar with the subject.

lipoja (approved):

yashvardhannanavati (approved):

PR #261, opened 2021-06-25
merge-index-image should deprecate invalid bundles instead of filtering

yashvardhannanavati:
Refers to CLOUDDST-6468

### DIFF
--- a/iib/workers/tasks/build_merge_index_image.py
+++ b/iib/workers/tasks/build_merge_index_image.py
@@ -65,12 +65,18 @@ def _add_bundles_missing_in_source(
     :param str overwrite_target_index_token: the token used for overwriting the input
         ``source_from_index`` image. This is required to use ``overwrite_target_index``.
         The format of the token must be in the format "user:password".
-    :return: bundles which were added to the index image.
-    :rtype: list
+    :return: tuple where the first value is a list of bundles which were added to the index image
+        and the second value is a list of bundles in the new index whose ocp_version range does not
+        satisfy the ocp_version value of the target index.
+    :rtype: tuple
     """
     set_request_state(request_id, 'in_progress', 'Adding bundles missing in source index image')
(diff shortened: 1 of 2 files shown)


iib/workers/tasks/build_merge_index_image.py
```diff
@@ -65,12 +65,17 @@ def _add_bundles_missing_in_source(
     :param str overwrite_target_index_token: the token used for overwriting the input
         ``source_from_index`` image. This is required to use ``overwrite_target_index``.
         The format of the token must be in the format "user:password".
-    :return: bundles which were added to the index image.
-    :rtype: list
+    :return: tuple of bundles which were added to the index image and the bundles in the new index
```
lcarva, line 68: Suggestion: `tuple where the first value is a list of bundles.... and the second value is a list of bundles...`

iib/workers/tasks/build_merge_index_image.py
```diff
@@ -97,25 +102,22 @@ def _add_bundles_missing_in_source(
...
+            invalid_bundles.append(bundle)
 
-    if missing_bundle_paths:
-        log.info('%s bundles are missing in the source index image.', len(missing_bundle_paths))
-    else:
+    if invalid_bundles:
         log.info(
-            'No bundles are missing in the source index image. However, the index image is '
-            'still being rebuilt with the new binary image.'
+            '%s bundles have invalid version label and will be deprecated.', len(invalid_bundles)
```
lcarva, line 113: Will the actual invalid bundles be logged anywhere?

yashvardhannanavati: Yes, when the `opm deprecatetruncate` command is called. It can be a large list and so I am trying to avoid logging it more than once

lcarva (approved):

gallettilance (approved):
lgtm

PR #262, opened 2021-06-29
Replace retry with retry package and set back-off interval for skopeo_inspect

lipoja:
Replace custom retry with retry package and setting back-off interval for skopeo_inspect

- replaced internal retry with retry package
- set back-off interval for skopeo_inspect with delay between retry 5, 10, 20, 40, 80 seconds

This change removes exception log message `'The maximum number of attempts (%s) have failed'` from logs.
Original (IIBError) exception is still raised.
And it changes retry message from `'Exception %r raised from %r.  Retrying now' `to `'%s, retrying in %s seconds...'`

refers to [CLOUDDST-6436]

### DIFF
--- a/iib/workers/tasks/build.py
+++ b/iib/workers/tasks/build.py
@@ -11,6 +11,7 @@ import tempfile
 import textwrap
 
 from operator_manifest.operator import ImageName
+from retry import retry
 
 from iib.exceptions import IIBError, AddressAlreadyInUse
 from iib.workers.api_utils import set_request_state, update_request
@@ -31,7 +32,6 @@ from iib.workers.tasks.utils import (
     podman_pull,
     request_logger,
     reset_docker_config,
-    retry,
     run_cmd,
     set_registry_token,
     skopeo_inspect,
@@ -105,7 +105,7 @@ def _cleanup():
     reset_docker_config()
 
 
-@retry(attempts=3, wait_on=IIBError, logger=log)
+@retry(exceptions=IIBError, tries=3, logger=log)
 def _create_and_push_manifest_list(request_id, arches):
     """
     Create and push the manifest list to the configured registry.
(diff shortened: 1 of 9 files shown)


lipoja:
I am suggesting to replace our custom retry function with retry from pypi. Which is tested and has a lot of features that we might use in feature. 
The downside is that logs will be changed. One log message from our retry is removed and another one is replaced by different text. For more detail please see commit message. 

Are  we OK with replacing/removing these log messages?

pbortlov (approved):

lipoja:
@yashvardhannanavati Could you have a look on this one, maybe also answer my questions. Thank you!

lipoja:
@querti @midnightercz could you have a look as well?

iib/workers/tasks/utils.py
```diff
@@ -447,7 +412,13 @@ def set_registry_auths(registry_auths):
         reset_docker_config()
 
 
-@retry(wait_on=IIBError, logger=log)
+@retry(
+    exceptions=IIBError,
+    tries=get_worker_config().iib_total_attempts,
+    delay=5,
```
querti, line 418: Could you explain why you chose these numbers?

querti (commented):
LGTM, just one minor comment

midnightercz (approved):

zanssa:
LGTM as well. 

> The downside is that logs will be changed. One log message from our retry is removed and another one is replaced by different text. For more detail please see commit message.

As long as the iibError messages show in the logs, I don't see this as a problem!

iib/workers/tasks/utils.py
lipoja: Because it looked reasonable to me. The delay will be 5, 10, 20, 40 and 80 seconds. 
Unfortunately I do not have any information about the time - how long does it take to recover.

yashvardhannanavati:
I am good with replacing the log messages. It does log that it's retrying and that's sufficient for me.

iib/workers/tasks/build.py
```diff
@@ -40,6 +39,7 @@
     prepare_request_for_build,
 )
 
+from retry import retry
```
yashvardhannanavati, line 42: I think this needs to go above the iib imports, no?

yashvardhannanavati (changes requested):
One minor comment. Looks good to me!

PR #265, opened 2021-07-13
Fix for memcachedKeyLengthError

lipoja:
memcache client supports keys with length <= 250.
It might happen that function name with all arguments is
longer then this limit. For this case we should create
SHA256 sum to have unique name that will be used to
determine correct cached value.

Error example: https://iib.engineering.redhat.com/api/v1/builds/90240/logs

refers to [CLOUDDST-7479]

### DIFF
--- a/iib/workers/dogpile_cache.py
+++ b/iib/workers/dogpile_cache.py
@@ -1,5 +1,6 @@
 # SPDX-License-Identifier: GPL-3.0-or-later
 import functools
+import hashlib
 
 from dogpile.cache import make_region
 
@@ -48,7 +49,12 @@ def generate_cache_key(fn, *args, **kwargs):
     arguments = '|'.join(
         [str(arg) for arg in args] + [f'{kwarg}={kwargs[kwarg]}' for kwarg in kwargs]
     )
-    return f'{fn}|{arguments}'
+    key_str = f'{fn}|{arguments}'
+    try:
+        key = hashlib.sha256(key_str).hexdigest()
+    except TypeError:
+        key = hashlib.sha256(key_str.encode('utf-8')).hexdigest()
+    return key
 
 
 def create_dogpile_region():
--- a/tests/test_workers/test_dogpile_cache.py
+++ b/tests/test_workers/test_dogpile_cache.py
@@ -1,0 +1,35 @@
+# SPDX-License-Identifier: GPL-3.0-or-later
+import pytest
+
+from iib.workers.dogpile_cache import generate_cache_key
+
+
+@pytest.mark.parametrize(
+    "args,kwargs",
+    [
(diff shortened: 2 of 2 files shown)


pbortlov (dismissed):

iib/workers/dogpile_cache.py
```diff
@@ -48,7 +49,13 @@ def generate_cache_key(fn, *args, **kwargs):
     arguments = '|'.join(
         [str(arg) for arg in args] + [f'{kwarg}={kwargs[kwarg]}' for kwarg in kwargs]
     )
-    return f'{fn}|{arguments}'
+    key = f'{fn}|{arguments}'
```
yashvardhannanavati, line 52: Do we need a test for this?

iib/workers/dogpile_cache.py
```diff
@@ -48,7 +49,13 @@ def generate_cache_key(fn, *args, **kwargs):
...
+    if len(key) >= 250:
```
midnightercz, line 53: why to use hash approach only when key is > 249 characters? Hash should work also with key < 250 right?

midnightercz (changes requested):

iib/workers/dogpile_cache.py
lipoja: I was thinking that this was the first time that this happen. And when we hash all the keys we would not be able to "debug" easily - I mean when we would investigate some issue we might want to get data from memcache - and it is easier to read the string created from function name and arguments then SHA number.

But if you think that it is better to count SHA for everything than I can change it.

yashvardhannanavati (approved):

iib/workers/dogpile_cache.py
lipoja: Since others agrees I changed that so SHA256 is always calculated.

midnightercz: > And when we hash all the keys we would not be able to "debug" easily

Hmm, have we ever done that actually? And I wonder if we are going to ever do that

midnightercz (dismissed):

tests/test_workers/test_dogpile_cache.py
```diff
@@ -0,0 +1,35 @@
...
+           natoque penatibus et magnis dis parturient montes, nascetur""",
+                "kwargs": """ridiculus mus. Donec quam felis, ultricies nec, pellentesque eu,
+           pretium quis, sem. Nulla consequat massa quis enim. Donec.""",
+            },
+        ),
+    ],
+)
+def test_generate_cache_key(args, kwargs):
+    passwd = generate_cache_key('function_name', *args, **kwargs)
+    assert len(passwd) <= 250
```
pbortlov, line 35: Is it still valid test when we removed it from code above?

pbortlov (approved):

PR #266, opened 2021-07-14
Set --arch when building images

lcarva:
This ensures the config blob contains the correct architecture.

* CLOUDDST-7520


lcarva:
I didn't a unit test would be helpful here. Let me know if you think otherwise.

chandwanitulsi (approved):

yashvardhannanavati (approved):

PR #269, opened 2021-07-16
Fix urllib3 vulnerability and add newer requests version [CLOUDDST-7548]

pbortlov:
update:
- urllib3
- requests

remove:
 - chardet

add:
- charset-normalizer


lcarva:
@pbortlov, I don't think we need to add the min version directive in `setup.py` and `requirements-test.in`. I see what you mean that if you just update `urllib3`, the version we want is not yet supported by `requests`. But you can update both packages at the same time instead. This seems to do the trick for me:
```
pip-compile --generate-hashes --output-file=requirements.txt -P requests -P urllib3
```

midnightercz (approved):

pbortlov:
@querti this PR is not ready. IT is hard to fix all the requirements. I will mark it as Draft

lipoja (approved):

querti (approved):

lcarva (approved):

PR #270, opened 2021-07-16
Add related_bundle_images API endpoint to fetch bundle images from re…

yashvardhannanavati:
…latedImages

section of bundle CSVs in regenerate-bundle requests

Refers to CLOUDDST-6892.

### DIFF
--- a/README.md
+++ b/README.md
@@ -184,10 +184,13 @@ The custom configuration options for the REST API are listed below:
 * `IIB_LOG_LEVEL` - the Python log level of the REST API (Flask). This defaults to `INFO`.
 * `IIB_MAX_PER_PAGE` - the maximum number of build requests that can be shown on a single page.
   This defaults to `20`.
+* `IIB_REQUEST_DATA_DAYS_TO_LIVE` - the amount of days after which per request temmporary data is
+  considered to be expired and may be removed. This defaults to `3`.
 * `IIB_REQUEST_LOGS_DIR` - the directory to load the request specific log files. If `None`, per
   request log files information will not appear in the API response. This defaults to `None`.
-* `IIB_REQUEST_LOGS_DAYS_TO_LIVE` - the amount of days after which per request logs are considered
-  to be expired and may be removed. This defaults to `3`.
+* `IIB_REQUEST_RELATED_BUNDLES_DIR` - the directory to load the request specific related
+  bundles files. If `None`, per request related bundles files information will not appear in
+  the API response. This defaults to `None`.
 * `IIB_USER_TO_QUEUE` - the mapping, `dict(<str>: <str>)`, of usernames to celery task queues.
   This is useful in isolating the workload from certain users. Some celery tasks must execute
   serially, while others can execute in parallel. Add the prefix `SERIAL:` or `PARALLEL:` to the
@@ -294,6 +297,12 @@ The custom configuration options for the Celery workers are listed below:
     is a string which specifies the glue to replace ``/`` (forward slashes) in the pull spec
     name and repo. The key `namespace` value is also a string which specifies the new namespace
     for pull specs of the ClusterServiceVersion files.
+  * The `related_bundles` customization type is a dictionary with no additional arguments and
+    it essentially finds all the bundle image pull specifications in the CSV files of the operator
+    bundle image in question. If this customization is not specified in the config for an
+    organization, the `related_bundles` endpoint won't work for regenerate-bundle requests of that
+    organization. If no organization is specified, IIB will try to find `related_bundles` for all
+    regenerate-bundle requests.
 
   Here is an example that ties this all together:
 
@@ -324,6 +333,8 @@ The custom configuration options for the Celery workers are listed below:
     }
   ```
 
+* `iib_request_related_bundles_dir` - the directory to write the request specific related bundles
+  file. If `None`, per request related bundles files are not created. This defaults to `None`.
 * `iib_request_logs_dir` - the directory to write the request specific log files. If `None`, per
   request log files are not created. This defaults to `None`.
 * `iib_request_logs_format` - the format for the log messages of the request specific log files.
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -68,6 +68,7 @@ services:
       - ./:/src:z
       - ./docker/message_broker/certs:/broker-certs:ro,z
       - request-logs-volume:/var/log/iib/requests:z
+      - request-related-bundles-volume:/var/lib/requests/related_bundles:z
     ports:
       - 8080:8080
     depends_on:
@@ -103,6 +104,7 @@ services:
       - ./docker/registry/certs:/registry-certs:ro,z
       - ./ca-bundle.crt:/host-ca-bundle.crt:ro
       - request-logs-volume:/var/log/iib/requests:z
+      - request-related-bundles-volume:/var/lib/requests/related_bundles:z
     depends_on:
       - rabbitmq
       - registry
@@ -124,4 +126,5 @@ services:
 volumes:
   message-broker-volume:
   request-logs-volume:
+  request-related-bundles-volume:
   worker_container_storage:
--- a/iib/web/api_v1.py
+++ b/iib/web/api_v1.py
@@ -172,7 +172,7 @@ def get_build_logs(request_id):
     request = Request.query.get_or_404(request_id)
     log_file_path = os.path.join(request_log_dir, f'{request_id}.log')
     if not os.path.exists(log_file_path):
-        expired = request.logs_expiration < datetime.utcnow()
+        expired = request.temporary_data_expiration < datetime.utcnow()
         if expired:
             raise Gone(f'The logs for the build request {request_id} no longer exist')
         finalized = request.state.state_name in RequestStateMapping.get_final_states()
@@ -183,6 +183,51 @@ def get_build_logs(request_id):
 
     with open(log_file_path) as f:
         return flask.Response(f.read(), mimetype='text/plain')
+
+
+@api_v1.route('/builds/<int:request_id>/related_bundles')
+def get_related_bundles(request_id):
+    """
+    Retrieve the related bundle images from the bundle CSV for a regenerate-bundle request.
+
+    :param int request_id: the request ID that was passed in through the URL.
+    :rtype: flask.Response
+    :raise NotFound: if the request is not found or there are no related bundles for the request
+    :raise Gone: if the related bundles for the build request have been removed due to expiration
+    """
+    request_related_bundles_dir = flask.current_app.config['IIB_REQUEST_RELATED_BUNDLES_DIR']
+    if not request_related_bundles_dir:
+        raise NotFound()
+
+    request = Request.query.get_or_404(request_id)
+    if request.type != RequestTypeMapping.regenerate_bundle.value:
+        raise ValidationError(
+            f'The request {request_id} is of type {request.type_name}. '
+            'This endpoint is only valid for requests of type regenerate-bundle.'
+        )
+
+    finalized = request.state.state_name in RequestStateMapping.get_final_states()
+    if not finalized:
+        raise ValidationError(
+            f'The request {request_id} is not complete yet.'
+            ' related_bundles will be available once the request is complete.'
+        )
+
+    related_bundles_file_path = os.path.join(
+        request_related_bundles_dir, f'{request_id}_related_bundles.json'
+    )
+    if not os.path.exists(related_bundles_file_path):
+        expired = request.temporary_data_expiration < datetime.utcnow()
+        if expired:
+            raise Gone(f'The related_bundles for the build request {request_id} no longer exist')
+        raise IIBError(
+            'IIB is done processing the request and cannot find related_bundles. Please make '
+            f'sure the iib_organizaiton_customizations for organization {request.organization}'
+            ' has related_bundles customization type set'
+        )
+
+    with open(related_bundles_file_path) as f:
+        return flask.Response(f.read(), mimetype='application/json')
 
 
 @api_v1.route('/builds')
--- a/iib/web/config.py
+++ b/iib/web/config.py
@@ -21,8 +21,9 @@ class Config(object):
     IIB_MESSAGING_DURABLE = True
     IIB_MESSAGING_KEY = '/etc/iib/messaging.key'
     IIB_MESSAGING_TIMEOUT = 30
+    IIB_REQUEST_DATA_DAYS_TO_LIVE = 3
     IIB_REQUEST_LOGS_DIR = None
-    IIB_REQUEST_LOGS_DAYS_TO_LIVE = 3
+    IIB_REQUEST_RELATED_BUNDLES_DIR = None
     IIB_USER_TO_QUEUE = {}
     IIB_WORKER_USERNAMES = []
     SQLALCHEMY_TRACK_MODIFICATIONS = False
@@ -43,6 +44,7 @@ class DevelopmentConfig(Config):
     IIB_MESSAGING_CA = '/etc/iib/messaging-ca.crt'
     IIB_MESSAGING_URLS = ['amqps://message-broker:5671']
     IIB_REQUEST_LOGS_DIR = '/var/log/iib/requests'
+    IIB_REQUEST_RELATED_BUNDLES_DIR = '/var/lib/requests/related_bundles'
     SQLALCHEMY_DATABASE_URI = 'postgresql+psycopg2://iib:iib@db:5432/iib'
     LOGIN_DISABLED = True
 
--- a/iib/web/models.py
+++ b/iib/web/models.py
@@ -424,7 +424,7 @@ class Request(db.Model):
             latest_state = states[0]
             if current_app.config['IIB_REQUEST_LOGS_DIR']:
                 rv['logs'] = {
-                    'expiration': self.logs_expiration.isoformat() + 'Z',
+                    'expiration': self.temporary_data_expiration.isoformat() + 'Z',
                     'url': url_for('.get_build_logs', request_id=self.id, _external=True),
                 }
         rv.update(latest_state or _state_to_json(self.state))
@@ -451,15 +451,15 @@ class Request(db.Model):
         return RequestTypeMapping.pretty(self.type)
 
     @property
-    def logs_expiration(self):
+    def temporary_data_expiration(self):
         """
-        Return the timestamp of when logs are considered expired.
+        Return the timestamp of when logs and related_bundles are considered expired.
 
-        :return: logs expiration timestamp
+        :return: temporary data expiration timestamp
         :rtype: str
         """
-        logs_lifetime = timedelta(days=current_app.config['IIB_REQUEST_LOGS_DAYS_TO_LIVE'])
-        return self.state.updated + logs_lifetime
+        data_lifetime = timedelta(days=current_app.config['IIB_REQUEST_DATA_DAYS_TO_LIVE'])
+        return self.state.updated + data_lifetime
 
 
 class Batch(db.Model):
@@ -1131,6 +1131,11 @@ class RequestRegenerateBundle(Request):
             self.from_bundle_image_resolved, 'pull_specification', None
         )
         rv['organization'] = self.organization
+        if current_app.config['IIB_REQUEST_RELATED_BUNDLES_DIR']:
+            rv['related_bundles'] = {
+                'expiration': self.temporary_data_expiration.isoformat() + 'Z',
+                'url': url_for('.get_related_bundles', request_id=self.id, _external=True),
+            }
 
         return rv
 
--- a/iib/workers/config.py
+++ b/iib/workers/config.py
@@ -30,6 +30,7 @@ class Config(object):
     )
     iib_request_logs_level = 'DEBUG'
     iib_required_labels = {}
+    iib_request_related_bundles_dir = None
     # Configuration for dogpile.cache
     # Disabled by default (by using 'dogpile.cache.null').
     # To enable caching set 'dogpile.cache.memcached' as backend.
@@ -72,6 +73,7 @@ class DevelopmentConfig(Config):
     iib_log_level = 'DEBUG'
     iib_organization_customizations = {
         'company-marketplace': [
+            {'type': 'related_bundles'},
             {
                 'type': 'csv_annotations',
                 'annotations': {
@@ -112,6 +114,7 @@ class DevelopmentConfig(Config):
                     'quay.io': 'registry.koji.company.com',
                 },
             },
+            {'type': 'related_bundles'},
             {'type': 'image_name_from_labels', 'template': '{name}-{version}'},
             {'type': 'enclose_repo', 'enclosure_glue': '----', 'namespace': "company-pending"},
             {
@@ -122,6 +125,7 @@ class DevelopmentConfig(Config):
     }
     iib_registry = 'registry:8443'
     iib_request_logs_dir = '/var/log/iib/requests'
+    iib_request_related_bundles_dir = '/var/lib/requests/related_bundles'
     iib_dogpile_backend = 'dogpile.cache.memcached'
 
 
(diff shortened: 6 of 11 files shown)


docker-compose.yml
```diff
@@ -67,7 +67,8 @@ services:
     volumes:
       - ./:/src:z
       - ./docker/message_broker/certs:/broker-certs:ro,z
-      - request-logs-volume:/var/log/iib/requests:z
+      - request-logs-volume:/var/log/iib/requests/request_logs:z
+      - request-related-bundles-volume:/var/log/iib/requests/related_bundles:z
```
lcarva, line 71: Maybe the related bundles information shouldn't go under `/var/log`? Is `/var/lib` a better path?

iib/web/api_v1.py
```diff
@@ -185,6 +185,48 @@ def get_build_logs(request_id):
...
 
 
+@api_v1.route('/builds/<int:request_id>/related_bundle_images')
+def get_related_bundle_images(request_id):
+    """
+    Retrieve the related bundle images from the bundle CSV for a regenerate-bundle request.
+
+    :param int request_id: the request ID that was passed in through the URL.
+    :rtype: flask.Response
+    :raise NotFound: if the request is not found or there are no logs for the request
```
lcarva, line 195: This line and the line below reference "logs" when they should reference related bundle images.

iib/web/api_v1.py
```diff
@@ -185,6 +185,48 @@ def get_build_logs(request_id):
...
+    :raise Gone: if the logs for the build request have been removed due to expiration
+    """
+    request_related_bundles_dir = flask.current_app.config['IIB_REQUEST_RELATED_BUNDLES_DIR']
+    if not request_related_bundles_dir:
+        raise NotFound()
+
+    request = Request.query.get_or_404(request_id)
+    if request.type_name != RequestTypeMapping.pretty(3):
```
lcarva, line 203: Consider:
```python
if request.type == RequestTypeMapping.regenerate_bundle:
  ...
```

iib/web/api_v1.py
```diff
@@ -185,6 +185,48 @@ def get_build_logs(request_id):
...
+        return flask.Response(
+            f'The request {request_id} is of type {request.type_name}. '
+            'This endpoint is only valid for requests of type regenerate-bundle.',
+            mimetype='application/json',
```
lcarva, line 207: This is not JSON. Maybe just raise a `ValidationError`?

iib/web/api_v1.py
```diff
@@ -185,6 +185,48 @@ def get_build_logs(request_id):
...
+        )
+
+    related_bundles_file_path = os.path.join(
+        request_related_bundles_dir, f'{request_id}_related_bundles.json'
+    )
+    if not os.path.exists(related_bundles_file_path):
+        expired = request.temporary_data_expiration < datetime.utcnow()
+        if expired:
+            raise Gone(f'The related_bundles for the build request {request_id} no longer exist')
```
lcarva, line 216: Let's be consistent in the wording that is exposed to the user. The API endpoint is `/related_bundle_images` but this error refers to it as `related_bundles`. Let's pick one and stick with it. My vote goes to the shorter form.

iib/web/api_v1.py
```diff
@@ -185,6 +185,48 @@ def get_build_logs(request_id):
...
+        finalized = request.state.state_name in RequestStateMapping.get_final_states()
+        if finalized:
+            raise NotFound()
+        # The request may not have been completed yet.
+        return flask.Response(
```
lcarva, line 221: To avoid a potential partial file, let's only load the related_bundles JSON if the request has completed. Wdyt? If you agree, we should move this check further up:

1. Check if request exists. Raise 404 if not
2. Check if request is of type regenerate. Raise 400 if not (raising the `ValidationError` exception will cause this)
3. Check if request is in complete state. Raise 400 if not.
4. Check if JSON file exists, if not check expiration date, if not expired it's a server error, Otherwise, it's `Gone`.
5. Finally! Return the data.

I think the algorithm has to be slightly different than the logic for returning log files. We only ever expect this endpoint to be accessed and return data once a regenerate bundle request finishes successfully, right?

iib/web/api_v1.py
```diff
@@ -185,6 +185,48 @@ def get_build_logs(request_id):
...
+            'The related_bundles will be available here once the request is complete',
+            mimetype='application/json',
```
lcarva, line 223: This is not JSON.

iib/web/api_v1.py
```diff
@@ -501,7 +543,7 @@ def regenerate_bundle():
     error_callback = failed_request_callback.s(request.id)
     try:
         handle_regenerate_bundle_request.apply_async(
-            args=args, link_error=error_callback, argsrepr=repr(safe_args), queue=_get_user_queue(),
+            args=args, link_error=error_callback, argsrepr=repr(safe_args), queue=_get_user_queue()
```
lcarva, line 546: Unnecessary change?

iib/web/models.py
```diff
@@ -1003,7 +1003,7 @@ def from_json(cls, kwargs, batch=None):
             or len(operators) == 0
             or any(not item or not isinstance(item, str) for item in operators)
         ):
-            raise ValidationError(f'"operators" should be a non-empty array of strings')
```
lcarva, line 1006: Separate commit?

iib/workers/config.py
```diff
@@ -187,14 +202,18 @@ def validate_celery_config(conf, **kwargs):
...
-                f'iib_request_logs_dir, {iib_request_logs_dir}, must exist and be a directory'
-            )
-        if not os.access(iib_request_logs_dir, os.W_OK):
-            raise ConfigError(f'iib_request_logs_dir, {iib_request_logs_dir}, is not writable!')
+    for directory in ('iib_request_logs_dir', 'iib_request_related_bundles_dir'):
+        iib_request_temp_data_dir = conf.get(directory)
+        if iib_request_temp_data_dir:
+            if not os.path.isdir(iib_request_temp_data_dir):
+                raise ConfigError(
+                    f'iib_request_temp_data_dir, {iib_request_temp_data_dir}, '
```
lcarva, line 210: `iib_request_temp_data_dir` is only defined within this function. Should the error message refer to the variable `directory` instead? It would make the error easier to diagnose.

iib/workers/tasks/build_regenerate_bundle.py
```diff
@@ -252,6 +262,7 @@ def _adjust_operator_bundle(manifests_path, metadata_path, organization=None, pi
                     operator_manifest, pinned_by_iib, perform_sanity_checks=False
                 )
                 _apply_registry_replacements(bundle_metadata, registry_replacements)
+                _get_related_bundle_images(bundle_metadata, request_id)
```
lcarva, line 265: If this is accessing the related bundle images, it won't always work. I believe in the redhat-marketplace case, the new registries used do not yet contain the images.

iib/workers/tasks/build_regenerate_bundle.py
```diff
@@ -529,3 +540,31 @@ def _apply_repo_enclosure(bundle_metadata, org_enclose_repo_namespace, org_enclo
...
+
+    :param dict bundle_metadata: the dictionary of CSV's and relatedImages pull specifications
+    :param int request_id: the ID of the IIB build request
+    """
+    worker_config = get_worker_config()
+    related_bundles_dir = worker_config['iib_request_related_bundles_dir']
+    related_bundles_file = os.path.join(related_bundles_dir, f'{request_id}_related_bundles.json')
+
+    if os.path.isfile(related_bundles_file):
+        log.debug('Related bundles already exist for this request. Skipping...')
```
lcarva, line 557: Why not overwrite it? It seems like if we're calling this method again, something has changed and we want to regenerate the list of related images.

iib/workers/tasks/build_regenerate_bundle.py
```diff
@@ -529,3 +540,31 @@ def _apply_repo_enclosure(bundle_metadata, org_enclose_repo_namespace, org_enclo
...
+        return
+
+    related_bundle_images = []
+    for related_pullspec_obj in bundle_metadata['found_pullspecs']:
+        related_pullspec = related_pullspec_obj.to_str()
+        if yaml.load(
+            get_image_label(related_pullspec, 'com.redhat.delivery.operator.bundle') or 'false'
```
lcarva, line 564: I'm confused about this. Why are we only providing the related images that are themselves bundles? Don't we need to provide a list of all relatedImages a bundle specified?

iib/workers/config.py
```diff
@@ -187,14 +202,18 @@ def validate_celery_config(conf, **kwargs):
...
-        if not os.path.isdir(iib_request_logs_dir):
-            raise ConfigError(
-                f'iib_request_logs_dir, {iib_request_logs_dir}, must exist and be a directory'
-            )
-        if not os.access(iib_request_logs_dir, os.W_OK):
-            raise ConfigError(f'iib_request_logs_dir, {iib_request_logs_dir}, is not writable!')
+    for directory in ('iib_request_logs_dir', 'iib_request_related_bundles_dir'):
+        iib_request_temp_data_dir = conf.get(directory)
+        if iib_request_temp_data_dir:
+            if not os.path.isdir(iib_request_temp_data_dir):
```
querti, line 208: Through what mechanism are these directories created? Isn't it better to create them instead of raising an error?

iib/web/api_v1.py
yashvardhannanavati: Python black complains if I don't do this

iib/workers/tasks/build_regenerate_bundle.py
yashvardhannanavati: Correct, hence the decision to not overwrite the file if it already exists.

iib/workers/tasks/build_regenerate_bundle.py
yashvardhannanavati: Because it won't always work. As you mentioned above, the managed services use case won't have the images present in the namespace yet.

iib/workers/tasks/build_regenerate_bundle.py
yashvardhannanavati: Nope, we only need relatedImages which are themselves bundles. This is for meta-operators which install other operators.

lcarva: Thanks for the clarification. We need to make this much clearer in the code and docs.

iib/web/api_v1.py
lcarva: Maybe a separate commit then?

iib/workers/tasks/build_regenerate_bundle.py
lcarva: Given that when this can be done depends on the customization for the organization, can we make this configurable? We could introduce a new type of customization that all it does is export the related bundle images. If not provided, it's always done at the end. Wdyt?

yashvardhannanavati: That sounds like a good idea to me

README.md
```diff
@@ -184,10 +184,13 @@ The custom configuration options for the REST API are listed below:
 * `IIB_LOG_LEVEL` - the Python log level of the REST API (Flask). This defaults to `INFO`.
 * `IIB_MAX_PER_PAGE` - the maximum number of build requests that can be shown on a single page.
   This defaults to `20`.
+* `IIB_REQUEST_DATA_DAYS_TO_LIVE` - the amount of days after which per request temmporary data is
+  considered to be expired and may be removed. This defaults to `3`.
```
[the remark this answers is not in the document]
yashvardhannanavati: This needs to be changed in the config

yashvardhannanavati:
@lcarva @querti I have updated the PR based on your suggestions. Could you please take a look? related_bundles is now a customization. The API endpoint has also been modified based on suggestions from Luiz. Could you PTAL?

@pbortlov I will update the default expiration time in a separate commit/PR

Thank you

iib/web/models.py
```diff
@@ -1131,6 +1131,11 @@ def to_json(self, verbose=True):
             self.from_bundle_image_resolved, 'pull_specification', None
         )
         rv['organization'] = self.organization
+        if current_app.config['IIB_REQUEST_RELATED_BUNDLES_DIR']:
+            rv['related_bundles'] = {
+                'expiration': self.temporary_data_expiration.isoformat() + 'Z',
+                'url': url_for('.get_related_bundle_images', request_id=self.id, _external=True),
```
pbortlov, line 1137: Can you add a comment here on how the URL will look like?

iib/workers/config.py
```diff
@@ -132,6 +136,7 @@ class TestingConfig(DevelopmentConfig):
     iib_greenwave_url = 'some_url'
     iib_omps_url = 'some_url'
     iib_request_logs_dir = None
+    iib_request_related_bundles_dir = None
```
pbortlov, line 139: Why TestingConfig rewrites `iib_request_related_bundles_dir` from DevelopmentConfig?

iib/workers/tasks/build_regenerate_bundle.py
pbortlov: Could we maybe comment the code?

tests/test_web/test_api_v1.py
```diff
@@ -1924,3 +1937,96 @@ def test_create_empty_index_not_allowed_params(mock_smfsc, data, error_msg, clie
...
+    related_bundles_content,
+    expired,
+    finalized,
+    expected,
+):
+    minimal_request_regenerate_bundle.add_state('in_progress', 'Starting things up!')
+    db.session.commit()
+
+    client.application.config['IIB_REQUEST_RELATED_BUNDLES_DIR'] = str(tmpdir)
+    if expired:
```
pbortlov, line 2003: Could we create the test for expired only?

tests/test_web/test_api_v1.py
```diff
@@ -1924,3 +1937,96 @@ def test_create_empty_index_not_allowed_params(mock_smfsc, data, error_msg, clie
...
+        client.application.config['IIB_REQUEST_DATA_DAYS_TO_LIVE'] = -1
+    if finalized:
```
pbortlov, line 2005: Could we create a test for finalized only?

tests/test_web/test_api_v1.py
```diff
@@ -1924,3 +1937,96 @@ def test_create_empty_index_not_allowed_params(mock_smfsc, data, error_msg, clie
...
+        minimal_request_regenerate_bundle.add_state('complete', 'The request is complete')
+        db.session.commit()
+    request_id = minimal_request_regenerate_bundle.id
+    related_bundles_file = tmpdir.join(f'{request_id}_related_bundles.json')
+    if related_bundles_content is not None:
```
pbortlov, line 2010: Could we have a new test? I don't see a reason to use if statements in tests.

iib/web/api_v1.py
```diff
@@ -185,6 +185,49 @@ def get_build_logs(request_id):
...
+        raise ValidationError(
+            f'The request {request_id} is not complete yet.'
+            ' related_bundles will be available once the request is complete.'
+        )
+
+    related_bundles_file_path = os.path.join(
+        request_related_bundles_dir, f'{request_id}_related_bundles.json'
+    )
+    if not os.path.exists(related_bundles_file_path):
+        expired = request.temporary_data_expiration < datetime.utcnow()
```
lcarva, line 220: Nitpick: consider adding a method to compute this: `Request.is_expired()`

(Not exactly sure why we didn't do that in the first place, but now that there are two places computing this, a method might prevent a future issue.)

iib/web/api_v1.py
```diff
@@ -185,6 +185,49 @@ def get_build_logs(request_id):
...
+        if expired:
+            raise Gone(f'The related_bundles for the build request {request_id} no longer exist')
+        finalized = request.state.state_name in RequestStateMapping.get_final_states()
```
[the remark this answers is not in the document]
lcarva: Actually, this line and the if-block below are not necessary. If the request is not finalized, it should've raised a `ValidationError` before getting here.


If the json file doesn't exist, the request is not expired, and the request is finalized, we should raise a server error IMO. This is completely unexpected behavior.

Consider:

```python
if not os.path.exists(related_bundles_file_path):
  expired = request.temporary_data_expiration < datetime.utcnow()
  if expired:
    raise Gone(f'The related_bundles for the build request {request_id} no longer exist')
  raise IIBError('something is up!') # better error message :) and verify this actually raises 5xx error.

File: iib/web/config.py [hunk 23]
```diff
@@ -43,6 +44,7 @@ class DevelopmentConfig(Config):
     IIB_MESSAGING_CA = '/etc/iib/messaging-ca.crt'
     IIB_MESSAGING_URLS = ['amqps://message-broker:5671']
     IIB_REQUEST_LOGS_DIR = '/var/log/iib/requests'
+    IIB_REQUEST_RELATED_BUNDLES_DIR = '/var/lib/requests/related_bundles'
```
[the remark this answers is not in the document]
lcarva: It inherits from `Config` so it should be all set.

iib/web/models.py
```diff
@@ -451,15 +451,15 @@ def type_name(self):
...
-        Return the timestamp of when logs are considered expired.
+        Return the timestamp of when logs and related_bundles are considered expired.
 
-        :return: logs expiration timestamp
+        :return: temporary data expiration timestamp
         :rtype: str
         """
-        logs_lifetime = timedelta(days=current_app.config['IIB_REQUEST_LOGS_DAYS_TO_LIVE'])
-        return self.state.updated + logs_lifetime
+        data_lifetime = timedelta(days=current_app.config['IIB_REQUEST_DATA_DAYS_TO_LIVE'])
```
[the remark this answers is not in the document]
lcarva: It should always be because the top-most config class, `Config`, defines it.

iib/web/models.py
```diff
@@ -1131,6 +1131,11 @@ def to_json(self, verbose=True):
             self.from_bundle_image_resolved, 'pull_specification', None
         )
         rv['organization'] = self.organization
+        if current_app.config['IIB_REQUEST_RELATED_BUNDLES_DIR']:
```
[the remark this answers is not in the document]
lcarva: Any sort of API configuration validation should be done in `iib/web/app.py:validate_api_config`

iib/web/models.py
lcarva, line 1137: I think this should be `.get_related_bundles` and the function in `api_v1.py` should be renamed to `get_related_bundles` as well. WDYT?

iib/workers/config.py
lcarva: I think the approach we have taken is that this must be pre-configured. In our case this is done via ansible. I like the current approach because it forces us to ensure these directories are created with the proper access.

iib/workers/tasks/build_regenerate_bundle.py
```diff
@@ -529,3 +540,31 @@ def _apply_repo_enclosure(bundle_metadata, org_enclose_repo_namespace, org_enclo
         replacement_pullspecs[pullspec] = new_pullspec
 
     _replace_csv_pullspecs(bundle_metadata, replacement_pullspecs)
+
+
+def _get_related_bundle_images(bundle_metadata, request_id):
```
lcarva, line 545: This shouldn't be called `get` if it doesn't return anything. Given what it does, this should probably be called `_write_related_bundles_file`

iib/workers/tasks/build_regenerate_bundle.py
lcarva: Given the new approach, should this overwrite now?

iib/workers/config.py
lcarva: This still needs to be addressed.

iib/workers/tasks/build_regenerate_bundle.py
lcarva: Thanks for taking this suggestion into account. I think the code looks much cleaner and contains less guess work.

iib/workers/tasks/build_regenerate_bundle.py
```diff
@@ -216,6 +220,7 @@ def _adjust_operator_bundle(manifests_path, metadata_path, organization=None, pi
     organization_customizations = conf['iib_organization_customizations'].get(organization, [])
```
lcarva, line 220: If the defined organization_customization does not specify `related_bundles`, then the related bundles file will not be generated, right? I think that's ok, but we need to document this. In the README.md?

iib/web/api_v1.py
```diff
@@ -185,6 +185,48 @@ def get_build_logs(request_id):
...
+        if finalized:
```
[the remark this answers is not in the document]
yashvardhannanavati: It was when the request is complete and we cannot find the related_bundles file for it. It has been changed now :)

iib/web/models.py
yashvardhannanavati: What do you mean? 
Check this out: https://flask.palletsprojects.com/en/2.0.x/api/#flask.url_for

yashvardhannanavati: Yes, indeed

iib/workers/config.py
yashvardhannanavati: DevelopmentConfig is used by the dev env and so we need the value there. For TestingConfig, since it inherits the values, they will be set. But we actually use a separate `tmpdir` as the location for those files for tests. We do not want to unnecessarily populate that location by the tests that might be written wrong. Hence the overwriting.

iib/workers/config.py
yashvardhannanavati: oops! apologies

tests/test_web/test_api_v1.py
yashvardhannanavati: Can I push back on this please. I actually think having a single test with multiple combinations of `finalized` and `expired` is more useful than having individual test cases.  It also helps avoids unnecessary LOC.

iib/workers/tasks/build_regenerate_bundle.py
yashvardhannanavati: The function definition mentions that it finds bundle images in the CSVs. Do we still need a comment here?

iib/workers/tasks/build_regenerate_bundle.py
```diff
@@ -529,3 +540,31 @@ def _apply_repo_enclosure(bundle_metadata, org_enclose_repo_namespace, org_enclo
...
+    related_bundles_file = os.path.join(related_bundles_dir, f'{request_id}_related_bundles.json')
+
+    if os.path.isfile(related_bundles_file):
+        log.debug('Related bundles already exist for this request. Skipping...')
+        return
+
+    related_bundle_images = []
+    for related_pullspec_obj in bundle_metadata['found_pullspecs']:
+        related_pullspec = related_pullspec_obj.to_str()
+        if yaml.load(
```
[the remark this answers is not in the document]
yashvardhannanavati: They look correct to me. Am I missing something?

iib/web/api_v1.py
```diff
@@ -185,6 +185,49 @@ def get_build_logs(request_id):
...
+
+    request = Request.query.get_or_404(request_id)
+    if request.type != RequestTypeMapping.regenerate_bundle.value:
+        raise ValidationError(
+            f'The request {request_id} is of type {request.type_name}. '
+            'This endpoint is only valid for requests of type regenerate-bundle.'
+        )
+
+    finalized = request.state.state_name in RequestStateMapping.get_final_states()
+    if not finalized:
```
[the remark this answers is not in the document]
yashvardhannanavati: Fixed it according to Luiz's suggesiton

iib/workers/config.py
lcarva: Sounds good. It removes ambiguity in the tests.

iib/workers/tasks/build_regenerate_bundle.py
lcarva: @yashvardhannanavati, I don't think it's _needed_ but I think we have a better chance of avoiding future confusion if we add something like this:
```python
# Important: not all related images are bundle images!
if yaml.load(...
```
But this is borderline paranoia so use your best judgement to accept or ignore this suggestion.

lcarva (dismissed):

midnightercz (approved):

yashvardhannanavati:
@midnightercz @lcarva I improved an error message and updated the commit. That wiped off your approvals. Could you please re-approve? thanks!

yashvardhannanavati:
fwiw, I have verified this as working on my dev-env :)

lcarva (approved):

midnightercz (approved):

PR #271, opened 2021-07-20
Use pinned runtime requirements in tests

lcarva:


requirements-test.txt
```diff
@@ -417,18 +465,25 @@ toml==0.10.1 \
...
     --hash=sha256:6dc65cf9091cf750012f56f2cad759fa9e879f511b5ff8685e456b4e3bf90d16
-    # via flask
-zipp==3.1.0 \
-    --hash=sha256:aa36550ff0c0b7ef7fa639055d797116ee891440eac1a56f378e2d3179e0320b \
-    --hash=sha256:c599e4d75c98f6798c509911d08a22e6c021d074469042177c8c86fb92eefd96
-    # via importlib-metadata
+    # via
+    #   -r requirements.txt
+    #   flask
+zipp==2.2.0 \
```
pbortlov, line 484: Here is a lower version than before

requirements-test.txt
```diff
@@ -1,32 +1,40 @@
 #
-# This file is autogenerated by pip-compile
+# This file is autogenerated by pip-compile with python 3.6
```
pbortlov, line 2: Do we really want to do it with python3.6?

lcarva: Yes, ma'am. IIB runs on python 3.6 in our deployments. We have also added the [.python-version](https://github.com/release-engineering/iib/blob/master/.python-version) file so dependabot uses python 3.6 as well.

A new version of pip-tools must be writing the python version now. Thus the difference.

requirements-test.txt
lcarva: Exactly! This is why this change is so important. We were running the tests with a different version than what was used in our deployments.

After this PR, I can have a follow up to update the `zipp` dependency.

requirements-test.txt
pbortlov: ok, thanks

pbortlov (approved):

PR #272, opened 2021-07-21
Release v4.2.0

pbortlov:
Release

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,16 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+## 4.2.0
+- Add create-empty-index functionality
+- Replace custom retry with retry package and setting backoff interval for skopeo_inspect
+- Set --arch when building images
+- Bump pytest-cov from 2.12.0 to 2.12.1
+- Remove retry delay during tests
+- Request and urllib3 update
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -6,6 +6,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+## 5.0.0
```
lcarva, line 9: Why the major version update?

pbortlov: new endpoint was added: Create-empty-index

lcarva: Oh ok. I think in the past we have bumped the major version to indicate a breaking change. Adding a new endpoint doesn't really break backwards compatibility though.

yashvardhannanavati: @lcarva It was my bad. While filing the issue for the release, I assumed the change for `related_bundle_images`  (#270) would be included and so I filed for 5.0.0 because that could potentially break some code bases that expect a specific response and not extra params (like the problems we faced with Pub CI breaking when we add a new param to the response)

@pbortlov I believe a 4.2.0 should suffice?

pbortlov: thanks

midnightercz (dismissed):

midnightercz (approved):

PR #275, opened 2021-08-06
Disable OMPS

lipoja:
The cluster OMPS is on, is scheduled to be decommissioned on 16th Aug. The plan is to stop supporting shipping operators to the legacy app-registry via IIB.

- legacy.py was removed
- tests were update
- API parameters are unchanged
- warning is logged  when request with cnr_token and organization is processed

refers to CLOUDDST-6578


iib/workers/tasks/build.py
```diff
@@ -868,13 +865,10 @@ def handle_add_request(
...
 
-    log.info('Checking if interacting with the legacy app registry is required')
-    legacy_support_packages = get_legacy_support_packages(
-        resolved_bundles, request_id, prebuild_info['ocp_version'], force_backport=force_backport
-    )
-    if legacy_support_packages:
-        validate_legacy_params_and_config(
-            legacy_support_packages, resolved_bundles, cnr_token, organization
+    if (cnr_token and organization) or force_backport:
+        log.warning(
```
yashvardhannanavati, line 869: Can change this warning to `Legacy support is disabled in IIB. cnr_token, organization and force_backport will be ignored. `

yashvardhannanavati: Since Jan is on PTO, I'll submit this change in a separate PR

yashvardhannanavati:
The documentation should also show those params as deprecated.

yashvardhannanavati (approved):

PR #286, opened 2021-08-31
Use buildah instead of manifest-tool to create manifest lists

yashvardhannanavati:
Relates to CLOUDDST-8606

### DIFF
--- a/iib/workers/tasks/build.py
+++ b/iib/workers/tasks/build.py
@@ -8,7 +8,6 @@ import stat
 import subprocess
 import time
 import tempfile
-import textwrap
 
 from operator_manifest.operator import ImageName
 from retry import retry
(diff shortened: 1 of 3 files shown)


iib/workers/tasks/build.py
```diff
@@ -116,45 +116,36 @@ def _create_and_push_manifest_list(request_id, arches):
...
-        manifest_yaml = os.path.abspath(os.path.join(temp_dir, 'manifest.yaml'))
-        with open(manifest_yaml, 'w+') as manifest_yaml_f:
-            manifest_yaml_f.write(
-                textwrap.dedent(
-                    f'''\
-                    image: {output_pull_spec}
-                    manifests:
-                    '''
-                )
+        for arch in sorted_arches:
```
pbortlov, line 124: would it be possible to use line 122 here? if you don't use sorted_aches anywhere else?

iib/workers/tasks/build.py
```diff
@@ -116,45 +116,36 @@ def _create_and_push_manifest_list(request_id, arches):
...
-                manifest_yaml_f.write(
-                    textwrap.dedent(
-                        f'''\
-                        - image: {arch_pull_spec}
-                          platform:
-                            architecture: {arch}
-                            os: linux
-                        '''
-                    )
+            if arch == sorted_arches[0]:
```
pbortlov, line 131: Could you add a comment here?

iib/workers/tasks/build.py
yashvardhannanavati: But I am using it on Line 130

iib/workers/tasks/build.py
yashvardhannanavati: Done

iib/workers/tasks/build.py
```diff
@@ -116,48 +115,36 @@ def _create_and_push_manifest_list(request_id, arches):
...
+    for arch in sorted_arches:
+        arch_pull_spec = _get_external_arch_pull_spec(request_id, arch, include_transport=True)
+        log.debug(
+            'Adding the manifest %s to the manifest list %s', arch_pull_spec, output_pull_spec
+        )
+        # Create the manifest list with the first arch and then add other arches to it using the
+        # buildah-manifest add command.
+        if arch == sorted_arches[0]:
+            run_cmd(
+                buildah_manifest_cmd + ['create', output_pull_spec, arch_pull_spec],
```
lcarva, line 131: You can call `create` with just the pull spec of the manifest list. Consider this pattern:
```python
run_cmd(
    buildah_manifest_cmd + ['create', output_pull_spec],
    exc_msg=f'Failed to create the manifest list locally: {output_pull_spec}',
)
...
for arch in sorted_arches:
...
  run_cmd(
       buildah_manifest_cmd + ['add', output_pull_spec, arch_pull_spec],
       exc_msg=(
          f'Failed to add {arch_pull_spec} to the' f' local manifest list: {output_pull_spec}'
       ),
  )
...
```

It has the benefit of not treating the first arch a special case.

iib/workers/tasks/build.py
```diff
@@ -116,48 +115,36 @@ def _create_and_push_manifest_list(request_id, arches):
...
+                exc_msg=f'Failed to create the manifest list locally: {output_pull_spec}',
             )
-
+            continue
         run_cmd(
-            ['manifest-tool', 'push', 'from-spec', manifest_yaml],
-            exc_msg=f'Failed to push the manifest list to {output_pull_spec}',
+            buildah_manifest_cmd + ['add', output_pull_spec, arch_pull_spec],
```
lcarva, line 136: Do we need to use `--os` or `--arch` here?

lcarva (commented):
Let's also remove all references to `manifest-tool`, e.g. in the Dockerfile.

iib/workers/tasks/build.py
yashvardhannanavati: I don't think we need `--os` here and we build the image using `--arch` so the metadata on the image is already set correctly. I could add it if you think it's safe to do

iib/workers/tasks/build.py
yashvardhannanavati: Done

iib/workers/tasks/build.py
lcarva: If it generates the correct manifest, then there's no need for it. Have you inspected the raw manifest list created in the registry?

yashvardhannanavati: yes, I have inspected it and it looks correct

lcarva: Cool!

lcarva:
> Let's also remove all references to `manifest-tool`, e.g. in the Dockerfile.

@yashvardhannanavati, did you see this ^?

lcarva (dismissed):

yashvardhannanavati:
> > Let's also remove all references to `manifest-tool`, e.g. in the Dockerfile.
> 
> @yashvardhannanavati, did you see this ^?

I hadn't seen it. I just added a commit removing it from docker-compose.

lcarva (approved):

lipoja (approved):

PR #290, opened 2021-09-10
Fix merge-index-image bug that created the same manifest list twice

yashvardhannanavati:


iib/workers/tasks/build_merge_index_image.py
```diff
@@ -273,7 +273,9 @@ def handle_merge_request(
             file_mode=(stat.S_IRUSR | stat.S_IWUSR | stat.S_IRGRP | stat.S_IWGRP),
         )
 
-    output_pull_spec = _create_and_push_manifest_list(request_id, prebuild_info['arches'])
+    output_pull_spec = _create_and_push_manifest_list(
+        request_id, prebuild_info['arches'], skip_manifest_list_creation=True
```
lcarva, line 277: Wouldn't it be better to delete the previous manifest list before making this call? I worry that the current approach may lead to odd behavior in the future where the leftover manifest list interferes with the new one.

yashvardhannanavati: @lcarva I was looking at the code and realized we do not need to create the manifest list twice. Could you ptal to make sure I'm not missing anything.

lcarva (approved):

iib/workers/tasks/build_merge_index_image.py
lcarva: Even better!

PR #293, opened 2021-09-14
Cleanup local manifest lists before creating a new one

yashvardhannanavati:
Refers to CLOUDDST-9111

### DIFF
--- a/iib/workers/tasks/build.py
+++ b/iib/workers/tasks/build.py
@@ -115,6 +115,10 @@ def _create_and_push_manifest_list(request_id, arches):
     :rtype: str
     :raises IIBError: if creating or pushing the manifest list fails
     """
+    run_cmd(
+        ['buildah', 'rmi', '--all'],
+        exc_msg='Failed to cleanup all local images',
+    )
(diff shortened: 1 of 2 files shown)


iib/workers/tasks/build.py
```diff
@@ -115,6 +115,10 @@ def _create_and_push_manifest_list(request_id, arches):
     :rtype: str
     :raises IIBError: if creating or pushing the manifest list fails
     """
+    run_cmd(
+        ['buildah', 'rmi', '--all'],
```
lcarva, line 119: It feels unexpected that this method cleans all images known to buildah. Especially because for images, buildah on podman share the same location. Can we just remove the manifest and ignore if it doesn't exist?

zanssa: agreed! based on the description for the method, I expect the behavior to remove only the manifest.

yashvardhannanavati:
Closing this in favor of #299  :)

iib/workers/tasks/build.py
yashvardhannanavati: Resolved it in #299

PR #297, opened 2021-09-22
Set OcpMaxVersion property in merge-index-image request in IIB

yashvardhannanavati:
Refers to CLOUDDST-8769

### DIFF
--- a/iib/workers/tasks/utils.py
+++ b/iib/workers/tasks/utils.py
@@ -8,7 +8,9 @@ import json
 import logging
 import os
 import re
+import sqlite3
 import subprocess
+import time
 
 from retry import retry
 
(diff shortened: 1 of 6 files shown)


iib/workers/tasks/utils.py
```diff
@@ -50,6 +121,130 @@ def get_binary_image_from_config(ocp_version, distribution_scope, binary_image_c
...
+            # check if the failure is due to the presence of deprecated APIs
+            # and absence of the 'olm.maxOpenShiftVersion' property
+            # Note: there is no other error in the sdk that mentions this field
+            for msg in output['outputs']:
+                if 'olm.maxOpenShiftVersion' in msg['message']:
+                    return True
+    return False
+
+
+def serve_index_registry(db_path):
```
pbortlov, line 163: @yashvardhannanavati I like your idea of moving code out of build.py but utils.py is growing extremely. Can we make a new file called opm or something like that?

yashvardhannanavati: A new file simply for these functions only wouldn't make sense IMO. We should file an issue and discuss it with the guild before doing a major refactoring

pbortlov: sounds good :) please ping me in the jira

yashvardhannanavati: Sounds good. In the meanwhile, could you please help review this PR?

lipoja (dismissed):

pbortlov:
@yashvardhannanavati did you rebase with the master? it seems there are problems with `_skope_copy` which is renamed to `skope_copy`.

iib/workers/tasks/utils.py
```diff
@@ -50,6 +121,136 @@ def get_binary_image_from_config(ocp_version, distribution_scope, binary_image_c
...
+    """
+    Get bundle json from grpcurl response.
+
+    :param list bundles: response from grpcurl call to retrieve list of bundles in
+        an index
+    """
+    return [json.loads(bundle) for bundle in re.split(r'(?<=})\n(?={)', bundles)]
+
+
+def _requires_max_ocp_version(bundle):
```
pbortlov, line 134: Could you move this in front of the function ` add_max_ocp_version_property`? It should be defined before the function is used.

iib/workers/tasks/utils.py
```diff
@@ -50,6 +121,136 @@ def get_binary_image_from_config(ocp_version, distribution_scope, binary_image_c
...
+            # check if the failure is due to the presence of deprecated APIs
+            # and absence of the 'olm.maxOpenShiftVersion' property
+            # Note: there is no other error in the sdk that mentions this field
+            for msg in output['outputs']:
+                if 'olm.maxOpenShiftVersion' in msg['message']:
+                    return True
+    return False
+
+
+def serve_index_registry(db_path):
```
pbortlov, line 169: Could you move this in front of the function `add_max_ocp_version_property`? It should be defined before the function is used.

iib/workers/tasks/utils.py
```diff
@@ -50,6 +121,136 @@ def get_binary_image_from_config(ocp_version, distribution_scope, binary_image_c
...
+        except AddressAlreadyInUse:
+            log.info('Port %d is in use, trying another.', port)
+
+    err_msg = f'No free port has been found after {conf.get("iib_grpc_max_port_tries")} attempts.'
+    log.error(err_msg)
+    raise IIBError(err_msg)
+
+
+@retry(exceptions=IIBError, tries=2, logger=log)
+def serve_index_registry_at_port(db_path, port, max_tries, wait_time):
```
pbortlov, line 202: Could you move this above the function `serve_index_registry`? It should be defined before the function is used.

pbortlov (changes requested):

tests/test_workers/test_tasks/test_build.py
```diff
@@ -1064,27 +1094,38 @@ def test_get_present_bundles_grpc_not_initialize(
 @mock.patch('time.sleep')
 @mock.patch('subprocess.Popen')
 @mock.patch('iib.workers.tasks.build.run_cmd')
```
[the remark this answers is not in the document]
yashvardhannanavati: Both of them are required.

iib/workers/tasks/utils.py
yashvardhannanavati: May I ask why? I don't think we need to do that in Python unless I am missing something

pbortlov: I would say best practices and my editor complains a lot :smile:

yashvardhannanavati:
@pbortlov By pulling it locally, and running a merge-index-image request in which, the target_index has more bundles that the source_index. Then in the logs, you can make sure that this piece of code is called when those extra bundles are being added.

iib/workers/tasks/utils.py
```diff
@@ -19,14 +21,83 @@
...
+    port, rpc_proc = serve_index_registry(db_path)
+
+    raw_bundles = run_cmd(
+        ['grpcurl', '-plaintext', f'localhost:{port}', 'api.Registry/ListBundles'],
+        exc_msg='Failed to get bundle data from index image',
+    )
+    rpc_proc.kill()
+
+    # Get bundle json for bundles in the request
+    updated_bundles = list(
```
pbortlov, line 83: Could you update a comment here?
`Get a bundlePath from raw_bundles and check the presence in resolved_bundles to return a list of present bundles`
I am lost here. You convert `raw_bundles` to JSON and then you check presence with resolved_bundles and create a LIST of `updated_bundles` but then you try to work with the `updated_bundles` as a DICTIONARY.

pbortlov: I check filter() here: https://www.programiz.com/python-programming/methods/built-in/filter

pbortlov: I have been testing it.

PR #299, opened 2021-09-29
Cleanup local manifest lists before creating a new one

yashvardhannanavati:
If one of the later steps fails, then in the retry, IIB tries
to recreate an existing manifest list which buildah doesn't allow
This commit resolves that issue by removing the existing local
manifest list before creating a new one.

Refers to CLOUDDST-9014

### DIFF
--- a/iib/workers/tasks/build.py
+++ b/iib/workers/tasks/build.py
@@ -118,6 +118,18 @@ def _create_and_push_manifest_list(request_id, arches):
     """
     buildah_manifest_cmd = ['buildah', 'manifest']
     output_pull_spec = get_rebuilt_image_pull_spec(request_id)
+
+    try:
+        run_cmd(
+            buildah_manifest_cmd + ['rm', output_pull_spec],
+            exc_msg=f'Failed to remove local manifest list. {output_pull_spec} does not exist',
+        )
+    except IIBError as e:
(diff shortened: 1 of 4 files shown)


midnightercz (approved):

iib/workers/tasks/build.py
```diff
@@ -118,6 +118,18 @@ def _create_and_push_manifest_list(request_id, arches):
     """
     buildah_manifest_cmd = ['buildah', 'manifest']
     output_pull_spec = get_rebuilt_image_pull_spec(request_id)
+
+    try:
+        run_cmd(
+            buildah_manifest_cmd + ['rm', output_pull_spec],
```
lcarva, line 124: I think this needs to be `rmi`, not `rm`.

yashvardhannanavati: Nope, the docs say it has to be `rm`. https://github.com/containers/buildah/blob/main/docs/buildah-manifest-rm.md

yashvardhannanavati: Unless, I am missing something

lcarva: Thanks for the correction :)

lcarva (approved):

PR #316, opened 2021-10-27
improve healthcheck endpoint

zanssa:
Ref: CLOUDDST-9948

### DIFF
--- a/iib/web/api_v1.py
+++ b/iib/web/api_v1.py
@@ -7,7 +7,6 @@ import flask
 import kombu
 from flask_login import current_user, login_required
 from sqlalchemy.orm import with_polymorphic
-from sqlalchemy.sql import text
 from sqlalchemy import or_
 from werkzeug.exceptions import Forbidden, Gone, NotFound
 
@@ -32,6 +31,7 @@ from iib.web.models import (
     User,
 )
 from iib.web.utils import pagination_metadata, str_to_bool
+from iib.web.healthcheck import status
 from iib.workers.tasks.build import (
     handle_add_request,
     handle_rm_request,
@@ -314,23 +314,35 @@ def get_builds():
     return flask.jsonify(response)
 
 
-@api_v1.route('/healthcheck')
-def get_healthcheck():
+@api_v1.route('/healthcheck/details', methods=['GET'])
+def get_status():
     """
-    Respond to a health check.
+    Return status of IIB workers and services that IIB depends on.
+
+    :return: json object representing the status of the services.
+    :rtype: flask.Response
+    """
+    return flask.jsonify(status())
+
+
+@api_v1.route('/healthcheck', methods=['GET'])
+def get_status_short():
+    """
+    Return 200 if all workers and services seem to be ok, 503 otherwise.
 
     :rtype: flask.Response
-    :return: json object representing the health of IIB
-    :raises IIBError: if the database connection fails
+    :return: json object representing the status of IIB services.
+    :raise IIBError: if a problem is found while retrieving the status.
     """
-    # Test DB connection
+    retval = {'status': 'Health check OK'}
+    ret_status = 200
     try:
-        db.engine.execute(text('SELECT 1'))
-    except Exception:
-        flask.current_app.logger.exception('DB test failed.')
-        raise IIBError('Database health check failed.')
+        status(short=True)
+    except IIBError as error:
+        retval = {'status': 'Health check failed', 'error': str(error)}
+        ret_status = 503
 
-    return flask.jsonify({'status': 'Health check OK'})
+    return flask.jsonify(retval), ret_status
 
 
 def _get_user_queue(serial=False):
--- a/iib/web/healthcheck.py
+++ b/iib/web/healthcheck.py
@@ -1,0 +1,138 @@
+import time
+import logging
+import sqlalchemy
+import kombu
+
+from iib.exceptions import IIBError
+from iib.web import db
+from iib.workers.config import get_worker_config
+from iib.workers.tasks.celery import app
+
+
+log = logging.getLogger(__name__)
+
+
+def database_connection_check():
+    """
+    Check if the database connection is working.
+
+    :return: tuple (ok: bool, reason: str or None)
+    """
+    session = db.session()
+    try:
+        session.execute('SELECT 1')
+    except sqlalchemy.exc.OperationalError as OperError:
+        log.error('database connection failed, exception: %s', OperError)
+        return False, 'database connection failed'
+    except sqlalchemy.exc.SQLAlchemyError as err:
+        log.error('an error occurred while trying to connect to the DB server: %s', err)
+        return False, 'an error occurred'
+    finally:
+        session.close()
+
+    return True, None
+
+
+def rabbitmq_connection_check(broker_url):
+    """
+    Check if RabbitMQ connection is working.
+
+    :param str broker_url: the AMQP(S) URL to connect to RabbitMQ
+    :return: tuple (ok: bool, reason: str or None)
+    """
+    with kombu.Connection(broker_url) as connection:
+        try:
+            connection.ensure_connection(max_retries=0)
+        except kombu.exceptions.OperationalError:
+            log.exception('broker connection failed')
+            return False, 'broker connection failed'
+        except kombu.exceptions.KombuError as err:
+            log.error('an error occurred while trying to connect to the RabbitMQ server: %s', err)
+            return False, 'an error occurred'
+
+    return True, None
+
+
+def _ping_celery_workers(retries):
+    """
+    Attempt to ping worker, retry on ConnectionError.
+
+    :param int retries: number of retry/retries before returning an empty response
+    :return: dict of {hostname: reply} as returned by ping()
+    """
+    for i in range(retries + 1):
+        if i > 0:
+            time.sleep(0.25 * 2 ** (i - 1))
+        try:
+            replies = app.control.inspect().ping()
+        except ConnectionError:
+            continue
+
+        return replies or {}
+
+    return {}
+
+
+def workers_status(retries=2):
+    """
+    Ping workers, check received replies.
+
+    :param int retries: number of retry/retries on ConnectionError when pinging workers
+    :return: list of workers status information (empty if no replies)
+    """
+    replies = _ping_celery_workers(retries=retries)
+
+    # replies is a dict of {hostname: reply}, convert to a sorted list of [(hostname, reply)]
+    reply_tuples = sorted(replies.items(), key=lambda kv: kv[0])
+    workers = []
+
+    for worker_name, reply in reply_tuples:
+        # the reply is in the following format {'ok': 'pong'}.
+        worker = {'name': worker_name, 'available': 'ok' in reply}
+
+        if 'ok' not in reply:
+            worker['reason'] = reply.get('error', 'unknown reason')
+
+        workers.append(worker)
+
+    return workers
+
+
+def status(short=False, worker_ping_retries=2):
+    """
+    Get status of IIB workers and services that IIB depends on.
+
+    :param bool short: raise an error as soon as any problem is found
+    :param int worker_ping_retries: number of retry/retries on ConnectionError when pinging workers
+    :return: dict with the following keys:
+        "services": list of status info for individual services
+        "workers": list of status info for individual workers
+    :raises IIBError: if short is True and a problem is found
+    """
+    services = []
+
+    def add_status(service_name, ok, reason):
+        if short and not ok:
+            raise IIBError(f'{service_name} unavailable: {reason}')
+        service = {'name': service_name, 'available': ok}
+        if not ok:
+            service['reason'] = reason
+        services.append(service)
+
+    add_status('DATABASE', *database_connection_check())
+    rabbit_ok, rabbit_reason = rabbitmq_connection_check(get_worker_config().broker_url)
+    add_status('RABBITMQ', rabbit_ok, rabbit_reason)
+
+    if rabbit_ok:
+        workers = workers_status(retries=worker_ping_retries)
+    else:
+        workers = []
+
+    any_worker_ok = any(worker['available'] for worker in workers)
+    if short and not any_worker_ok:
+        raise IIBError('no workers are available')
+
+    return {
+        'services': services,
+        'workers': workers,
+    }
--- a/iib/web/static/api_v1.yaml
+++ b/iib/web/static/api_v1.yaml
@@ -210,30 +210,77 @@ paths:
               $ref: '#/components/schemas/RequestUpdate'
       security:
         - Kerberos Authentication: []
-  /healthcheck:
+  /healthcheck/details:
     get:
-      description: Perform a health check of the service
+      description: Get the status of the services IIB depends on
       responses:
         '200':
-          description: The health check OK message
+          description: services status information
           content:
             application/json:
               schema:
                 type: object
                 properties:
-                  status:
-                    type: string
-                    example: Health check OK
-        '500':
-          description: Health check failed
+                  services:
+                    type: array
+                    items:
+                      type: object
+                      properties:
+                        name:
+                          type: string
+                          description: Name of a service
+                          example: database
+                        available:
+                          type: boolean
+                          description: indicator that service is running fine.
+                          example: false
+                        reason:
+                          type: string
+                          description: Reason why service is not running.
+                          example: Database connection failed
+                    required:
+                      - name
+                      - available
+                  workers:
+                    type: array
+                    items:
+                      type: object
+                      properties:
+                        name:
+                          type: string
+                          description: Name of a worker
+                          example: celery@025f84b0154f
+                        available:
+                          type: boolean
+                          description: indicator that worker is available.
+                          example: false
+                        reason:
+                          type: string
+                          description: Reason why worker is not available.
+                          example: worker didn't respond to ping
+                    required:
+                      - name
+                      - available
+            required:
+              - services
+              - workers
+  /healthcheck:
+    get:
+      summary: Get a short IIB service status
+      description: >
+        if services including RabbitMQ, database, and workers
+        are working properly, return 200.
+      responses:
+        '200':
+          description: services and workers are working as expected.
           content:
             application/json:
-              schema:
-                type: object
-                properties:
-                  error:
-                    type: string
-                    example: Database health check failed
+              example: {'status': 'Health check OK'}
+        '503':
+          description: one or more service/s are reporting an issue
+          content:
+            application/json:
+              example: {'status': 'Health check failed', 'error': 'DATABASE unavailable: Database connection failed'}
   '/builds/{id}/logs':
     get:
       description: Return the logs for a specific build request
--- a/iib/web/static/api_v1.yaml
+++ b/iib/web/static/api_v1.yaml
@@ -275,11 +275,15 @@ paths:
           description: services and workers are working as expected.
           content:
             application/json:
+              schema:
+                type: object
               example: {'status': 'Health check OK'}
         '503':
           description: one or more service/s are reporting an issue
           content:
             application/json:
+              schema:
+                type: object
               example: {'status': 'Health check failed', 'error': 'DATABASE unavailable: Database connection failed'}
   '/builds/{id}/logs':
     get:
--- a/tests/test_web/test_api_v1.py
+++ b/tests/test_web/test_api_v1.py
@@ -3,10 +3,10 @@ import json
 from unittest import mock
 
 import pytest
-from sqlalchemy.exc import DisconnectionError
 
 from iib.web.api_v1 import _get_unique_bundles
 from iib.web.models import Image, RequestAdd, RequestRm, RequestCreateEmptyIndex
+from iib.exceptions import IIBError
 
 
 def test_get_build(app, auth_env, client, db):
@@ -197,20 +197,6 @@ def test_get_builds_invalid_batch(batch, app, client, db):
     rv = client.get(f'/api/v1/builds?batch={batch}')
     assert rv.status_code == 400
     assert rv.json == {'error': 'The batch must be a positive integer'}
-
-
-@mock.patch('sqlalchemy.engine.base.Engine.execute')
-def test_get_healthcheck_db_fail(mock_db_execute, app, client, db):
-    mock_db_execute.side_effect = DisconnectionError('DB failed')
-    rv = client.get('/api/v1/healthcheck')
-    assert rv.status_code == 500
-    assert rv.json == {'error': 'Database health check failed.'}
-
-
-def test_get_healthcheck_ok(app, client, db):
-    rv = client.get('/api/v1/healthcheck')
-    assert rv.status_code == 200
-    assert rv.json == {'status': 'Health check OK'}
 
 
 @pytest.mark.parametrize(
@@ -2103,3 +2089,27 @@ def test_get_build_related_bundles_not_configured(client, db, minimal_request_re
     rv = client.get(f'/api/v1/builds/{request_id}')
     assert rv.status_code == 200
     assert 'related_bundles' not in rv.json
+
+
+@mock.patch('iib.web.api_v1.status')
+def test_get_status(mock_status, client):
+    mock_status.return_value = {'services': {}, 'workers': []}
+    rv = client.get('/api/v1/healthcheck/details')
+    assert rv.status_code == 200
+    assert rv.json == mock_status.return_value
+    mock_status.assert_called_once()
+
+
+@pytest.mark.parametrize('error', [None, 'something is wrong'])
+@mock.patch('iib.web.api_v1.status')
+def test_get_status_short(mock_status, error, client):
+    if error:
+        mock_status.side_effect = [IIBError(error)]
+    rv = client.get('/api/v1/healthcheck')
+
+    if error:
+        assert rv.json == {'status': 'Health check failed', 'error': error}
+        assert rv.status_code == 503
+    else:
+        assert rv.json == {'status': 'Health check OK'}
+        assert rv.status_code == 200
--- a/tests/test_web/test_healthcheck.py
+++ b/tests/test_web/test_healthcheck.py
@@ -1,0 +1,240 @@
+from unittest import mock
+
+import kombu
+import pytest
+import sqlalchemy
+
+from iib.exceptions import IIBError
+from iib.web import healthcheck
+
+
+@pytest.mark.parametrize(
+    'exc, exception_msg',
+    [
+        (None, None),
+        (sqlalchemy.exc.SQLAlchemyError('exception'), 'an error occurred'),
+        (
+            # sqlalchemy.exc.OperationalError('statement', 'params', 'orig')
+            sqlalchemy.exc.OperationalError('another exception', None, None),
+            'database connection failed',
+        ),
+    ],
+)
+@mock.patch.object(healthcheck.db, 'session')
+def test_database_connection(mock_db_session, exc, exception_msg):
+    session = mock_db_session.return_value
+
+    if exc is not None:
+        session.execute.side_effect = [exc]
+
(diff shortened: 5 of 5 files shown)


iib/web/healthcheck.py
```diff
@@ -0,0 +1,202 @@
...
+    else:
+        workers = []
+
+    any_worker_ok = any(worker["ok"] for worker in workers)
+    if short and not any_worker_ok:
+        raise IIBError("no workers are available")
+
+    return {
+        "services": services,
+        "software": package_version_check(),
```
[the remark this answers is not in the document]
zanssa: For retrieving version of the software, there was a suggestion to represent the version in the requests logs as there would be multiple workers in different VMs. hence, the removal for these functions in the PR.

zanssa:
conflicts have been resolved. TravisCI failing because of safety, which be handled in a different story!
I'll squash the commits once this is approved.

iib/web/status.py
```diff
@@ -0,0 +1,140 @@
...
+        workers.append(worker)
+
+    return workers
+
+
+def status(short=False, worker_ping_retries=2):
+    """
+    Get status of IIB workers and services that IIB depends on.
+
+    Display the current version of podman, skopeo, opm.
```
lipoja, line 105: This was for previous version of this function, right?

iib/web/status.py
```diff
@@ -0,0 +1,140 @@
...
+    :param bool short: raise an error as soon as any problem is found
+    :param int worker_ping_retries: number of retry/retries on ConnectionError when pinging workers
+    :return: dict with the following keys:
+        "services": list of status info for individual services
+        "software": list of softwares and their versions
```
lipoja, line 110: This was for previous version of this function, right?
Since I see return just "status" and "workers".

iib/web/status.py
```diff
@@ -0,0 +1,140 @@
...
+        workers = workers_status(retries=worker_ping_retries)
+    else:
+        workers = []
+
+    any_worker_ok = any(worker["ok"] for worker in workers)
+    if short and not any_worker_ok:
+        raise IIBError("no workers are available")
+
+    return {
+        "status": services,
```
lipoja, line 138: I am not sure if we can change type of "status". Until now it is plain text, however with this change we are returning dictionary. It can break parsing of the response.

iib/web/api_v1.py
```diff
@@ -314,23 +314,33 @@ def get_builds():
     return flask.jsonify(response)
 
 
-@api_v1.route('/healthcheck')
-def get_healthcheck():
+@api_v1.route("/status", methods=["GET"])
```
lipoja, line 317: Why are we removing `/healthcheck` endpoint?
This change would cause bumping API version. And I thought we do not want to do it if is not necessary. 
Adding things is OK, but deleting or changing might cause troubles.

lipoja (changes requested):

lipoja:
@zanssa I did not check everything since I think it should be modified once again. I think we should not make major changes to current API endpoints. We can add but we should not change or delete if it is not necessary.

iib/web/status.py
zanssa: Oops! Yeah, I missed it while reviewing.

iib/web/api_v1.py
zanssa: Oh! yeah so, I changed it to `status` when there were functions for retrieving the version of the pkgs as it would be more suitable than healthchek. Now, that version won't be part of this, I would change it back to the healthcheck.

midnightercz:
Looks good, but lets keep /healthcheck here

zanssa:
@lipoja @midnightercz 
so, I am keeping `healthcheck` with the same return type of `"status"`, which is string. If all is good, it will return this:

`{
  "status": "Health check OK"
}`

and I am adding another endpoint  `healthcheck/details` which will contain the new attributes ( services & workers ) 

Thanks for reviewing!

zanssa:
in last push: 
- replaced `ok` with `available`

iib/web/healthcheck.py
```diff
@@ -0,0 +1,138 @@
...
+    :return: tuple (ok: bool, reason: str or None)
+    """
+    with kombu.Connection(broker_url) as connection:
+        try:
+            connection.ensure_connection(max_retries=0)
+        except kombu.exceptions.OperationalError as OperError:
+            log.exception("broker connection failed, exception: %s", OperError)
+            return False, "broker connection failed"
+        except kombu.exceptions.KombuError as err:
+            log.error("an error occurred while trying to connect to the RabbitMQ server: %s", err)
```
lipoja, line 50: Could you please change double quotes " for single quotes '. In all files. Doc-string is exception.
It is noted in README.md -> `* Single quotes are used instead of double quotes`

iib/web/healthcheck.py
```diff
@@ -0,0 +1,138 @@
...
+    Check if RabbitMQ connection is working.
+
+    :param str broker_url: the AMQP(S) URL to connect to RabbitMQ
+    :return: tuple (ok: bool, reason: str or None)
+    """
+    with kombu.Connection(broker_url) as connection:
+        try:
+            connection.ensure_connection(max_retries=0)
+        except kombu.exceptions.OperationalError as OperError:
+            log.exception("broker connection failed, exception: %s", OperError)
```
lcarva, line 47: I think because you're using `log.exception`, you don't have to include the exception in the log message. The `exception` method will just automatically take the latest exception on the stack. Consider changing this line to: `log.exception("broker connection failed")`

iib/web/healthcheck.py
```diff
@@ -0,0 +1,138 @@
...
+    :return: dict with the following keys:
+        "services": list of status info for individual services
+        "workers": list of status info for individual workers
+    :raises IIBError: if short is True and a problem is found
+    """
+    services = []
+
+    def add_status(service_name, ok, reason):
+        if short and not ok:
+            raise IIBError(f"{service_name} unavailable: {reason}")
```
lcarva, line 116: If a service is not ok, then the other services are never checked. It may be desirable to always provide the status for every service.

iib/web/api_v1.py
```diff
@@ -314,23 +314,33 @@ def get_builds():
...
-    except Exception:
-        flask.current_app.logger.exception('DB test failed.')
-        raise IIBError('Database health check failed.')
+        status(short=True)
+        retval = {"status": "Health check OK"}
+    except IIBError as error:
+        retval = {"status": "Health check failed", "error": str(error)}
 
-    return flask.jsonify({'status': 'Health check OK'})
+    return flask.jsonify(retval), 200 if retval["status"] == "Health check OK" else 503
```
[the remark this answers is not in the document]
zanssa: It does return error. For example, in case db conncection failed 

`"status": "Health check failed", "error": "DATABASE unavailable: Database connection failed"`

may I ask what made you think it won't return any response text?

iib/web/healthcheck.py
zanssa: in case of short is True, it would only show the status in general (Status: health check failed and the reason) - just like the current healthcheck endpoint so nothing would break as it was suggested & mentioned in this PR. 

As this is a short healthcheck, if one of the services is not working, it should show the status as failed. 

If you wanna see the status of all services, you will need to check `healthcheck/details` - it contains a list of the services ( database & rabbitmq ), workers and their status.

iib/web/api_v1.py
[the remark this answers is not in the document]
lcarva: Oh, my bad! I had written the line as:
```python
    return (flask.jsonify(retval), 200) if retval["status"] == "Health check OK" else 503
```

When in fact, it's processed like this (added parenthesis for clarity):
```python
return flask.jsonify(retval), (200 if retval["status"] == "Health check OK" else 503)
```

This doesn't change my comment about the conditional being brittle.

iib/web/healthcheck.py
lcarva: Fair point. It's best to stop checking the other services in this case. Thanks for the clarification.

iib/web/api_v1.py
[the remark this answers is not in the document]
zanssa: I'll modify it! :)

iib/web/healthcheck.py
lipoja: Don't blame me, I like double quotes but we should keep code style in a project.

zanssa: I've just replaced them! :)

lcarva (dismissed):
Good job!

iib/web/static/api_v1.yaml
```diff
@@ -210,30 +210,77 @@ paths:
...
+      summary: Get a short IIB service status
+      description: >
+        if services including RabbitMQ, database, and workers
+        are working properly, return 200.
+      responses:
+        '200':
+          description: services and workers are working as expected.
           content:
             application/json:
-              schema:
```
pbortlov, line 231: Do we know if nobody uses the schema to automate if IIB works or not? 
I really like to use schema to at least show to have structured output.

tests/test_web/test_api_v1.py
```diff
@@ -2103,3 +2089,27 @@ def test_get_build_related_bundles_not_configured(client, db, minimal_request_re
...
+    mock_status.return_value = {'services': {}, 'workers': []}
+    rv = client.get('/api/v1/healthcheck/details')
+    assert rv.status_code == 200
+    assert rv.json == mock_status.return_value
+    mock_status.assert_called_once()
+
+
+@pytest.mark.parametrize('error', [None, 'something is wrong'])
+@mock.patch('iib.web.api_v1.status')
+def test_get_status_short(mock_status, error, client):
```
pbortlov, line 2105: Could we have tests for success and failure separated? The good practice is to use separate functions if you want to test `if` statement
- test_get_status_short_success
- test_get_status_short_failure

tests/test_web/test_healthcheck.py
```diff
@@ -0,0 +1,240 @@
...
+
+    config = mock_worker_config()
+
+    mock_get_worker_config.return_value = config
+
+    mock_database_connection_check.return_value = (True, None)
+    mock_rabbitmq_connection_check.return_value = (True, None)
+    mock_workers_status.return_value = workers_status
+
+    if short and not expected_result:
```
pbortlov, line 221: Could add to parametrize a variable error and error_message instead of adding `if` statement?

tests/test_web/test_api_v1.py
zanssa: aha. I'll keep that in mind when creating tests in the future!

iib/web/static/api_v1.yaml
zanssa: Can you elaborate more here how this would be used to automate if IIB works or not? 

Isn't ("Health Check OK"  or "Health Check failed") what would be used for validating the status of iib. am I missing something here?

pbortlov: The response  `example: {'status': 'Health check OK'}` is ok but the schema says that it is a json object. Having just an example doesn't say if it is a string or object.

I mean Semaphore used the JSON from the healthcheck to show if IIB is working. I am not sure if there are any similar scripts

tests/test_web/test_healthcheck.py
```diff
@@ -0,0 +1,240 @@
...
+        with pytest.raises(IIBError, match='no workers are available'):
+            healthcheck.status(short=True)
+        return
+
+    result = healthcheck.status(short=False)
```
lipoja, line 226: should not this `(short=False)` be `(short=short)` instead? Or why do you have the short parametrized?

zanssa: Can you explain to me why do you think it should be `short=short` and not `short=False` ?

iib/web/healthcheck.py
```diff
@@ -0,0 +1,202 @@
...
+    rabbit_ok, rabbit_reason = rabbitmq_connection_check(get_worker_config().broker_url)
+    add_status("RABBITMQ", rabbit_ok, rabbit_reason)
+    add_status("QUAY", *quay_status_check())
+
+    if rabbit_ok:
+        workers = workers_status(retries=worker_ping_retries)
+    else:
+        workers = []
+
+    any_worker_ok = any(worker["ok"] for worker in workers)
```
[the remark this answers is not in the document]
lipoja: @yashvardhannanavati could you help me remember the conclusion on this topic. Thanks!

iib/web/healthcheck.py
```diff
@@ -0,0 +1,202 @@
...
+
+
+def workers_status(retries=2):
+    """
+    Ping workers, check received replies.
+
+    :param int retries: number of retry/retries on ConnectionError when pinging workers
+    :return: list of workers status information (empty if no replies)
+    """
+    replies = _ping_celery_workers(retries=retries)
```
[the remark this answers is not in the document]
zanssa: I think I addressed this before here: https://github.com/release-engineering/iib/pull/316#discussion_r737681926 

you could also refer to the story description, and the last comment! I just updated the story to point out that version of software is removed from this and healthcheck/short is replaced by healthcheck/details!

iib/web/static/api_v1.yaml
zanssa: My understanding is that this file is only for documentation. correct ? 

I know Semaphore would be shutdown soon, and I am not familiar how it checks if iib is working or not based on the schema. So, I am looping @midnightercz as he knows more and can tell if it would break something or not. wdyt Jindrich?

midnightercz (dismissed):

tests/test_web/test_healthcheck.py
lipoja: Please, do not resolve conversation opened by others. I mean why not to use the variable from parametrize?

zanssa: I've always done it this way, resolving after addressing the reviewer's concerns. I marked it as resolved as there is no change required here. If you see there is another approach, that's great, but I don't think I'll change these tests :)

iib/web/static/api_v1.yaml
lipoja: @zanssa If we accept this change this new/modified endpoint will be the only one that does not have it documented.
And you are saying "only for documentation"? I mean - you never relied on documentation? Are you always looking to the code to see how to work with any API that you are using?

I agree with @pbortlov that it is good to have it documented. Why we should lower our standards with this PR?

Could you add the schema here, please?

tests/test_web/test_api_v1.py
lipoja: On the other hand, if you do it in this PR we do not have to modify it in a future, right?
You can have a look to this article: https://www.nerdwallet.com/blog/engineering/5-pytest-best-practices/ 
From the article:
> Parametrize the same behavior, have different tests for different behaviors
> 
> tl;dr: Parametrize when asserting the same behavior with various inputs and expected outputs. Make separate tests for distinct behaviors.

Correct me if I am wrong, but I think it is different behavior when test ends with error and when it is successful.

iib/web/static/api_v1.yaml
zanssa: Oh! I think the change is already approved :)

There is a difference between using the documentation as instructions and using the schema there to test if iib is working on not.

What is your standard that is lowered with this PR?

tests/test_web/test_api_v1.py
zanssa: Yeah, if it is that important, we can do it in the future! I think the tests in test_healthcheck file is enough. Plus, I don't have the capacity for further requests. :)

midnightercz (approved):

iib/web/static/api_v1.yaml
lipoja: For me the standard for every documented endpoint is to have this schema - therefore I would assume this one should have it as well.

iib/web/healthcheck.py
```diff
@@ -0,0 +1,138 @@
...
+
+
+log = logging.getLogger(__name__)
+
+
+def database_connection_check():
+    """
+    Check if the database connection is working.
+
+    :return: tuple (ok: bool, reason: str or None)
```
yashvardhannanavati, line 19: could you add `:raises` and `rtype` to the doc string.

yashvardhannanavati: Please refer other existing doc strings and keep the format uniform

iib/web/healthcheck.py
```diff
@@ -0,0 +1,138 @@
...
+        session.close()
+
+    return True, None
+
+
+def rabbitmq_connection_check(broker_url):
+    """
+    Check if RabbitMQ connection is working.
+
+    :param str broker_url: the AMQP(S) URL to connect to RabbitMQ
```
yashvardhannanavati, line 40: Same as above. The doc string needs to be fixed.

iib/web/healthcheck.py
```diff
@@ -0,0 +1,138 @@
...
+            return False, 'an error occurred'
+
+    return True, None
+
+
+def _ping_celery_workers(retries):
+    """
+    Attempt to ping worker, retry on ConnectionError.
+
+    :param int retries: number of retry/retries before returning an empty response
```
yashvardhannanavati, line 60: Same as above.

iib/web/healthcheck.py
```diff
@@ -0,0 +1,138 @@
...
+        return replies or {}
+
+    return {}
+
+
+def workers_status(retries=2):
+    """
+    Ping workers, check received replies.
+
+    :param int retries: number of retry/retries on ConnectionError when pinging workers
```
yashvardhannanavati, line 80: Same as above.

iib/web/healthcheck.py
```diff
@@ -0,0 +1,138 @@
...
+def status(short=False, worker_ping_retries=2):
+    """
+    Get status of IIB workers and services that IIB depends on.
+
+    :param bool short: raise an error as soon as any problem is found
+    :param int worker_ping_retries: number of retry/retries on ConnectionError when pinging workers
+    :return: dict with the following keys:
+        "services": list of status info for individual services
+        "workers": list of status info for individual workers
+    :raises IIBError: if short is True and a problem is found
```
yashvardhannanavati, line 110: Missing `:rtype`

iib/web/static/api_v1.yaml
yashvardhannanavati: @zanssa It shouldn't matter if the change is approved. One person might miss something which the other would notice. Please do address comments from all reviewers. We shouldn't merge a change unless concerns of all our guild members are resolved. 

fwiw, I agree we shouldn't cut corners in documentation and keep it as accurate as possible. One use case could be if another service uses IIB's endpoint to check its status and assumes it's going to be string instead of JSON from the documentation.

tests/test_web/test_healthcheck.py
yashvardhannanavati: @zanssa This comment was resolved without a response. Could you please explain why you prefer to keep it this way?

Ideally, it would be nice if you could split this into two different tests instead of one. But I am willing to look the other way

tests/test_web/test_healthcheck.py
yashvardhannanavati: @zanssa I agree, we should use the variable value instead of hard coded values here since you have parameterized the value already. 

It's very easy for someone else altering your test to miss the hard coded value. +1 on using the variable name.

yashvardhannanavati (changes requested):

iib/web/static/api_v1.yaml
zanssa: @yashvardhannanavati I addressed the comments from all and I even took into consideration what I was asked in gchat instead of here, and despite that some of them are not really necessary or even enforced in the repo.

I would expect some change requests and I don't really mind making the changes but it would be nice if those change requests are mentioned after reviewing once or twice; not one change every few days. This is just slowing the progress and not helping! :)

zanssa:
> @zanssa It shouldn't matter if the change is approved. One person might miss something which the other would notice. Please do address comments from all reviewers. We shouldn't merge a change unless concerns of all our guild members are resolved.

> fwiw, I agree we shouldn't cut corners in documentation and keep it as accurate as possible. One use case could be if another service uses IIB's endpoint to check its status and assumes it's going to be string instead of JSON from the documentation.


@yashvardhannanavati I addressed the comments from all and I even took into consideration what I was asked in gchat instead of here, and despite that some of them are not really necessary or even enforced in the repo.

I would expect some change requests and I don't really mind making the changes but it would be nice if those change requests are mentioned after reviewing once or twice; not one change every few days. This is just slowing the progress and not helping! :)

iib/web/healthcheck.py
zanssa: Sure! I can do that. :) 

There was a request to add these in another file, which I did. But I don't see this implemented in all files. That's why I didn't add it here; same thing with single & double quotes. I do understand that this is not a reason to not follow the format, but, if we want to enforce a rule, we should always apply it to all files, not just decide to enforce it when it suits us.

iib/web/static/api_v1.yaml
```diff
@@ -275,11 +275,15 @@ paths:
           description: services and workers are working as expected.
           content:
             application/json:
+              schema:
+                type: object
```
lipoja, line 279: Thank you for adding the schema, I really appreciate the effort!
However the schema was nested and there were also properties set for given object which is still missing.
Could you add correct properties for both responses (200 and 500)?

Thank you!

lipoja (changes requested):

iib/web/healthcheck.py
lipoja: I agree with you Zainab! It should be in all files. However I do not know how to automate these checks. So it might happen that developer forget to add it and it slips through code-review. 

If you see missing `:raises` and `rtype` feel free to create separate PR for that. Same with single/double quotes.

Or if you know a way how to automate this checks, please share it with us. It would be great to have these checks in our test suit - so it is really enforced.

zanssa: @lipoja you agreed and asked for it which means you believe it is important. but I don't see this. :) 

I reminded you about the code standard you have a concern about here and at the same time forgot about in one of your PRs. You changed it there and then sent another PR with the same missing code standard. 

I am not against it but let's all comply with the same standards. 

There is flake8 extension to lint for quotes (flake8-quotes) 

You could install it by pip and test it in your local. 

`pip install flake8-quotes
`

as an example, run this cmd: `flake8 --inline-quotes "'" iib/workers/tasks/utils.py`

the result should be similar to this: 

`iib/workers/tasks/utils.py:73:26: Q000 Double quotes found but single quotes preferred
iib/workers/tasks/utils.py:280:15: Q000 Double quotes found but single quotes preferred
iib/workers/tasks/utils.py:280:32: Q000 Double quotes found but single quotes preferred
iib/workers/tasks/utils.py:280:54: Q000 Double quotes found but single quotes preferred
iib/workers/tasks/utils.py:333:9: Q000 Double quotes found but single quotes preferred
iib/workers/tasks/utils.py:334:9: Q000 Double quotes found but single quotes preferred
iib/workers/tasks/utils.py:335:9: Q000 Double quotes found but single quotes preferred
iib/workers/tasks/utils.py:336:9: Q000 Double quotes found but single quotes preferred
iib/workers/tasks/utils.py:337:9: Q000 Double quotes found but single quotes preferred
iib/workers/tasks/utils.py:356:9: Q000 Double quotes found but single quotes preferred
iib/workers/tasks/utils.py:357:9: Q000 Double quotes found but single quotes preferred
iib/workers/tasks/utils.py:358:9: Q000 Double quotes found but single quotes preferred
iib/workers/tasks/utils.py:379:38: Q000 Double quotes found but single quotes preferred
iib/workers/tasks/utils.py:379:52: Q000 Double quotes found but single quotes preferred
iib/workers/tasks/utils.py:982:15: Q000 Double quotes found but single quotes preferred
iib/workers/tasks/utils.py:982:22: Q000 Double quotes found but single quotes preferred
iib/workers/tasks/utils.py:982:31: Q000 Double quotes found but single quotes preferred
iib/workers/tasks/utils.py:1007:38: Q000 Double quotes found but single quotes preferred
iib/workers/tasks/utils.py:1018:11: Q000 Double quotes found but single quotes preferred
iib/workers/tasks/utils.py:1018:25: Q000 Double quotes found but single quotes preferred
iib/workers/tasks/utils.py:1018:35: Q000 Double quotes found but single quotes preferred
iib/workers/tasks/utils.py:1018:56: Q000 Double quotes found but single quotes preferred
iib/workers/tasks/utils.py:1018:66: Q000 Double quotes found but single quotes preferred
iib/workers/tasks/utils.py:1018:82: Q000 Double quotes found but single quotes preferred
iib/workers/tasks/utils.py:1032:39: Q000 Double quotes found but single quotes preferred
iib/workers/tasks/utils.py:1066:43: Q000 Double quotes found but single quotes preferred
iib/workers/tasks/utils.py:1067:35: Q000 Double quotes found but single quotes preferred
iib/workers/tasks/utils.py:1094:25: Q000 Double quotes found but single quotes preferred
iib/workers/tasks/utils.py:1094:54: Q000 Double quotes found but single quotes preferred`

there is also support for multiline. For ref: https://github.com/zheller/flake8-quotes

iib/web/healthcheck.py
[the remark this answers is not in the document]
yashvardhannanavati: Thinking about it now, we should probably fail even if one of the workers is no working. Here's why, CVP might have 6 workers and 1 of them might be down. But CVP doesn't know which of its request will be routed to that 1 unstable worker giving them a bad experience. We as a team should be proactive to react to it rather than ignoring it as a warning. 

So I think simply listing the status of the workers as @zanssa mentioned is sufficient, let's just make the health check fail on that. @zanssa I apologize for this last minute change but it just occured to me. I would be ok, if you want to cover this change in a separate ticket and keep this PR working as we discussed previously

tests/test_web/test_api_v1.py
yashvardhannanavati: @zanssa Could you please make the change on this suggestion. I agree success and failure should be separate tests.

iib/web/healthcheck.py
yashvardhannanavati: @zanssa Thanks for pointing this out to us. I agree, it should be in all files. Maybe we missed it. Let's file an issue to add this enforcing? In the meanwhile, let's stay consistent going forward. I would really appreciate if you could make the change. Sorry to be nitpicking

yashvardhannanavati:
> > @zanssa It shouldn't matter if the change is approved. One person might miss something which the other would notice. Please do address comments from all reviewers. We shouldn't merge a change unless concerns of all our guild members are resolved.
> 
> > fwiw, I agree we shouldn't cut corners in documentation and keep it as accurate as possible. One use case could be if another service uses IIB's endpoint to check its status and assumes it's going to be string instead of JSON from the documentation.
> 
> @yashvardhannanavati I addressed the comments from all and I even took into consideration what I was asked in gchat instead of here, and despite that some of them are not really necessary or even enforced in the repo.
> 
> I would expect some change requests and I don't really mind making the changes but it would be nice if those change requests are mentioned after reviewing once or twice; not one change every few days. This is just slowing the progress and not helping! :)

I agree we need to be quick in reviewing and giving feedback. But at the same time, it's also true that whether the comment comes early or late, if it's enforcing good coding practices we should act on them. I understand your concern and we should definitely address it in the guild meeting next week.

iib/web/healthcheck.py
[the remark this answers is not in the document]
lipoja: @yashvardhannanavati Thanks for this idea, it is easier to implement then the solution I had in mind.

iib/web/healthcheck.py
lipoja: @zanssa Thanks for finding this out! Yes, as I said in https://github.com/release-engineering/iib/pull/316#discussion_r748579261 I do not like it and I prefer double quotes. Therefore I have hard time to write code with single quotes as well. It would be great if we can enforce it.

pbortlov:
@release-engineering/exd-guild-hello-operator Could we do a final review?

PR #318, opened 2021-11-01
Upgrade werkzeug to fix failing safety check

yashvardhannanavati:


yashvardhannanavati:
This is going to require more work. Will file a ticket for it.

PR #320, opened 2021-11-03
[CLOUDDST-7664]  Handle OPM deprecation bundle error

JAVGan:
- new regex to handle OPM errors on top of help display
- maintains the previous logic for legacy versions of OPM (error on bottom of help display)

### DIFF
--- a/iib/workers/tasks/utils.py
+++ b/iib/workers/tasks/utils.py
@@ -723,6 +723,13 @@ def run_cmd(cmd, params=None, exc_msg=None, strict=True):
     if strict and response.returncode != 0:
         log.error('The command "%s" failed with: %s', ' '.join(cmd), response.stderr)
         if cmd[0] == 'opm':
+            # The new OPM shows the error at the beginning, let's try with this first
+            regex = r'^(?:Error: )(.+)'
+            match = re.match(regex, response.stderr)
+            if match:
+                raise IIBError(f'{exc_msg.rstrip(".")}: {match.groups()[0]}')
+            # If we didn't find the message at the beginning, let's look for it at the end
+            # after the help message
             # Capture the error message right before the help display
             regex = r'^(?:Error: )(.+)$'
             # Start from the last log message since the failure occurs near the bottom


iib/workers/tasks/utils.py
```diff
@@ -723,13 +723,23 @@ def run_cmd(cmd, params=None, exc_msg=None, strict=True):
...
-                match = re.match(regex, msg)
-                if match:
-                    raise IIBError(f'{exc_msg.rstrip(".")}: {match.groups()[0]}')
+            # The new OPM shows the error at the beginning, let's try with this first
+            regex = r'^(?:Error: )(.+)'
+            msg = response.stderr.splitlines()
+            match = re.match(regex, "".join(msg))
+            if match:
+                raise IIBError(f'{exc_msg.rstrip(".")}: {match.groups()[0]}')
+            else:
```
lipoja, line 732: nitpick - This `else` is not necessary, you can save indentation.

JAVGan: Thanks! I'll change it.

JAVGan: I've sent the amended commit

iib/workers/tasks/utils.py
```diff
@@ -723,11 +723,19 @@ def run_cmd(cmd, params=None, exc_msg=None, strict=True):
     if strict and response.returncode != 0:
         log.error('The command "%s" failed with: %s', ' '.join(cmd), response.stderr)
         if cmd[0] == 'opm':
+            # The new OPM shows the error at the beginning, let's try with this first
+            regex = r'^(?:Error: )(.+)'
+            msg = response.stderr.splitlines()
+            match = re.match(regex, "".join(msg))
```
lipoja, line 729: @JAVGan Could you help understand how does this work?

 It looks like it will create list of lines from `stderr`; all new line characters `\n` are removed by [splitlines](https://docs.python.org/3/library/stdtypes.html#str.splitlines) and then the message is joined to one long string. 
It means it returns everything - not only the error message? Or there is nothing else on `stderr` just this error message?

Thanks!

JAVGan: Hi @lipoja, you're correct.

The `response.stderr` contains a string with the entire error message from `subprocess.Popen`.

In the original code they were creating a list of strings for each line to search for the error in reversed way for each substring, like this:

```
for msg in reversed(response.stderr.splitlines()):
    match = re.match(regex, msg)
    ...
```

Then I tried to combine the legacy code with my new and moved this command to `msg` in line 728 to use in both regex, but in fact it doesn't need to be like this.

Since I'm splitting and then joining it make no sense to keep this instruction like this.

I can change it to parse the entire string in my regex and let the split just occur in the reverse search.

JAVGan: Pushing the amended code

iib/workers/tasks/utils.py
```diff
@@ -723,6 +723,13 @@ def run_cmd(cmd, params=None, exc_msg=None, strict=True):
     if strict and response.returncode != 0:
         log.error('The command "%s" failed with: %s', ' '.join(cmd), response.stderr)
         if cmd[0] == 'opm':
+            # The new OPM shows the error at the beginning, let's try with this first
```
lipoja, line 726: I would suggest to use `opm` version here in the comment. "new OMP" is relative and it might mean something else in few years.

JAVGan: As far as I looked on OPM versions each one puts the error on top, I don't know why they decided to use the reverse search but the comment `#Capture the error message right before the help display` might indicate that it was always on top.

So I believe I was touching in the wrong part of the code. The regex is not broken, if we look in the [logs](https://iib.qe.engineering.redhat.com/api/v1/builds/5924/logs) near bottom it says:

```iib.exceptions.IIBError: Failed to deprecate the bundles: error deprecating bundle registry-proxy.engineering.redhat.com/rh-osbs/amq7-amq-streams-rhel7-operator-metadata@sha256:25b8153b1f5cace8963715dad8019b07fbbc2501d42fb21b189193baa596a41d: Bundle deprecation causing default channel removal```

Which indicates IIB raised the exception, thus regex always worked fine.

Doing the tests locally I confirmed that original regex works fine, so that's not the problem, the problem is that user received the message `An unknown error occurred. See logs for details` which is the default `exec_msg` when not set. Instead, the user might receive  `Bundle deprecation causing default channel removal` or something similar.

Thus the solution need to be different.

As I see [here](https://github.com/operator-framework/operator-registry/releases/tag/v1.18.0) OPM 1.18 seems to have support for default channel removal, so I need to implement something to detect if OPM returned `Bundle deprecation causing default channel removal` and then return this message to the user instead of default `exec_msg`

For the moment I'll close the PR and put this information on JIRA.

Thanks for reviewing!

PR #323, opened 2021-11-05
Updating serve index methods to serve FBC image

lipoja:
CLOUDDST-8300

### DIFF
--- a/iib/workers/config.py
+++ b/iib/workers/config.py
@@ -16,7 +16,7 @@ class Config(object):
         os.path.expanduser('~'), '.docker', 'config.json.template'
     )
     iib_greenwave_url = None
-    iib_grpc_init_wait_time = 3
+    iib_grpc_init_wait_time = 10
     iib_grpc_max_port_tries = 100
     iib_grpc_max_tries = 5
     iib_grpc_start_port = 50051
@@ -142,6 +142,7 @@ class TestingConfig(DevelopmentConfig):
     iib_request_related_bundles_dir = None
     # disable dogpile cache for tests
     iib_dogpile_backend = 'dogpile.cache.null'
+    iib_grpc_init_wait_time = 3
 
 
 def configure_celery(celery_app):
--- a/iib/workers/config.py
+++ b/iib/workers/config.py
@@ -46,6 +46,8 @@ class Config(object):
         'iib.workers.tasks.build_create_empty_index',
         'iib.workers.tasks.general',
     ]
+    # Path to hidden location of SQLite database
+    hidden_index_db_path = '/var/lib/iib/_hidden/do.not.edit.db'
     # The task messages will be acknowledged after the task has been executed,
     # instead of just before
     task_acks_late = True
--- a/iib/workers/config.py
+++ b/iib/workers/config.py
@@ -56,6 +56,8 @@ class Config(object):
     task_default_routing_key = 'iib'
     # Requeue the message if the worker abruptly exits or is signaled
     task_reject_on_worker_lost = True
+    # Path to index.db in our temp directories used in IIB code
+    temp_index_db_path = 'database/index.db'
     # For now, only allow a single process so that all tasks are processed serially
     worker_concurrency = 1
     # Don't allow the worker to fetch more messages than it can handle at a time. This is so that
@@ -142,7 +144,6 @@ class TestingConfig(DevelopmentConfig):
     iib_request_related_bundles_dir = None
     # disable dogpile cache for tests
     iib_dogpile_backend = 'dogpile.cache.null'
-    iib_grpc_init_wait_time = 3
 
 
 def configure_celery(celery_app):
--- a/iib/workers/tasks/fbc_utils.py
+++ b/iib/workers/tasks/fbc_utils.py
@@ -6,8 +6,6 @@ import logging
 from iib.exceptions import IIBError
 from iib.workers.config import get_worker_config
 
-# path to index.db in our temp directories used in IIB code
-TEMP_INDEX_DB_PATH = 'database/index.db'
 
 log = logging.getLogger(__name__)
 
@@ -32,7 +30,7 @@ def is_image_fbc(image):
     return fbc_image_label in skopeo_output['Labels']
 
 
-def get_index_fbc_dir(from_index, base_dir):
+def get_catalog_dir(from_index, base_dir):
     """
     Get file-based catalog directory from the specified index image and save it locally.
 
@@ -45,6 +43,7 @@ def get_index_fbc_dir(from_index, base_dir):
     from iib.workers.tasks.build import _copy_files_from_image
     from iib.workers.tasks.utils import get_image_label
 
+    log.info("Store file-based catalog directory from %s", from_index)
     fbc_dir_path = get_image_label(from_index, 'operators.operatorframework.io.index.configs.v1')
     if not fbc_dir_path:
         error_msg = f'Index image {from_index} does not contain file-based catalog.'
@@ -66,8 +65,9 @@ def get_hidden_index_database(from_index, base_dir):
     """
     from iib.workers.tasks.build import _copy_files_from_image
 
-    db_path = get_worker_config()['hidden_index_db_path']
-    base_db_file = os.path.join(base_dir, TEMP_INDEX_DB_PATH)
+    log.info("Store hidden index.db from %s", from_index)
+    conf = get_worker_config()
+    base_db_file = os.path.join(base_dir, conf['temp_index_db_path'])
     os.makedirs(os.path.dirname(base_db_file), exist_ok=True)
-    _copy_files_from_image(from_index, db_path, base_db_file)
+    _copy_files_from_image(from_index, conf['hidden_index_db_path'], base_db_file)
     return base_db_file
--- a/iib/workers/tasks/fbc_utils.py
+++ b/iib/workers/tasks/fbc_utils.py
@@ -28,8 +28,8 @@ def is_image_fbc(image):
     from iib.workers.tasks.utils import skopeo_inspect
 
     skopeo_output = skopeo_inspect(f'docker://{image}')
-    dc_image_label = 'operators.operatorframework.io.index.configs.v1'
-    return dc_image_label in skopeo_output['Labels']
+    fbc_image_label = 'operators.operatorframework.io.index.configs.v1'
+    return fbc_image_label in skopeo_output['Labels']
 
 
 def get_index_fbc_dir(from_index, base_dir):
--- a/iib/workers/tasks/fbc_utils.py
+++ b/iib/workers/tasks/fbc_utils.py
@@ -1,7 +1,15 @@
 # SPDX-License-Identifier: GPL-3.0-or-later
 # This file contains functions that are common for File-Based Catalog image type
+import os
+import logging
 
-from iib.workers.tasks.utils import skopeo_inspect
+from iib.exceptions import IIBError
+from iib.workers.config import get_worker_config
+
+# path to index.db in our temp directories used in IIB code
+TEMP_INDEX_DB_PATH = 'database/index.db'
+
+log = logging.getLogger(__name__)
 
 
 def is_image_fbc(image):
@@ -17,6 +25,49 @@ def is_image_fbc(image):
     :return: True if image is FBC type, False otherwise (SQLite)
     :rtype: bool
     """
+    from iib.workers.tasks.utils import skopeo_inspect
+
     skopeo_output = skopeo_inspect(f'docker://{image}')
     dc_image_label = 'operators.operatorframework.io.index.configs.v1'
     return dc_image_label in skopeo_output['Labels']
+
+
+def get_index_fbc_dir(from_index, base_dir):
+    """
+    Get file-based catalog directory from the specified index image and save it locally.
+
+    :param str from_index: index image to get file-based catalog directory from.
+    :param str base_dir: base directory to which the database file should be saved.
+    :return: path to the copied file-based catalog directory.
+    :rtype: str
+    :raises IIBError: if any podman command fails.
+    """
+    from iib.workers.tasks.build import _copy_files_from_image
+    from iib.workers.tasks.utils import get_image_label
+
+    fbc_dir_path = get_image_label(from_index, 'operators.operatorframework.io.index.configs.v1')
+    if not fbc_dir_path:
+        error_msg = f'Index image {from_index} does not contain file-based catalog.'
+        log.error(error_msg)
+        raise IIBError(error_msg)
+
+    _copy_files_from_image(from_index, fbc_dir_path, base_dir)
+    return os.path.join(base_dir, os.path.basename(fbc_dir_path))
+
+
+def get_hidden_index_database(from_index, base_dir):
+    """
+    Get hidden database file from the specified index image and save it locally.
+
+    :param str from_index: index image to get database file from.
+    :param str base_dir: base directory to which the database file should be saved.
+    :return: path to the copied database file.
+    :rtype: str
+    """
+    from iib.workers.tasks.build import _copy_files_from_image
+
+    db_path = get_worker_config()['hidden_index_db_path']
+    base_db_file = os.path.join(base_dir, TEMP_INDEX_DB_PATH)
+    os.makedirs(os.path.dirname(base_db_file), exist_ok=True)
+    _copy_files_from_image(from_index, db_path, base_db_file)
+    return base_db_file
--- a/iib/workers/tasks/opm_operations.py
+++ b/iib/workers/tasks/opm_operations.py
@@ -1,0 +1,192 @@
+import logging
+import os
+import socket
+import subprocess
+import time
+
+from retry import retry
+
+from iib.exceptions import AddressAlreadyInUse, IIBError
+from iib.workers.config import get_worker_config
+from iib.workers.tasks.fbc_utils import is_image_fbc, get_catalog_dir
+
+log = logging.getLogger(__name__)
+
+
+def _get_free_port(port_start, port_end):
+    """
+    Return free port that is safe to use for opm command.
+
+    :param int port_start: port from which we should start trying to connect to
+    :param int port_end: port on which we should stop trying
+    :return: free port from given interval
+    :rtype: int
+    :raises IIBError: if all tried ports are in use
+    """
+    log.debug('Finding free port from interval [%d,%d)', port_start, port_end)
+    for port in range(port_start, port_end):
+        sock = socket.socket()
+        try:
+            sock.bind(('', port))
+            sock.close()
+            log.debug('Free port found: %d', port)
+            return port
+
+        except OSError:
+            log.info('Port %d is in use, trying another.', port)
+            port += 1
+        sock.close()
+
+    err_msg = f'No free port has been found after {port_end - port_start} attempts.'
+    log.error(err_msg)
+    raise IIBError(err_msg)
+
+
+def _get_free_port_for_grpc():
+    """Return free port for gRPC service from range set in IIB config."""
+    log.debug('Finding free port for gRPC')
+    conf = get_worker_config()
+    port_start = conf['iib_grpc_start_port']
+    port_end = port_start + conf['iib_grpc_max_port_tries']
+
+    return _get_free_port(port_start, port_end)
+
+
+def opm_serve_from_index(base_dir, from_index):
+    """
+    Locally start OPM registry service, which can be communicated with using gRPC queries.
+
+    Due to IIB's paralellism, the service can run multiple times, which could lead to port
+    binding conflicts. Resolution of port conflicts is handled in this function as well.
+
+    :param str base_dir: base directory to create temporary files in.
+    :param str from_index: index image to inspect.
+    :return: tuple containing port number of the running service and the running Popen object.
+    :rtype: (int, Popen)
+    """
+    from iib.workers.tasks.build import _get_index_database
+
+    log.info('Serving data from image %s', from_index)
+    if not is_image_fbc(from_index):
+        db_path = _get_index_database(from_index, base_dir)
+        return opm_registry_serve(db_path)
+
+    catalog_dir = get_catalog_dir(from_index, base_dir)
+    return opm_serve(catalog_dir)
+
+
+@retry(exceptions=AddressAlreadyInUse, tries=get_worker_config().iib_grpc_max_tries, logger=log)
+def opm_serve(catalog_dir):
+    """
+    Locally start OPM service, which can be communicated with using gRPC queries.
+
+    Due to IIB's paralellism, the service can run multiple times, which could lead to port
+    binding conflicts. Resolution of port conflicts is handled in this function as well.
+
+    :param str catalog_dir: path to file-based catalog directory that should be served.
+    :return: tuple containing port number of the running service and the running Popen object.
+    :rtype: (int, Popen)
+    """
+    log.info('Serving data from file-based catalog %s', catalog_dir)
+
+    port = _get_free_port_for_grpc()
+    cmd = ['opm', 'serve', catalog_dir, '-p', str(port), '-t', '/dev/null']
+    cwd = os.path.abspath(os.path.join(catalog_dir, os.path.pardir))
+    return (
+        port,
+        _serve_cmd_at_port_defaults(cmd, cwd, port),
+    )
+
+
+@retry(exceptions=AddressAlreadyInUse, tries=get_worker_config().iib_grpc_max_tries, logger=log)
+def opm_registry_serve(db_path):
+    """
+    Locally start OPM registry service, which can be communicated with using gRPC queries.
+
+    Due to IIB's paralellism, the service can run multiple times, which could lead to port
+    binding conflicts. Resolution of port conflicts is handled in this function as well.
+
+    :param str db_path: path to index database containing the registry data.
+    :return: tuple containing port number of the running service and the running Popen object.
+    :rtype: (int, Popen)
+    """
+    log.info('Serving data from index.db %s', db_path)
+
+    port = _get_free_port_for_grpc()
+    cmd = ['opm', 'registry', 'serve', '-p', str(port), '-d', db_path, '-t', '/dev/null']
+    cwd = os.path.dirname(db_path)
+    return (
+        port,
+        _serve_cmd_at_port_defaults(cmd, cwd, port),
+    )
+
+
+def _serve_cmd_at_port_defaults(serve_cmd, cwd, port):
+    """
+    Call `_serve_cmd_at_port()` with default values from IIB config.
+
+    :param list serve_cmd: opm command to be run (serve FBC or index.db).
+    :param str cwd: path to folder which should be set as current working directory.
+    :param str int port: port to start the service on.
+    """
+    log.debug('Run _serve_cmd_at_port with default loaded from IIB config.')
+    conf = get_worker_config()
+    return _serve_cmd_at_port(
+        serve_cmd, cwd, port, conf['iib_grpc_max_tries'], conf['iib_grpc_init_wait_time']
+    )
+
+
+@retry(exceptions=IIBError, tries=2, logger=log)
+def _serve_cmd_at_port(serve_cmd, cwd, port, max_tries, wait_time):
+    """
+    Start an opm service at a specified port.
+
+    :param list serve_cmd: opm command to be run (serve FBC or index.db).
+    :param str cwd: path to folder which should be set as current working directory.
+    :param str int port: port to start the service on.
+    :param max_tries: how many times to try to start the service before giving up.
+    :param wait_time: time to wait before checking if the service is initialized.
+    :return: object of the running Popen process.
+    :rtype: subprocess.Popen
+    :raises IIBError: if the process has failed to initialize too many times, or an unexpected
+        error occurred.
+    :raises AddressAlreadyInUse: if the specified port is already being used by another service.
+    """
+    from iib.workers.tasks.utils import run_cmd
+
+    log.debug('Run command %s with up to %d retries', ' '.join(serve_cmd), max_tries)
+    for _ in range(max_tries):
+        rpc_proc = subprocess.Popen(
+            serve_cmd,
+            cwd=cwd,
+            stdout=subprocess.PIPE,
+            stderr=subprocess.PIPE,
+            universal_newlines=True,
+        )
+        start_time = time.time()
+        while time.time() - start_time < wait_time:
+            time.sleep(1)
+            ret = rpc_proc.poll()
+            # process has terminated
+            if ret is not None:
+                stderr = rpc_proc.stderr.read()
+                if 'address already in use' in stderr:
+                    raise AddressAlreadyInUse(f'Port {port} is already used by a different service')
+                raise IIBError(f'Command "{" ".join(serve_cmd)}" has failed with error "{stderr}"')
+
+            # query the service to see if it has started
+            try:
+                output = run_cmd(
+                    ['grpcurl', '-plaintext', f'localhost:{port}', 'list', 'api.Registry']
+                )
+            except IIBError:
+                output = ''
+
+            if 'api.Registry.ListBundles' in output or 'api.Registry.ListPackages' in output:
+                log.debug('Started the command "%s"', ' '.join(serve_cmd))
+                log.info('Index registry service has been initialized.')
+                return rpc_proc
+
+        rpc_proc.kill()
+
+    raise IIBError(f'Index registry has not been initialized after {max_tries} tries')
--- a/iib/workers/tasks/utils.py
+++ b/iib/workers/tasks/utils.py
@@ -24,7 +24,7 @@ from operator_manifest.operator import ImageName
 from iib.exceptions import IIBError, AddressAlreadyInUse
 from iib.workers.config import get_worker_config
 from iib.workers.api_utils import set_request_state
-from iib.workers.tasks.fbc_utils import is_image_fbc, get_index_fbc_dir
+from iib.workers.tasks.fbc_utils import is_image_fbc, get_index_fbc_dir, TEMP_INDEX_DB_PATH
 
 log = logging.getLogger(__name__)
 dogpile_cache_region = create_dogpile_region()
@@ -71,7 +71,7 @@ def add_max_ocp_version_property(resolved_bundles, temp_dir):
     :param str temp_dir: directory location of the index image
     """
     # Get the CSV name and version (not just the bundle path)
-    db_path = temp_dir + "/database/index.db"
+    db_path = os.path.join(temp_dir, TEMP_INDEX_DB_PATH)
     port, rpc_proc = serve_registry(os.path.dirname(db_path), db_path=db_path)
 
     raw_bundles = run_cmd(
--- a/iib/workers/tasks/utils.py
+++ b/iib/workers/tasks/utils.py
@@ -10,7 +10,6 @@ import os
 import re
 import sqlite3
 import subprocess
-import time
 
 from retry import retry
 
@@ -21,10 +20,10 @@ from iib.workers.dogpile_cache import (
 )
 from operator_manifest.operator import ImageName
 
-from iib.exceptions import IIBError, AddressAlreadyInUse
+from iib.exceptions import IIBError
 from iib.workers.config import get_worker_config
 from iib.workers.api_utils import set_request_state
-from iib.workers.tasks.fbc_utils import is_image_fbc, get_index_fbc_dir, TEMP_INDEX_DB_PATH
+from iib.workers.tasks.opm_operations import opm_registry_serve, opm_serve_from_index
 
 log = logging.getLogger(__name__)
 dogpile_cache_region = create_dogpile_region()
@@ -71,8 +70,9 @@ def add_max_ocp_version_property(resolved_bundles, temp_dir):
     :param str temp_dir: directory location of the index image
     """
     # Get the CSV name and version (not just the bundle path)
-    db_path = os.path.join(temp_dir, TEMP_INDEX_DB_PATH)
-    port, rpc_proc = serve_registry(os.path.dirname(db_path), db_path=db_path)
+    temp_index_db_path = get_worker_config()['temp_index_db_path']
+    db_path = os.path.join(temp_dir, temp_index_db_path)
+    port, rpc_proc = opm_registry_serve(db_path=db_path)
 
     raw_bundles = run_cmd(
         ['grpcurl', '-plaintext', f'localhost:{port}', 'api.Registry/ListBundles'],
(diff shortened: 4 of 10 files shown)


iib/workers/tasks/utils.py
```diff
@@ -175,28 +176,49 @@ def _requires_max_ocp_version(bundle):
     return False
 
 
-def serve_index_registry(db_path):
+def serve_registry(base_dir, from_index=None, fbc_dir=None, db_path=None):
```
pbortlov, line 179: if you change this.. you will have to change it in build_create_empty_index.py, build_merge_index_image.py

pbortlov: it would be better to do the refactoring out of this PR

iib/workers/tasks/utils.py
```diff
@@ -208,25 +230,25 @@ def serve_index_registry(db_path):
 
 
 @retry(exceptions=IIBError, tries=2, logger=log)
-def serve_index_registry_at_port(db_path, port, max_tries, wait_time):
+def serve_cmd_at_port(serve_cmd, cwd, port, max_tries, wait_time):
```
pbortlov, line 233: Careful with refactoring

pbortlov (changes requested):

iib/workers/tasks/utils.py
lipoja: I hope that I am careful. Do you see something that I do not? Maybe I overlooked something...
Thanks

iib/workers/tasks/utils.py
lipoja: Those two does not use `serve_registry` directly.

iib/workers/tasks/fbc_utils.py
```diff
@@ -17,6 +19,45 @@ def is_image_fbc(image):
     :return: True if image is FBC type, False otherwise (SQLite)
     :rtype: bool
     """
+    from iib.workers.tasks.utils import skopeo_inspect
+
     skopeo_output = skopeo_inspect(f'docker://{image}')
     dc_image_label = 'operators.operatorframework.io.index.configs.v1'
```
chandwanitulsi, line 25: Minor and unrelated observation: dc_image_label should be renamed to fbc_image_label in a separate PR

iib/workers/tasks/fbc_utils.py
```diff
@@ -17,6 +19,45 @@ def is_image_fbc(image):
...
     return dc_image_label in skopeo_output['Labels']
+
+
+def _get_index_fbc_dir(from_index, base_dir):
+    """
+    Get file-based catalog directory from the specified index image and save it locally.
+
+    :param str from_index: index image to get database file from.
+    :param str base_dir: base directory to which the database file should be saved.
```
chandwanitulsi, line 34: I am a little confused here, do we get the database file from the FBC index image? Can you probably elaborate what you mean by "database file" in these 2 lines?

iib/workers/tasks/fbc_utils.py
```diff
@@ -17,6 +19,45 @@ def is_image_fbc(image):
...
+    :return: path to the copied file-based catalog directory.
+    :rtype: str
+    :raises IIBError: if any podman command fails.
+    """
+    from iib.workers.tasks.build import _copy_files_from_image
+    from iib.workers.tasks.utils import get_image_label
+
+    db_path = get_image_label(from_index, 'operators.operatorframework.io.index.configs.v1')
```
chandwanitulsi, line 42: should this be catalog_path or fbc_dir as you used in other places instead?

iib/workers/tasks/fbc_utils.py
lipoja: copy-paste error ... I forgot to rewrite it to FBC

iib/workers/config.py
```diff
@@ -46,6 +46,8 @@ class Config(object):
         'iib.workers.tasks.build_create_empty_index',
         'iib.workers.tasks.general',
     ]
+    # Path to hidden location of SQLite database
+    hidden_index_db_path = '.iib/index.db'
```
pbortlov, line 50: A thought: Is it wise to add 'hidden_path' to shared config? 
I would prefer to add an empty variable only.

iib/workers/tasks/fbc_utils.py
```diff
@@ -17,6 +19,45 @@ def is_image_fbc(image):
     :return: True if image is FBC type, False otherwise (SQLite)
     :rtype: bool
     """
+    from iib.workers.tasks.utils import skopeo_inspect
+
     skopeo_output = skopeo_inspect(f'docker://{image}')
-    dc_image_label = 'operators.operatorframework.io.index.configs.v1'
-    return dc_image_label in skopeo_output['Labels']
+    fbc_image_label = 'operators.operatorframework.io.index.configs.v1'
```
pbortlov, line 25: could we use function `get_image_labels()` from `utils.py` line #528?

iib/workers/tasks/utils.py
```diff
@@ -24,6 +24,7 @@
 from iib.exceptions import IIBError, AddressAlreadyInUse
 from iib.workers.config import get_worker_config
 from iib.workers.api_utils import set_request_state
+from iib.workers.tasks.fbc_utils import is_image_fbc, _get_index_fbc_dir
```
pbortlov, line 27: the function `_get_index_fbc_dir` should be without `_` at the beginning if we import it here

iib/workers/tasks/utils.py
```diff
@@ -175,28 +176,49 @@ def _requires_max_ocp_version(bundle):
...
 
+    :param str base_dir: base directory to create temporary files in.
+    :param str from_index: index image to inspect.
+    :param str fbc_dir: path to file-based catalog directory.
     :param str db_path: path to index database containing the registry data.
     :return: tuple containing port number of the running service and the running Popen object.
     :rtype: (int, Popen)
     :raises IIBError: if all tried ports are in use, or the command failed for another reason.
     """
+    from iib.workers.tasks.build import _get_index_database
```
pbortlov, line 194: the function _get_index_database should be without _ at the beginning if we import it here

iib/workers/tasks/utils.py
```diff
@@ -175,28 +176,49 @@ def _requires_max_ocp_version(bundle):
...
+
     conf = get_worker_config()
     port_start = conf['iib_grpc_start_port']
     port_end = port_start + conf['iib_grpc_max_port_tries']
 
+    assert (
```
pbortlov, line 200: Could you please explain why is the assert use here? I am curious :)

iib/workers/tasks/utils.py
```diff
@@ -175,28 +176,49 @@ def _requires_max_ocp_version(bundle):
...
+        len([i for i in (from_index, fbc_dir, db_path) if i]) == 1
+    ), "Only one keyword parameter can be set"
+
+    if from_index:
+        if not is_image_fbc(from_index):
```
pbortlov, line 205: If we use get_image_label we don't have to use this function

iib/workers/config.py
lipoja: I mean anybody can extract the index image and see what is inside, right?

iib/workers/tasks/utils.py
lipoja: I am thinking about asserts as check for developers that they did not do anything that is not supported. In this case I wanted to be sure that nobody will use this function without any keyword argument or with multiple arguments set. Because it would not work otherwise.

iib/workers/tasks/utils.py
lipoja: That is true. I did not thought about it at first. On the other hand, this will prevent having the long name of the label in multiple places. But I guess you are right, we could have done that.

iib/workers/config.py
lipoja: @camilamacedo86 @gallettilance - I've suggested this path for `index.db`

Note: This is still in code review and things might change.

iib/workers/tasks/utils.py
```diff
@@ -175,28 +176,49 @@ def _requires_max_ocp_version(bundle):
...
     """
+    from iib.workers.tasks.build import _get_index_database
+
     conf = get_worker_config()
     port_start = conf['iib_grpc_start_port']
     port_end = port_start + conf['iib_grpc_max_port_tries']
 
+    assert (
+        len([i for i in (from_index, fbc_dir, db_path) if i]) == 1
+    ), "Only one keyword parameter can be set"
```
zanssa, line 202: Shouldn't single quotes be used here?

lipoja: +1

iib/workers/tasks/fbc_utils.py
lipoja: https://github.com/release-engineering/iib/pull/328

iib/workers/config.py
chandwanitulsi: Hey @gallettilance I was hoping to verify whether we want to keep this hidden path completely hidden from customers? I mean theoretically this path can still be seen by anyone from IIB source code. OR Is it sufficient that index.db is not directly available at older location now, so its semi-hidden

iib/workers/tasks/fbc_utils.py
```diff
@@ -17,6 +19,45 @@ def is_image_fbc(image):
...
+    """
+    Get file-based catalog directory from the specified index image and save it locally.
+
+    :param str from_index: index image to get file-based catalog directory from.
+    :param str base_dir: base directory to which the database file should be saved.
+    :return: path to the copied file-based catalog directory.
+    :rtype: str
+    :raises IIBError: if any podman command fails.
+    """
+    from iib.workers.tasks.build import _copy_files_from_image
```
pbortlov, line 39: Here we have again private function imported to different file

iib/workers/tasks/fbc_utils.py
```diff
@@ -17,6 +19,45 @@ def is_image_fbc(image):
...
+    """
+    Get hidden database file from the specified index image and save it locally.
+
+    :param str from_index: index image to get database file from.
+    :param str base_dir: base directory to which the database file should be saved.
+    :return: path to the copied database file.
+    :rtype: str
+    :raises IIBError: if any podman command fails.
+    """
+    from iib.workers.tasks.build import _copy_files_from_image
```
pbortlov, line 59: private function

pbortlov (changes requested):

iib/workers/tasks/fbc_utils.py
lipoja: I agree with you! 
However I am not sure if I am able to refactor everything in this PR.
`_copy_files_from_image` is already imported like this in `build_regenerate_bundle.py `

lipoja: Let's address this issue in CLOUDDST-10478, OK?

iib/workers/tasks/fbc_utils.py
lipoja: https://github.com/release-engineering/iib/pull/323#discussion_r754223192

iib/workers/tasks/fbc_utils.py
lipoja: I agree - let me address that in separate PR because it is not direct part of this PR.
Thanks!

pbortlov: Are we going to have a new PR?

iib/workers/tasks/fbc_utils.py
pbortlov: @release-engineering/exd-guild-hello-operator PTAL

iib/workers/tasks/utils.py
pbortlov: any update here?

iib/workers/config.py
gallettilance: it's sufficient for it not to be available at the old location

gallettilance: the label on the image pointing to the index.db location should not be there anyway so OPM shouldn't be able to access it even if it were at the old location.

iib/workers/tasks/utils.py
chandwanitulsi: I think we had a PR about this one, but it was closed later

lipoja: Because  I am torn between two solutions ... should we have `is_image_fbc` or should we just have the FBC specific LABEL as a constant and use  `get_image_label` directly (on all places where `is_image_fbc` is)

iib/workers/tasks/fbc_utils.py
lipoja: Another quick solution is to move these two functions directly to `build.py`. 
Personally I am against it because then we would mix FBC related functions with rest of  a code. Another thing is that `build.py` is already pretty big.

chandwanitulsi: FWIW, +1 to keeping fbc related functions in a different file and not bloating up build.py

iib/workers/tasks/fbc_utils.py
lipoja: yes

iib/workers/tasks/fbc_utils.py
pbortlov: Maybe we should rather create new files for add/rm functions.. But I am fine

iib/workers/tasks/utils.py
lipoja: can we resolve this one? It is no longer private function

iib/workers/tasks/fbc_utils.py
```diff
@@ -17,6 +19,45 @@ def is_image_fbc(image):
...
+    from iib.workers.tasks.utils import get_image_label
+
+    fbc_dir_path = get_image_label(from_index, 'operators.operatorframework.io.index.configs.v1')
+    if not fbc_dir_path:
+        raise IIBError('Index image doesn\'t have the label specifying its catalog location.')
+    _copy_files_from_image(from_index, fbc_dir_path, base_dir)
+    return os.path.join(base_dir, os.path.dirname(fbc_dir_path))
+
+
+def get_hidden_index_database(from_index, base_dir):
```
pbortlov, line 49: Would it be possible to merge functions **get_hidden_index_database** and **get_index_fbc_dir** then seem to have similar use case. Copy and join the path are used in both of them.

iib/workers/tasks/fbc_utils.py
```diff
@@ -17,6 +19,45 @@ def is_image_fbc(image):
...
+    """
+    Get hidden database file from the specified index image and save it locally.
+
+    :param str from_index: index image to get database file from.
+    :param str base_dir: base directory to which the database file should be saved.
+    :return: path to the copied database file.
+    :rtype: str
+    :raises IIBError: if any podman command fails.
```
pbortlov, line 57: Is it used in this function?

iib/workers/tasks/fbc_utils.py
lipoja: IIRC: Those FBC functions should be temporary - this is solution for Phase 1. 
We might remove them in a future, I would not mix that with our code too much.

iib/workers/tasks/utils.py
```diff
@@ -175,28 +176,49 @@ def _requires_max_ocp_version(bundle):
...
+    if from_index:
+        if not is_image_fbc(from_index):
+            db_path = _get_index_database(from_index, base_dir)
+        else:
+            fbc_dir = get_index_fbc_dir(from_index, base_dir)
+
     for port in range(port_start, port_end):
+        if fbc_dir:
+            cmd = ['opm', 'serve', fbc_dir, '-p', str(port), '-t', '/dev/null']
+            cwd = os.path.abspath(os.path.join(fbc_dir, os.path.pardir))
```
pbortlov, line 213: What does this variable mean? 'cwd'? Could you add a comment or change the name?

iib/workers/tasks/fbc_utils.py
lipoja: I do not see a nice way how to merge it. Do you have something specific in mind?

iib/workers/tasks/fbc_utils.py
lipoja: Thanks - removed

iib/workers/tasks/utils.py
lipoja: cwd - current working directory - it is kwarg for https://docs.python.org/3/library/subprocess.html#subprocess.run 
That is why I would prefer to stick with this name.

iib/workers/tasks/fbc_utils.py
pbortlov: Could it be a part of the code where you call the get_hidden_index_database?

JAVGan: IIUC `get_index_fbc_dir` returns the catalog path for FBC images and `get_hidden_index_database` returns the DB path for legacy images.

IMO it's possible to merge them in a single function, whatever the name it is by using a single and generic *whatever_path* variable, more or less like this:

1. Call `is_image_fbc(from_index)` to verify if the image is FBC or not
2. **If** FBC do try to get `whatever_path` using `get_image_label(from_index,  'operators.operatorframework.io.index.configs.v1')` and verify if it's set, raising IIBError if not, as it's implemented; **else**  try to get path from `get_worker_config().get('hidden_index_db_path')` and verify if it's set or raise `IIBError` due to misconfiguration
3. Once you have your generic `whatever_path` you use `_copy_files_from_image` replacing `fcb_dir_path` and `db_path` with this generic one obtained in step 2
4. Finally return the `os.path.join(base_dir, os.path.basename(whatever_path))`

**Note:** The variable `whatever_path` should have a nice name, so here I'm just using this name to exemplify how I imagine these 2 functions can be merged

iib/workers/tasks/utils.py
JAVGan: I believe using `is_image_fbc` keeps the code more readable and decoupled. 

It's better to keep the "intelligence" in `is_image_fbc` than using a constant IMO since today it's a simple label, one day it can be something more complex than that and the effort to change a lot of functions will be greater than simply changing `is_image_fbc` behavior.

iib/workers/tasks/fbc_utils.py
JAVGan: Also I don't see the function `get_hidden_index_database` being used anywhere in this PR... So I believe it can be merged, but if you do that please rename `get_index_fbc_dir` to something more generic, like `get_index_catalog_dir` and remember to change its calls on `iib.workers.tasks.utils` to match the new name

iib/workers/tasks/utils.py
```diff
@@ -175,28 +176,49 @@ def _requires_max_ocp_version(bundle):
...
 
+    assert (
+        len([i for i in (from_index, fbc_dir, db_path) if i]) == 1
+    ), 'Only one keyword parameter can be set'
+
+    if from_index:
+        if not is_image_fbc(from_index):
+            db_path = _get_index_database(from_index, base_dir)
+        else:
+            fbc_dir = get_index_fbc_dir(from_index, base_dir)
```
JAVGan, line 208: In case of merging the `get_index_fbc_dir` with `get_hidden_index_database` you should change its name here.

Also I still don't understand the use of `get_hidden_index_database` since I thought it would be used here but instead you're using `_get_index_database`.

Both `iib.workers.tasks.fbc_utils.get_hidden_index_database` and `iib.workers.tasks.build._get_index_database` seems redundant to me. I know they're getting the SQLite from different locations, the first one from hidden path set in configuration and the last one from the image's label, but do we really need that?

It seems we're replicating the very same functions over different places.

iib/workers/tasks/fbc_utils.py
chandwanitulsi: Are we going to have this function or it would be removed?

iib/workers/tasks/fbc_utils.py
JAVGan: @lipoja please let me know the use `get_hidden_index_dabase` since I don't see calling for it anywhere and also whether you consider the merge with `get_index_fbc_dir`

iib/workers/tasks/utils.py
lipoja: yeah, lets agree on keeping `is_image_fbc` for now. however I will modify it to use `get_image_label` inside. In different PR.

iib/workers/tasks/fbc_utils.py
lipoja: @pbortlov @JAVGan Personally it does not feel right to me to merge them. I wanted to have simple way how to get hidden index.db (and path to it) and also similar to get path to file-based catalog directory copied from index image. 
If I merge it then the function is more complicated and the caller has to specify what should be the output - if you want index.db or catalog directory. 

And also please have a look now - I was fixing the code and both functions have now common only calling `_copy_files_from_image` . The rest is changed. Therefore I would prefer to keep them separate.

iib/workers/config.py
```diff
@@ -46,6 +46,8 @@ class Config(object):
         'iib.workers.tasks.build_create_empty_index',
         'iib.workers.tasks.general',
     ]
+    # Path to hidden location of SQLite database
+    hidden_index_db_path = '/_hidden/do.not.edit.db'
```
lipoja, line 50: @gallettilance I modified the path for index.db based discussion on yesterdays meeting. To be explicit that this should not be edited manually. Is this sufficient? Or would you suggest different name/location... ?
Thank you!

JAVGan (dismissed):

iib/workers/tasks/utils.py
lipoja: @JAVGan `get_hidden_index_database` will be used only for File-Based Catalog images and since we have `if` on line above saying that the image is not FBC, then we should get the index.db from usual location.

Both are getting same SQLite database. Only for FBC image it should be hidden and used only for Phase 1 of this "project" we have to use this hidden database internally in IIB. We are expecting to remove behavior this in future.

iib/workers/config.py
```diff
@@ -16,7 +16,7 @@ class Config(object):
         os.path.expanduser('~'), '.docker', 'config.json.template'
     )
     iib_greenwave_url = None
-    iib_grpc_init_wait_time = 3
+    iib_grpc_init_wait_time = 10
```
pbortlov, line 19: Could you share why you changed the waiting time?

iib/workers/tasks/fbc_utils.py
```diff
@@ -1,7 +1,12 @@
 # SPDX-License-Identifier: GPL-3.0-or-later
 # This file contains functions that are common for File-Based Catalog image type
+import os
 
-from iib.workers.tasks.utils import skopeo_inspect
+from iib.exceptions import IIBError
+from iib.workers.config import get_worker_config
+
+# path to index.db in our temp directories used in IIB code
+TEMP_INDEX_DB_PATH = 'database/index.db'
```
pbortlov, line 9: would it be better to add it to config.py?

iib/workers/tasks/fbc_utils.py
pbortlov: thanks Jan for the work, explanation and changes :)

iib/workers/tasks/utils.py
pbortlov: I know I bother you so much with this but.. Could you create a task or PR and share it here? I worry this will get lost. Feel free to close this conversation afterward.

lipoja: https://github.com/release-engineering/iib/pull/328

iib/workers/tasks/fbc_utils.py
lipoja: I was thinking about it.

lipoja: I guess we can put it there but I would like to have some place more as constants then "config". So I was not sure if to put it there or not. This should not be changed (configured). That is why I put it here...

chandwanitulsi: we are having same path in https://github.com/release-engineering/iib/pull/323/files#diff-1bece9798c37acfe1ca705657d12580325f65a8edd965d6c3d004c7fa3237b40R74, could we reuse the variable? in that case we would like to have it a common location

iib/workers/config.py
yashvardhannanavati: Should this probably reside under /var/lib/iib/_hidden/do.not.edit.db?

iib/workers/tasks/fbc_utils.py
yashvardhannanavati: @lipoja Can I please ask why do you not want this to be configured?

iib/workers/tasks/fbc_utils.py
```diff
@@ -17,6 +22,46 @@ def is_image_fbc(image):
...
+    :return: path to the copied file-based catalog directory.
+    :rtype: str
+    :raises IIBError: if any podman command fails.
+    """
+    from iib.workers.tasks.build import _copy_files_from_image
+    from iib.workers.tasks.utils import get_image_label
+
+    fbc_dir_path = get_image_label(from_index, 'operators.operatorframework.io.index.configs.v1')
+    if not fbc_dir_path:
+        raise IIBError('Index image doesn\'t have the label specifying its catalog location.')
```
yashvardhannanavati, line 47: Could you please log the label in the error message to make the error less ambiguous?

iib/workers/tasks/utils.py
```diff
@@ -175,28 +176,49 @@ def _requires_max_ocp_version(bundle):
...
     :raises IIBError: if all tried ports are in use, or the command failed for another reason.
     """
+    from iib.workers.tasks.build import _get_index_database
+
     conf = get_worker_config()
     port_start = conf['iib_grpc_start_port']
     port_end = port_start + conf['iib_grpc_max_port_tries']
 
+    assert (
+        len([i for i in (from_index, fbc_dir, db_path) if i]) == 1
```
yashvardhannanavati, line 201: nitpick: Could you please use a variable name instead of i?

iib/workers/tasks/utils.py
```diff
@@ -175,28 +176,49 @@ def _requires_max_ocp_version(bundle):
...
+    ), 'Only one keyword parameter can be set'
```
yashvardhannanavati, line 202: Could you please change this to mention the parameter? Something like `Exactly one of x,y and z should be set`

iib/workers/tasks/utils.py
yashvardhannanavati, line 179: Since this has two code paths, I would recommend having two separate functions instead of clubbing them into one. That way it would also be easier to get rid of the old code once it becomes obsolete. 

You can probably make another helper for `opm serve/ opm registry serve` which will run the command based on a flag being passed.

yashvardhannanavati: Also, if you do create the helper for that opm command, I would recommend putting it in a new module `opm_operations.py`. It might be a nice start to code cleaning. But I would leave this as optional

yashvardhannanavati (changes requested):

iib/workers/config.py
lipoja: I agree, this is better path. Thanks Yash!

iib/workers/tasks/fbc_utils.py
lipoja: @chandwanitulsi The path should be reused already! That is why I created the constant.  :man_facepalming: and I see that I have that commit in the other PR. I will move it here. Thanks!

@yashvardhannanavati From what I see it is more constant than something to be allowed for user to change. It is used only as path where index.db is copied to in temp folder for request. I just need a place where this path can be set and used from multiple places.

iib/workers/tasks/fbc_utils.py
lipoja: Thanks!

iib/workers/tasks/utils.py
lipoja: @yashvardhannanavati The Nitpick Strikes Back: but `i` is variable name. It is variable which I named "i".

iib/workers/tasks/utils.py
lipoja: Better, thanks!

iib/workers/tasks/utils.py
lipoja: Let me deeply thing about better solution. :thinking:

iib/workers/config.py
pbortlov: thanks

iib/workers/tasks/utils.py
chandwanitulsi: so currently we don't have any use-case where fbc_dir parameter is used, is that for future cases? In that case should we add it here already?

iib/workers/config.py
```diff
@@ -46,6 +46,8 @@ class Config(object):
         'iib.workers.tasks.build_create_empty_index',
         'iib.workers.tasks.general',
     ]
+    # Path to hidden location of SQLite database
+    hidden_index_db_path = '/var/lib/iib/_hidden/do.not.edit.db'
```
lipoja, line 50: @gallettilance updated based on @yashvardhannanavati suggestion which I liked more. 
@gallettilance do you agree with this path or do you have different idea where to store the "hidden" index.db

iib/workers/tasks/utils.py
lipoja: I agree that we are not using it now, we can remove `fbc_dir`

lipoja: @yashvardhannanavati Since I liked your ideas I did them... good luck with code-review :smile:

iib/workers/tasks/fbc_utils.py
```diff
@@ -17,6 +25,49 @@ def is_image_fbc(image):
...
+    from iib.workers.tasks.utils import skopeo_inspect
+
     skopeo_output = skopeo_inspect(f'docker://{image}')
-    dc_image_label = 'operators.operatorframework.io.index.configs.v1'
-    return dc_image_label in skopeo_output['Labels']
+    fbc_image_label = 'operators.operatorframework.io.index.configs.v1'
+    return fbc_image_label in skopeo_output['Labels']
+
+
+def get_index_fbc_dir(from_index, base_dir):
```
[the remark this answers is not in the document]
lipoja: `get_catalog_dir()` it is

lipoja:
@yashvardhannanavati @chandwanitulsi @JAVGan @pbortlov 
Could you go through the code again. I liked Yash's idea and I move some code to separate file. I also split serve_registry functions and addressed some issues. Right now it is as one additional commit because I do not have time to do it properly. 
I plan to split it to smaller commits on Monday.

iib/workers/tasks/opm_operations.py
```diff
@@ -0,0 +1,169 @@
...
+    """
+    Locally start OPM registry service, which can be communicated with using gRPC queries.
+
+    Due to IIB's paralellism, the service can run multiple times, which could lead to port
+    binding conflicts. Resolution of port conflicts is handled in this function as well.
+
+    :param str db_path: path to index database containing the registry data.
+    :return: tuple containing port number of the running service and the running Popen object.
+    :rtype: (int, Popen)
+    :raises IIBError: if all tried ports are in use, or the command failed for another reason.
```
pbortlov, line 100: Is this function raises IIBError or _serve_cmd_at_port()

iib/workers/tasks/opm_operations.py
```diff
@@ -0,0 +1,169 @@
...
+
+    if not is_image_fbc(from_index):
+        db_path = _get_index_database(from_index, base_dir)
+        return opm_registry_serve(db_path)
+
+    catalog_dir = get_catalog_dir(from_index, base_dir)
+    return opm_serve(catalog_dir)
+
+
+def opm_serve(catalog_dir):
```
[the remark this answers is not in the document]
pbortlov: This function seems to be almost the same with opm_registry_serve. Can we distinguish by name if this is used for  FBC?

pbortlov (changes requested):

iib/workers/config.py
```diff
@@ -140,6 +144,7 @@ class TestingConfig(DevelopmentConfig):
     iib_request_related_bundles_dir = None
     # disable dogpile cache for tests
     iib_dogpile_backend = 'dogpile.cache.null'
+    iib_grpc_init_wait_time = 3
```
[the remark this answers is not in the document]
lipoja: Test were failing with the 10 - so it was faster to fix them here then modify all the tests.

iib/workers/tasks/opm_operations.py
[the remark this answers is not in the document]
lipoja: ad naming: I was thinking about it however I went with this because it directly reflects what is called `opm serve` and `opm registry serve`. This is how it is distinguish in `opm`, should we do it differently? 
Is not doc-string description of the function sufficient?

Thanks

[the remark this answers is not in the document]
pbortlov: sick day?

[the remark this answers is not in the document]
lipoja: Yes. But we shod deal with this PR asap :) 
It is blocking future work.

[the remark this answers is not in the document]
JAVGan: The functions `opm_serve` and `opm_registry_serve` are indeed quite similar, the only differences are:

1. Parameter `catalog_dir` vs `db_path`
2. Inside function: `cmd` and `cwd` content differs

And that's it, the rest of the functions are simply copy+paste

One suggestion to avoid this would be creating something like that (if you think that's useful, otherwise I'm open to suggestions):

```{python}
def opm_serve(base_dir, is_fbc=True):
    [...]
    if is_fbc:
        cmd = ['opm', 'serve', catalog_dir, '-p', str(port), '-t', '/dev/null']
        cwd = os.path.abspath(os.path.join(catalog_dir, os.path.pardir))
    else:
        cmd = ['opm', 'registry', 'serve', '-p', str(port), '-d', db_path, '-t', '/dev/null']
        cwd = os.path.dirname(db_path)
    [...]

```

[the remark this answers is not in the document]
lipoja: I had it before however Yash suggested to split it to two because they are used to different things.
What you are suggesting leads to having different inputs for one function and I do like this separation better.

iib/workers/tasks/opm_operations.py
lipoja: Removed, thanks

iib/workers/tasks/opm_operations.py
```diff
@@ -0,0 +1,169 @@
...
+    cwd = os.path.abspath(os.path.join(catalog_dir, os.path.pardir))
+    return (
+        port,
+        _serve_cmd_at_port(
+            cmd, cwd, port, conf['iib_grpc_max_tries'], conf['iib_grpc_init_wait_time']
+        ),
+    )
+
+
+def opm_registry_serve(db_path):
```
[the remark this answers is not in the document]
lipoja: added

iib/workers/tasks/opm_operations.py
```diff
@@ -0,0 +1,169 @@
...
+    """
+    Locally start OPM service, which can be communicated with using gRPC queries.
+
+    Due to IIB's paralellism, the service can run multiple times, which could lead to port
+    binding conflicts. Resolution of port conflicts is handled in this function as well.
+
+    :param str catalog_dir: path to file-based catalog directory that should be served.
+    :return: tuple containing port number of the running service and the running Popen object.
+    :rtype: (int, Popen)
+    :raises IIBError: if all tried ports are in use, or the command failed for another reason.
```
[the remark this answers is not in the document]
lipoja: I prefer to remove it, since it is not raised directly. 
Removed

iib/workers/tasks/opm_operations.py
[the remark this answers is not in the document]
lipoja: I updated the code and simplified it a bit, I still prefer to not mix those functions but the duplicate code was separated.

iib/workers/tasks/opm_operations.py
```diff
@@ -0,0 +1,169 @@
...
+    Locally start OPM registry service, which can be communicated with using gRPC queries.
+
+    Due to IIB's paralellism, the service can run multiple times, which could lead to port
+    binding conflicts. Resolution of port conflicts is handled in this function as well.
+
+    :param str base_dir: base directory to create temporary files in.
+    :param str from_index: index image to inspect.
+    :return: tuple containing port number of the running service and the running Popen object.
+    :rtype: (int, Popen)
+    :raises IIBError: if all tried ports are in use, or the command failed for another reason.
```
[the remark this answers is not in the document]
lipoja: Yeah, I forgot that. Thanks!
Moved

iib/workers/tasks/opm_operations.py
```diff
@@ -0,0 +1,169 @@
...
+            log.info('Port %d is in use, trying another.', port)
+            port += 1
+        sock.close()
+
+    err_msg = f'No free port has been found after {port_end - port_start} attempts.'
+    log.error(err_msg)
+    raise IIBError(err_msg)
+
+
+def opm_serve_from_index(base_dir, from_index):
```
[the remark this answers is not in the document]
lipoja: Added

iib/workers/tasks/opm_operations.py
```diff
@@ -0,0 +1,169 @@
...
+
+log = logging.getLogger(__name__)
+
+
+def _get_free_port(port_start, port_end):
+    """
+    Return free port that is safe to use for opm command.
+
+    :param int port_start: port from which we should start trying to connect to
+    :param int port_end: port on which we should stop trying
```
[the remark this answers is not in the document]
lipoja: :facepalm: Thanks!

iib/workers/tasks/opm_operations.py
```diff
@@ -0,0 +1,169 @@
...
+from retry import retry
+
+from iib.exceptions import AddressAlreadyInUse, IIBError
+from iib.workers.config import get_worker_config
+from iib.workers.tasks.fbc_utils import is_image_fbc, get_catalog_dir
+
+log = logging.getLogger(__name__)
+
+
+def _get_free_port(port_start, port_end):
```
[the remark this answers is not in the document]
lipoja: added

iib/workers/tasks/fbc_utils.py
```diff
@@ -17,6 +23,49 @@ def is_image_fbc(image):
...
+    if not fbc_dir_path:
+        error_msg = f'Index image {from_index} does not contain file-based catalog.'
+        log.error(error_msg)
+        raise IIBError(error_msg)
+
+    _copy_files_from_image(from_index, fbc_dir_path, base_dir)
+    return os.path.join(base_dir, os.path.basename(fbc_dir_path))
+
+
+def get_hidden_index_database(from_index, base_dir):
```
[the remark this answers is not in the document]
lipoja: added

iib/workers/tasks/fbc_utils.py
```diff
@@ -17,6 +23,49 @@ def is_image_fbc(image):
...
+
+    :param str from_index: index image to get file-based catalog directory from.
+    :param str base_dir: base directory to which the database file should be saved.
+    :return: path to the copied file-based catalog directory.
+    :rtype: str
+    :raises IIBError: if any podman command fails.
+    """
+    from iib.workers.tasks.build import _copy_files_from_image
+    from iib.workers.tasks.utils import get_image_label
+
```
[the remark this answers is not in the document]
lipoja: added

iib/workers/tasks/fbc_utils.py
```diff
@@ -17,6 +23,49 @@ def is_image_fbc(image):
...
+    from iib.workers.tasks.utils import skopeo_inspect
+
     skopeo_output = skopeo_inspect(f'docker://{image}')
-    dc_image_label = 'operators.operatorframework.io.index.configs.v1'
-    return dc_image_label in skopeo_output['Labels']
+    fbc_image_label = 'operators.operatorframework.io.index.configs.v1'
+    return fbc_image_label in skopeo_output['Labels']
+
+
+def get_catalog_dir(from_index, base_dir):
```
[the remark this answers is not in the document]
lipoja: added

iib/workers/config.py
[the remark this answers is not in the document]
lipoja: change removed

iib/workers/tasks/opm_operations.py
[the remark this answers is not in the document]
pbortlov: @JAVGan please check.. I am ok with the changes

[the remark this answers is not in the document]
JAVGan: Sure, I got the point, I'm ok with it too.

JAVGan (dismissed):

iib/workers/tasks/opm_operations.py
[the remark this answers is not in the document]
pbortlov: Closing

pbortlov:
@release-engineering/exd-guild-hello-operator PLEASE final review

pbortlov (approved):

yashvardhannanavati (dismissed):
@lipoja This looks rad! Thank you for making the changes. Approved!

tests/test_workers/test_tasks/test_build_merge_index_image.py
```diff
@@ -57,7 +57,7 @@ def test_handle_merge_request(
     mock_pi,
     mock_vii,
     mock_uiips,
-    mock_sir,
+    mock_sr,
```
chandwanitulsi, line 60: Trivial: mock_ors instead of mock_sr (I assume mock_sr was used with previous patch but the function changed later)

tests/test_workers/test_tasks/test_opm_operations.py
```diff
@@ -0,0 +1,157 @@
...
+    mock_socket.socket().bind.side_effect = OSError()
+    with pytest.raises(IIBError, match='No free port has been found after 3 attempts.'):
+        opm_operations._get_free_port(50051, 50054)
+
+
+@mock.patch('time.sleep')
+@mock.patch('subprocess.Popen')
+@mock.patch('iib.workers.tasks.opm_operations.socket')
+@mock.patch('iib.workers.tasks.utils.run_cmd')
+def test_opm_serve_registry(mock_run_cmd, mock_socket, mock_popen, mock_sleep):
```
chandwanitulsi, line 41: Trivial: test_opm_registry_serve?

chandwanitulsi (approved):
Added few trivial suggestions, but that shouldn't block the PR from merging

chandwanitulsi (approved):

PR #324, opened 2021-11-05
Updating werkzeug to 2.0.2

lipoja:
Update dependency to satisfy `safety` check.


lipoja:
This is going to require more work. Will file a ticket for it. :D as Yash said

PR #327, opened 2021-11-11
Avoid failing to create-empty-index when labels are not set

JAVGan:
This commit avoids raising exception on create-empty-index if the parameter "labels" are not set in request.

Since "labels" are not a mandatory parameter the request shouldn't break.

This is a contribution out of any JIRA cards.


lipoja (approved):

pbortlov (approved):

chandwanitulsi (approved):

pbortlov:
@JAVGan could you rebase?

PR #328, opened 2021-11-16
Using get_image_label instead of skopeo_inspect for is_image_fbc

lipoja:
- supports logging
- better caching - since we might call get labels multiple times


yashvardhannanavati (dismissed):

chandwanitulsi (approved):

lipoja:
rebased

PR #330, opened 2021-11-22
Fix comparing bundle versions with ocp versions [CLOUDDST-10476]

pbortlov:
- modify version comparison
- add packaging library
- add debug comments

### DIFF
--- a/iib/workers/tasks/build_merge_index_image.py
+++ b/iib/workers/tasks/build_merge_index_image.py
@@ -4,6 +4,8 @@ import logging
 import stat
 import tempfile
 
+from packaging.version import Version
+
 from iib.exceptions import IIBError
 from iib.workers.api_utils import set_request_state
 from iib.workers.tasks.build import (
@@ -308,33 +310,36 @@ def is_bundle_version_valid(bundle_path, valid_ocp_version):
     :return: a boolean indicating if the bundle_path satisfies the index ocp_version
     :rtype: bool
 
-           |  "v4.5"   |   "=v4.6"    | "v4.5-v4.7" | "v4.5,v4.6"  | "v4.6,v4.5"
+           |  "v4.5"   |   "=v4.6"    | "v4.5-v4.7"  | "v4.5,v4.6" | "v4.6,v4.5"
     -------------------------------------------------------------------------------
-    v4.5   | included  | NOT included |  included   |  included    |  NOT included
+    v4.5   | included  | NOT included |   included   |  included   |  NOT included
     -------------------------------------------------------------------------------
-    v4.6   | included  |   included   |  included   |  included    |  included
+    v4.6   | included  |   included   |   included   |  included   |  included
     -------------------------------------------------------------------------------
-    v4.7   | included  | NOT included |  included   |  included    |  included
+    v4.7   | included  | NOT included |   included   |  included   |  included
+    -------------------------------------------------------------------------------
+    v4.8   | included  | NOT included | NOT included |  included   |  included
     """
     try:
-        float_valid_ocp_version = float(valid_ocp_version.replace('v', ''))
+        ocp_version = Version(valid_ocp_version.replace('v', ''))
     except ValueError:
         raise IIBError(f'Invalid OCP version, "{valid_ocp_version}", specified in Index Image')
     try:
         bundle_version_label = get_image_label(bundle_path, 'com.redhat.openshift.versions')
         bundle_version = bundle_version_label.replace('v', '')
+        log.debug(f'Bundle version {bundle_version}, Index image version {valid_ocp_version}')
         if bundle_version_label.startswith('='):
-            if float(bundle_version.strip('=')) == float_valid_ocp_version:
+            if Version(bundle_version.strip('=')) == ocp_version:
                 return True
         elif '-' in bundle_version_label:
-            min_version, max_version = [float(version) for version in bundle_version.split('-')]
-            if min_version <= float_valid_ocp_version <= max_version:
+            min_version, max_version = [Version(version) for version in bundle_version.split('-')]
+            if min_version <= ocp_version <= max_version:
                 return True
         elif "," in bundle_version_label:
-            versions = [float(version) for version in bundle_version.split(",")]
-            if float_valid_ocp_version >= versions[0]:
+            versions = [Version(version) for version in bundle_version.split(",")]
+            if versions[0] <= ocp_version:
                 return True
-        elif float_valid_ocp_version >= float(bundle_version):
+        elif Version(bundle_version) <= ocp_version:
             return True
     except ValueError:
         log.warning(
--- a/iib/workers/tasks/utils.py
+++ b/iib/workers/tasks/utils.py
@@ -915,6 +915,7 @@ def get_all_index_images_info(build_request_config, index_version_map):
     """
     infos = {}
     for (index, version) in index_version_map:
+        log.debug(f'Get index image info {index} for version {version}')
         if not hasattr(build_request_config, index):
             from_index = None
         else:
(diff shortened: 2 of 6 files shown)


iib/workers/tasks/build_merge_index_image.py
```diff
@@ -200,8 +200,8 @@ def handle_merge_request(
             raise IIBError(err_msg)
 
         set_request_state(request_id, 'in_progress', 'Getting bundles present in the index images')
-        log.info('Getting bundles present in the source index image')
         with set_registry_token(overwrite_target_index_token, source_from_index):
+            log.info('Getting bundles present in the source index image')
```
yashvardhannanavati, line 204: Is this change required?

iib/workers/tasks/build_merge_index_image.py
```diff
@@ -344,3 +345,13 @@ def is_bundle_version_valid(bundle_path, valid_ocp_version):
         )
 
     return False
+
+def correct_float_precision(version):
```
yashvardhannanavati, line 349: Maybe rename it to simply `float_precision` and have the precision as a parameter which defaults to 2?

yashvardhannanavati (changes requested):

iib/workers/tasks/build_merge_index_image.py
```diff
@@ -317,24 +317,25 @@ def is_bundle_version_valid(bundle_path, valid_ocp_version):
...
+            min_version, max_version = [correct_float_precision(version) for version in bundle_version.split('-')]
             if min_version <= float_valid_ocp_version <= max_version:
                 return True
         elif "," in bundle_version_label:
-            versions = [float(version) for version in bundle_version.split(",")]
+            versions = [correct_float_precision(version) for version in bundle_version.split(",")]
             if float_valid_ocp_version >= versions[0]:
                 return True
-        elif float_valid_ocp_version >= float(bundle_version):
+        elif float_valid_ocp_version >= correct_float_precision(bundle_version):
```
lcarva, line 338: I don't think this will work as expected:

```python
>>> '4.10' > '4.9'
False
```

iib/workers/tasks/build_merge_index_image.py
```diff
@@ -317,24 +317,25 @@ def is_bundle_version_valid(bundle_path, valid_ocp_version):
     v4.7   | included  | NOT included |  included   |  included    |  included
     """
```
lcarva, line 318: Let's update the unit tests

iib/workers/tasks/build_merge_index_image.py
pbortlov: sure

iib/workers/tasks/build_merge_index_image.py
lcarva: I think you'll need something like this: https://gist.github.com/limed/473a498641bbc7761a20

pbortlov: I see. We can't compare those numbers

iib/workers/tasks/build_merge_index_image.py
```diff
@@ -344,3 +357,15 @@ def is_bundle_version_valid(bundle_path, valid_ocp_version):
...
+def sort_floats(versions):
+    """
+    Sort list of strings with floats
+
+    :param list versions: unsorted list of versions
+    :rtype: list
+    """
+    convert = lambda text: int(text) if text.isdigit() else text
+    alphanum_key = lambda key: [convert(c) for c in re.split('([0-9]+)', key)]
+    versions.sort(key=alphanum_key)
```
lcarva, line 363: This sorts the given list in place. Either use `sorted` instead, or don't return the list.

iib/workers/tasks/build_merge_index_image.py
```diff
@@ -308,33 +309,38 @@ def is_bundle_version_valid(bundle_path, valid_ocp_version):
...
         if bundle_version_label.startswith('='):
-            if float(bundle_version.strip('=')) == float_valid_ocp_version:
+            if bundle_version.strip('=') == ocp_version:
                 return True
         elif '-' in bundle_version_label:
-            min_version, max_version = [float(version) for version in bundle_version.split('-')]
-            if min_version <= float_valid_ocp_version <= max_version:
+            min_version, max_version = [version for version in bundle_version.split('-')]
+
+            if sort_floats(list(min_version + ocp_version)) == list(min_version + ocp_version) and \
```
lcarva, line 336: The usage of lists seems unnatural here. Consider this. Instead of using `sort_floats`, introduce a new function, e.g. `to_version_tuple`, which does the conversion that is currently embedded in `sort_floats`, e.g. `'v4.10'` -> `(4, 10)`. Then, you can just compare the tuples as you'd compare any other value. `sort_floats` wouldn't be needed.

iib/workers/tasks/build_merge_index_image.py
yashvardhannanavati: Probably should be a separate commit. fwiw,  I think it wouldn't matter if we bring it 2 lines below or keep it as is :)

yashvardhannanavati (changes requested):

iib/workers/tasks/build_merge_index_image.py
pbortlov: what do you think about this? https://codereview.stackexchange.com/questions/217992/compare-two-version-numbers-in-the-form-major-minor-patch-build-to-see-if-one-is

iib/workers/tasks/build_merge_index_image.py
pbortlov: removed

iib/workers/tasks/build_merge_index_image.py
lcarva: Ooh I like that! `LooseVersion` should do the trick for us.

```python
>>> from distutils.version import LooseVersion
>>> LooseVersion('4.10') > LooseVersion('4.9')
True
>>> LooseVersion('4.1') > LooseVersion('4.9')
False
>>> LooseVersion('4.10') > LooseVersion('4.101')
False
>>> LooseVersion('4.10') > LooseVersion('4.09')
True
```

pbortlov: IT seems that this already deprecated. We can use this:
>>> from packaging import version
>>> version.parse("2.3.1") < version.parse("10.1.2")
True
>>> version.parse("1.3.a4") < version.parse("10.1.2")
True
>>> isinstance(version.parse("1.3.a4"), version.Version)
True
>>> isinstance(version.parse("1.3.xy123"), version.LegacyVersion)
True
>>> version.Version("1.3.xy123")
Traceback (most recent call last):
...
packaging.version.InvalidVersion: Invalid version: '1.3.xy123'

iib/workers/tasks/build_merge_index_image.py
pbortlov: Done

lcarva (dismissed):

yashvardhannanavati (approved):

iib/workers/tasks/build_merge_index_image.py
```diff
@@ -308,33 +310,36 @@ def is_bundle_version_valid(bundle_path, valid_ocp_version):
...
-            min_version, max_version = [float(version) for version in bundle_version.split('-')]
-            if min_version <= float_valid_ocp_version <= max_version:
+            min_version, max_version = [Version(version) for version in bundle_version.split('-')]
+            if min_version <= ocp_version <= max_version:
                 return True
         elif "," in bundle_version_label:
-            versions = [float(version) for version in bundle_version.split(",")]
-            if float_valid_ocp_version >= versions[0]:
+            versions = [Version(version) for version in bundle_version.split(",")]
+            if versions[0] <= ocp_version:
```
chandwanitulsi, line 340: The comparisons are reversed here, but that shouldn't a problem

chandwanitulsi (approved):

PR #331, opened 2021-11-23
Upgrade IIB to Flask 2.0

JAVGan:
# Upgrade IIB to Flask 2.0

This pull request refers to **CLOUDDST-10176** which goal is to upgrade `Flask` and `Werkzeug` to 2.0 and ensure IIB works as it was before without changing any functional behavior.

## Problems

* When upgrading `Flask` on `py36` it adds a dependency named `dataclasses` but this dependency doesn't exist in further versions of Python, probably [due to this PEP](https://www.python.org/dev/peps/pep-0557/).
* Due to problem below it's impossible to have the same `requirements.txt` file for py36 and next version.
* Python 3.6 goes [EOL December this year](https://endoflife.date/python)

## Solution proposal

@lcarva suggested the following solution (summed up for IIB):

1. Remove py37 from the tests
2. Add py38 to the tests, and create a new set of requirements files that are meant to be used in a py38 environment
3. Deploy and test using this new file
4. On success, remove support for py36 and update `.python-version` to the desired 3.8 release.
5. Rename the requirements*-py38.txt back to remove the py38 suffix. 
6. Deploy again using requirements.txt

For more details please check the ticket.

## This pull request

In this pull request I'm sending 3 commits, so it's possible to cherry-pick them if needed.

1. `Flask` and `Werkzeug` to 2.0 on `py36` and remove `py37`
2. Include `py38` and keep separate files as requirements
3. Use `py38` as default on `Dockerfile-api` and `Dockerfile-workers`

### Observations

1. IIB code didn't had to suffer almost any changes on `pip-compile` but 2 automated tests broke, namely:
    * `tests.test_web.test_messaging.test_get_ssl_domain_files_missing`
    * `tests.test_web.test_messaging.test_get_ssl_domain_cert_config_not_set`
Both tests broke for the same reason: `unittest.mock` can't patch `Flask.current_app` [in new version](https://stackoverflow.com/questions/57947935/mocking-current-app-in-flask-unit-pytest).
2. Tox, build tests and some manual checks passed in my local env.
3. The requirement files with `*-py38.txt` will be renamed back in the future


docker/Dockerfile-api
```diff
@@ -15,18 +15,19 @@ RUN dnf -y install \
...
+RUN update-alternatives --set python3 $(which python3.8)
 COPY . .
 COPY ./docker/iib-httpd.conf /etc/httpd/conf/httpd.conf
 
-# default python3-pip version for rhel8 python3.6 is 9.0.3 and it can't be updated by dnf
+# default python3-pip version for rhel8 python3.8 is 19.3.1 and it can't be updated by dnf
 # we have to update it by pip to version above 21.0.0
 RUN pip3 install --upgrade pip
-RUN pip3 install -r requirements.txt --no-deps --require-hashes
+RUN pip3 install -r requirements-py38.txt --no-deps --require-hashes
```
pbortlov, line 30: Do we have to use pip3

JAVGan: It was using `pip3` before my change and I kept it, the only difference is the requirements file and also that I used `update-alternatives` to ensure all commands goes to python 3.8 instead of the default on RHEL (3.6).
The names `python` and `pip` are generally not set, as you can see [here](https://developers.redhat.com/blog/2019/05/07/what-no-python-in-red-hat-enterprise-linux-8)

JAVGan:
Rebased with Release v4.9.0

tests/test_web/test_messaging.py
```diff
@@ -145,25 +145,25 @@ def test_get_ssl_domain_files_missing(mock_exists, cert_exists, key_exists, ca_e
...
     ssl_domain = messaging._get_ssl_domain()
 
     assert ssl_domain is None
 
 
 @pytest.mark.parametrize('durable', (True, False))
-@mock.patch('iib.web.messaging.current_app')
-def test_json_to_envelope(mock_current_app, durable):
-    mock_current_app.config = {'IIB_MESSAGING_DURABLE': durable}
+def test_json_to_envelope(durable, app):
```
pbortlov, line 165: I am missing the `app` variable. Do you mock it somehwere?

tests/test_web/test_messaging.py
```diff
@@ -145,25 +145,25 @@ def test_get_ssl_domain_files_missing(mock_exists, cert_exists, key_exists, ca_e
     'missing_key', ('IIB_MESSAGING_CERT', 'IIB_MESSAGING_KEY', 'IIB_MESSAGING_CA')
 )
 @mock.patch('iib.web.messaging.os.path.exists')
-@mock.patch('iib.web.messaging.current_app')
-def test_get_ssl_domain_cert_config_not_set(mock_current_app, mock_exists, missing_key):
+def test_get_ssl_domain_cert_config_not_set(mock_exists, missing_key, app):
```
pbortlov, line 148: What about the `app`

pbortlov (changes requested):

tests/test_web/test_messaging.py
JAVGan: `app` is a fixture defined on `tests/conftest.py` using the configuration from `iib.web.config.TestingConfig` which has the following config:

```
    DEBUG = True
    IIB_WORKER_USERNAMES = ['worker@DOMAIN.LOCAL']
    SQLALCHEMY_DATABASE_URI = f'sqlite:///{TEST_DB_FILE}'
    LOGIN_DISABLED = False
``` 

As stated in Observation 1 it's no longer possible to mock [Flask.current_app](https://stackoverflow.com/questions/57947935/mocking-current-app-in-flask-unit-pytest) so instead of doing it I'm using the fixture already made and including the configuration override which was being used in the original mock.

In this particular test case no additional configuration is required for `Flask.app` it just asserts that the returned domain is the expected one, with the `call_count == 3` and the certs/auths are correct.

But since `_get_ssl_domain` function in `iib.web.messaging` retrieves the config from `Flask.current_app.config` it's impossible to call it without a Flask's app set. Thus I'm using the already made fixture.

tests/test_web/test_messaging.py
JAVGan: `app` is a fixture defined on `tests/conftest.py` using the configuration from `iib.web.config.TestingConfig` which has the following config:

```
    DEBUG = True
    IIB_WORKER_USERNAMES = ['worker@DOMAIN.LOCAL']
    SQLALCHEMY_DATABASE_URI = f'sqlite:///{TEST_DB_FILE}'
    LOGIN_DISABLED = False
``` 

As stated in Observation 1 it's no longer possible to mock [Flask.current_app](https://stackoverflow.com/questions/57947935/mocking-current-app-in-flask-unit-pytest) so instead of doing it I'm using the fixture already made and including the configuration override which was being used in the original mock.

In this particular test case it requires the `app.configuration` to have the following key:

```
    {'IIB_MESSAGING_DURABLE': durable}
```
And `durable` is a boolean being parametrized as both `True` and `False`.

Since `IIB_MESSAGING_DURABLE` was not set in fixture's config it's safe to use and even if it was set the parametrize changes is from `True` to `False`, so no problem here.

JAVGan: Just to clarify, `app` here is the fixture in place of the old argument `mock_current_app` which was defined like this:

```
@mock.patch('iib.web.messaging.current_app')
```

This is no longer possible, it throws the following error which happens on `mock.patch` decorator (I even tried to remove everything from test and let `assert True == True` and got the same result):

```
________________________________________________________________________________________________________ test_json_to_envelope[True] _________________________________________________________________________________________________________

args = (), keywargs = {'durable': True}

    @wraps(func)
    def patched(*args, **keywargs):
>       with self.decoration_helper(patched,
                                    args,
                                    keywargs) as (newargs, newkeywargs):

/usr/lib64/python3.8/unittest/mock.py:1322: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
/usr/lib64/python3.8/contextlib.py:113: in __enter__
    return next(self.gen)
/usr/lib64/python3.8/unittest/mock.py:1304: in decoration_helper
    arg = exit_stack.enter_context(patching)
/usr/lib64/python3.8/contextlib.py:425: in enter_context
    result = _cm_type.__enter__(cm)
/usr/lib64/python3.8/unittest/mock.py:1416: in __enter__
    if spec is None and _is_async_obj(original):
/usr/lib64/python3.8/unittest/mock.py:51: in _is_async_obj
    if hasattr(obj, '__func__'):
.tox/py38/lib/python3.8/site-packages/werkzeug/local.py:432: in __get__
    obj = instance._get_current_object()
.tox/py38/lib/python3.8/site-packages/werkzeug/local.py:554: in _get_current_object
    return self.__local()  # type: ignore
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

    def _find_app():
        top = _app_ctx_stack.top
        if top is None:
>           raise RuntimeError(_app_ctx_err_msg)
E           RuntimeError: Working outside of application context.
E           
E           This typically means that you attempted to use functionality that needed
E           to interface with the current application object in some way. To solve
E           this, set up an application context with app.app_context().  See the
E           documentation for more information.

.tox/py38/lib/python3.8/site-packages/flask/globals.py:47: RuntimeError

```

requirements.txt
```diff
@@ -2,7 +2,7 @@
 # This file is autogenerated by pip-compile with python 3.6
 # To update, run:
 #
-#    pip-compile --generate-hashes --output-file=requirements.txt
+#    pip-compile --generate-hashes
```
lipoja, line 5: This is helpful for people who never used pip-compile - could you please run `pip-compile --generate-hashes --output-file=requirements.txt` so we have the text `--output-file=requirements.txt` back here.

JAVGan: Sure!

JAVGan: I've regenerated it and tested with `tox`, everything ok. 

I'm sending the new commit.

pbortlov:
@JAVGan you don't have `packaging` library added to your PR. I think you have to rebase.

pbortlov (changes requested):
run pip-compile again to add `packaging` to requirements

JAVGan:
> run pip-compile again to add `packaging` to requirements

Thanks a lot for your observation! 

I did a backup of all requirements files, rebased my local, ran again the `pip-compile` and checked with `diff` and `tox`, everything seems ok now.

pbortlov:
> > run pip-compile again to add `packaging` to requirements
> 
> Thanks a lot for your observation!
> 
> I did a backup of all requirements files, rebased my local, ran again the `pip-compile` and checked with `diff` and `tox`, everything seems ok now.

Please run iib and test with at least one endpoint /create-empty-index or /add

@JAVGan

JAVGan:
@pbortlov the tests were made and they're available on JIRA card.

tests/test_web/test_messaging.py
pbortlov: Could you add a comment there so it doesn't get lost?

JAVGan: Sure, I'll do it and send the changes in a moment

JAVGan: @pbortlov @yashvardhannanavati I'm sending the files with comments as requested but I noticed that `safety` now complains about `celery` and `kombu` after running `tox`, no changes were made in `requirements*.txt` during this amended commit.

If I should include these dependency upgrades I'll need to retest everything again:
```
+==============================================================================+
|                                                                              |
|                               /$$$$$$            /$$                         |
|                              /$$__  $$          | $$                         |
|           /$$$$$$$  /$$$$$$ | $$  \__//$$$$$$  /$$$$$$   /$$   /$$           |
|          /$$_____/ |____  $$| $$$$   /$$__  $$|_  $$_/  | $$  | $$           |
|         |  $$$$$$   /$$$$$$$| $$_/  | $$$$$$$$  | $$    | $$  | $$           |
|          \____  $$ /$$__  $$| $$    | $$_____/  | $$ /$$| $$  | $$           |
|          /$$$$$$$/|  $$$$$$$| $$    |  $$$$$$$  |  $$$$/|  $$$$$$$           |
|         |_______/  \_______/|__/     \_______/   \___/   \____  $$           |
|                                                          /$$  | $$           |
|                                                         |  $$$$$$/           |
|  by pyup.io                                              \______/            |
|                                                                              |
+==============================================================================+
| REPORT                                                                       |
| checked 48 packages, using free DB (updated once a month)                    |
+============================+===========+==========================+==========+
| package                    | installed | affected                 | ID       |
+============================+===========+==========================+==========+
| celery                     | 4.4.0     | <5.2.0                   | 42498    |
| kombu                      | 4.6.7     | <5.2.1                   | 42497    |
+==============================================================================+
```

Please let me know if you think it's better to leave these packages since they're not part of JIRA card or if I take advantage of `pip-compile upgrade` and include this in my task and add them to this PR.

Thanks,

pbortlov: It would be great to have NO safety comaplains. If you can upgrade :), THank you

tests/test_web/test_messaging.py
lipoja: This approach was already used in `test_send_messages`.
Question for discussion: Do we want to have these comments everywhere where `app` is used? 
Would not be sufficient just to mention it in commit message but without the comments? 
I can see that `app` is used almost everywhere (`test_messaging.py`, `test_api_v1.py` ... ) without this comment.

tests/test_web/test_messaging.py
JAVGan: Well.... Now we have another problem.

I'm doing `pip-compile upgrade` for `celery`and `kombu` and they got the correct versions on `py38` only. 

On `py36` they get stuck at `celery==5.1.2` and `komu==5.1.0` which are both `< 5.2` and `security` complains.

According to this the changelogs of [celery](https://docs.celeryproject.org/en/stable/changelog.html) and [kombu](https://docs.celeryproject.org/projects/kombu/en/stable/changelog.html) the support for `py36` is dropped on 5.2, just `CTRL+F` for `drop python` to see.

I'll run the `build/run` tests again and send the amended commit but I believe we will need to drop `py36` support definitively to fix Travis as it goes [EOL this month](https://endoflife.date/python).

@yashvardhannanavati @lcarva what do you think?

JAVGan:
I've upgraded `celery` and `kombu`, tested everything and sent the amended commit.

Note that for `py36` both packages are below the version required by `security` since they dropped support for it.

Build and run tests can be found on JIRA.

tests/test_web/test_messaging.py
yashvardhannanavati: +1 on dropping py36 support. But we will have to do this carefully

yashvardhannanavati:
@JAVGan I agree, we should drop py36 support completely and only have py38

@lcarva thoughts?

tests/test_web/test_messaging.py
lipoja: But is not that too many changes for one PR? 
We already have everything prepared for running Flask 2.0 in py3.6 right @JAVGan?

Should not we merge this, test it and then move to py38. I know it will prolong the failing CI, but is in it better to split this task?

lipoja:
@yashvardhannanavati @lcarva I agree with dropping py36 support. However in different PR, if possible. 
We can have py38 or even py39 as default (both are available in RHEL `rhel-8-for-x86_64-appstream-rpms` repo).

tests/test_web/test_messaging.py
JAVGan: @lipoja yes, everything is prepared to run Flask 2.0 on both `py36` and `py38`, just the CI will fail while we keep the `py36` supported.

JAVGan:
Well.... now `py36` fails on CI due to security even using the Python 3.6 on its environment.

I believe there will be no solution for the CI fix while we keep `py36` supported and I agree with @lipoja to do it in a different PR

tests/test_web/test_messaging.py
pbortlov: I will leave it up to you. I haven't been a part of the developer team since the IIB started so I couldn't comment on this. I would like to approach this as an improvement of code documentation for the newcommers

pbortlov:
@lcarva @yashvardhannanavati @release-engineering/exd-guild-hello-operator COuld we make a decision if we keep py36 tests here?

lipoja:
> @lcarva @yashvardhannanavati @release-engineering/exd-guild-hello-operator COuld we make a decision if we keep py36 tests here?

We keep them, since we do not remove py36 support in this PR.

pbortlov:
> > @lcarva @yashvardhannanavati @release-engineering/exd-guild-hello-operator COuld we make a decision if we keep py36 tests here?
> 
> We keep them, since we do not remove py36 support in this PR.

If we don't remove it in this PR when the test will be fixed? We progressed but we didn't fix the tests :(

lipoja:
> > > @lcarva @yashvardhannanavati @release-engineering/exd-guild-hello-operator COuld we make a decision if we keep py36 tests here?
> > 
> > 
> > We keep them, since we do not remove py36 support in this PR.
> 
> If we don't remove it in this PR when the test will be fixed? We progressed but we didn't fix the tests :(

Yes, that is sad life of IIB. PRs are hanging here for too long :(

The reason why not to remove tests here: We are still running IIB on py3.6 in PROD. I do not like the idea to not have IIB tested on Python version which is running on in production.

pbortlov (approved):

yashvardhannanavati (dismissed):
Awesome!

yashvardhannanavati:
@JAVGan This probably needs a rebase?

JAVGan:
Rebased with current master

pbortlov (approved):

yashvardhannanavati (approved):

docker/Dockerfile-api
```diff
@@ -16,18 +16,19 @@ RUN dnf -y install \
...
     && dnf update -y \
     && dnf clean all
+RUN update-alternatives --set python3 $(which python3.8)
 COPY . .
 COPY ./docker/iib-httpd.conf /etc/httpd/conf/httpd.conf
 
-# default python3-pip version for rhel8 python3.6 is 9.0.3 and it can't be updated by dnf
+# default python3-pip version for rhel8 python3.8 is 19.3.1 and it can't be updated by dnf
 # we have to update it by pip to version above 21.0.0
 RUN pip3 install --upgrade pip
```
lcarva, line 30: We should consider removing this. The latest version installed by dnf might sufficient now that it's using python 3.8. Same on `Dockerfile-workers`.

(Apologies for the late review)

JAVGan: Excellent point!

We can discuss as a team and I can implement in next PR

PR #335, opened 2021-12-02
Implement retries for buildah bud commands

JAVGan:
# Implement Retries for Buildah Bud Commands & More
 
 This PR includes the following enhancements:
 
1. Retry mechanism for buildah when receiving HTTP 50X from registry
2. Type annotations for new/touched pieces of code based on [this guide](https://towardsdatascience.com/how-to-make-python-statically-typed-the-essential-guide-e087cf4fa400)
3. Assure buildah will use Docker v2 manifest when building images to avoid generating images with label `application/vnd.oci.image.manifest.v1+json` and later raising `IIBError(The pull specification of ... is neither a v2 manifest list nor a v2 manifest)` when using the previous image

## Observations
1. Since the retry would occur in case of error I created a new exception called `ExternalServiceError` which inherits from `IIBError` to allow the `failed_request_callback` to catch it when all retry attempts failed as well as to allow the `retry` module to recognize it as a different exception.
2. New tests were added for retry mechanism
2. No tests were added for Docker v2 manifest but one was changed to assure the call was the same
3. Buildah uses [OCI manifest](https://www.mankier.com/1/buildah-build#--format) as default when not specifying the format, I just don't know from which version it started doing it

Refers to CLOUDDST-10213


iib/workers/tasks/build.py
```diff
@@ -54,21 +61,26 @@ def _build_image(dockerfile_dir, dockerfile_name, request_id, arch):
...
     # on the image, **and** on its config blob are set correctly.
+    #
+    # NOTE: The argument "--format docker" ensures buildah will not generate an index image with
+    # default OCI v1 manifest but always use Docker v2 format.
     run_cmd(
         [
             'buildah',
             'bud',
             '--no-cache',
+            '--format',
```
pbortlov, line 82: :heart:

pbortlov (dismissed):

yashvardhannanavati (changes requested):
Minor comments. Good job!

iib/workers/config.py
```diff
@@ -10,36 +11,38 @@ class Config(object):
...
-    iib_grpc_max_port_tries = 100
-    iib_grpc_max_tries = 5
-    iib_grpc_start_port = 50051
-    iib_image_push_template = '{registry}/iib-build:{request_id}'
-    iib_index_image_output_registry = None
-    iib_log_level = 'INFO'
-    iib_organization_customizations = {}
-    iib_request_logs_dir = None
-    iib_request_logs_format = (
+    iib_greenwave_url: str = None
```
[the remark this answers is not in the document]
JAVGan: The static types you mean?

iib/workers/config.py
```diff
@@ -10,36 +11,38 @@ class Config(object):
...
-    iib_skopeo_timeout = '300s'
-    iib_total_attempts = 5
-    include = [
+    iib_dogpile_backend: str = 'dogpile.cache.null'
+    iib_dogpile_expiration_time: int = 600
+    iib_dogpile_arguments: Dict[str, List[str]] = {'url': ['127.0.0.1']}
+    iib_skopeo_timeout: str = '300s'
+    iib_total_attempts: int = 5
+    iib_retry_delay: int = 4
+    iib_retry_jitter: int = 2
```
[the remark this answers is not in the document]
JAVGan: Sure!

iib/workers/config.py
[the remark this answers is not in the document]
JAVGan: I believe I figured it out. The static types for configs not related to buildah.

I'm sending a new commit, removing them and also adding the variables on README.

Please let me know if it looks to for you then I re-add the types in a second commit

JAVGan:
@pbortlov @yashvardhannanavati I merged it with master and solved the conflicts, also delivered the changes, please let me know if everything is ok

pbortlov (changes requested):
I hit this error by adding "--format docker" creating new index image /create-empty-index and /add: "The pull specification of registry:8443/iib-build@sha256:1053ca28aabec257c151eb23423e61294c9737d05cefd3c2912358f128d0f554 is neither a v2 manifest list nor a v2 manifest.Media Type: None",

iib/workers/tasks/build.py
```diff
@@ -54,21 +61,26 @@ def _build_image(dockerfile_dir, dockerfile_name, request_id, arch):
...
+            'docker',
```
yashvardhannanavati, line 83: Why not set this to `v2s2`?

JAVGan: According to the [manual](https://www.mankier.com/1/buildah-build#--format) the format should be either `oci` or `docker` being `oci` the default one.

Unfortunately the problem on DEV env goes beyond that... Today I tried to figure out a solution but no luck so far, but in case I don't find it do you prefer to remove this option and use this PR just for `buildah bud` retry?

Then we can investigate this issue on DEV env later, what do you think?

JAVGan:
So, finally I got the problem.

Buildah was correctly creating images with the parameter `--format docker` but when it was going to create the final manifest it was missing the version, as @pbortlov mentioned.

The issue was caused by [this little guy here](https://www.mankier.com/1/buildah-manifest-push#--format) and took me a while to discover.

Basically `buildah-manifest` also uses `oci` as default so I had to change it to `v2s2`.

Finally I changed a bit the `docker-compose.yml` to ensure `podman`, `buildah` and `skopeo` don't get lost with registry on DEV env.

iib/workers/tasks/build.py
JAVGan: Fixed.

When `buildah bud` we need to pass it as `--format docker`

When `buildah manifest ` we need to pass it as `--format v2s2`

docker-compose.yml
```diff
@@ -113,6 +113,7 @@ services:
           -- celery -A iib.workers.tasks worker --loglevel=info
     environment:
       IIB_DEV: 'true'
+      REGISTRY_AUTH_FILE: '/root/.docker/config.json'
```
lipoja, line 116: Just question: It did not work without it? 
I thought that container tools are looking for this path by default.

iib/workers/tasks/utils.py
```diff
@@ -634,18 +637,24 @@ def run_cmd(cmd, params=None, exc_msg=None, strict=True):
...
             # Capture the error message right before the help display
-            regex = r'^(?:Error: )(.+)$'
+            regex: str = r'^(?:Error: )(.+)$'
             # Start from the last log message since the failure occurs near the bottom
             for msg in reversed(response.stderr.splitlines()):
-                match = re.match(regex, msg)
+                match: re.Match = re.match(regex, msg)
                 if match:
                     raise IIBError(f'{exc_msg.rstrip(".")}: {match.groups()[0]}')
+        if cmd[0] == 'buildah':
```
lipoja, line 652: Nitpick: `elif`? :thinking:

docker-compose.yml
pbortlov: But it didn't work properly but it does now :heart:

iib/workers/tasks/utils.py
```diff
@@ -634,18 +637,24 @@ def run_cmd(cmd, params=None, exc_msg=None, strict=True):
...
+            # Check for HTTP 50X errors on buildah
+            regex: str = r'(error creating build container).*(HTTP status: 50[0-9]) .*'
```
lipoja, line 654: This might not catch everything, have a look on this error msg: 
```ERROR utils.run_cmd The command "buildah bud --no-cache --override-arch ppc64le --arch ppc64le -t iib-build:55156-ppc64le -f /tmp/iib-cyojeb6s/index.Dockerfile" failed with: Trying to pull *******
error creating build container: parsing image configuration: Error fetching blob: invalid status code from registry 503 (Service Unavailable)
time="2021-12-08T00:09:54-05:00" level=error msg="exit status 125"```

**Review by lipoja — CHANGES_REQUESTED** [review]

**Review by pbortlov — DISMISSED** [review]

File: iib/workers/tasks/utils.py (hunk 7 above)
Comment (JAVGan) [reply iib/workers/tasks/utils.py:654]: Thanks!

I'll change it to this regex and push again:

```
(error creating build container).*(50[0-9]) .*
```

iib/workers/tasks/utils.py
JAVGan: Fixed :D

iib/workers/tasks/utils.py
JAVGan: Pushed, I tested this regex against your file and the error on JIRA, please let me know if it's good :)

pbortlov (dismissed):

docker-compose.yml
yashvardhannanavati, line 116: This should probably be a different commit?

yashvardhannanavati (commented):
One minor comment. Otherwise, LGTM

docker-compose.yml
JAVGan: @lipoja @pbortlov please let me know what do you think.

For me whatever decision about separating or keeping in a single commit is good

JAVGan:
Rebased

docker-compose.yml
lipoja: Personally, I like separate commits. Each commit should contain only one logical change. 

1. Implementing buildah retries
2. Adding REGISTRY_AUTH_FILE to dev env
3. Updating buildah commands to generate/push manifest v2
4. Static type not related to buildah

Then is we use "rebase and merge" instead of "squash and merge" we end up with these four commits in master branch. Which is good because when something is wrong you can revert the exact commit. 
For example you can buildah revert retries but also keeping commit that is fixing the manifest v2. 

Or if you are not sure if someone will use rebase and merge then you can put all of these to separate PRs. 
My goal is to persuade everybody to do rebase and merge and commit the code properly (one logical change per commit). 

We should also keep in mind if we do it this way that **every commit must work**. 
No commit should break behavior of IIB.

JAVGan: Perfect!
Thanks a lot for the clarification, I'll work on it.

JAVGan: Push is done, now with separate commits.

Code is the same.

pbortlov (approved):

lipoja (approved):

PR #337, opened 2021-12-07
Update opm  to v1.19.5 in worker Dockerfile

lipoja:


chandwanitulsi (approved):
LGTM, should also check later if we need to bump up the version in qe, stage and prod envs

PR #338, opened 2021-12-08
Add support for AWS S3 bucket artifacts storage

yashvardhannanavati:


yashvardhannanavati:
Note: I haven't added tests for `api_v1.py` because I wanted to confirm that the approach I am taking is reasonable before adding tests.

iib/workers/s3_utils.py
```diff
@@ -0,0 +1,39 @@
...
+from iib.workers.config import get_worker_config
+
+log = logging.getLogger(__name__)
+
+conf = get_worker_config()
+
+
+def upload_file_to_s3_bucket(file_path, key_prefix, file_key):
+    """
+    Upload artifact file to AWS S3 bucket.
```
JAVGan, line 16: Why do we have `conf` outside the function? 

Wouldn't it be better to put it inside? I saw somewhere that `get_worker_config` can lead to some circular imports, so if no other function in this file will use it maybe it's better to put it inside.

BTW, thanks for creating a separate utils for S3!

iib/workers/s3_utils.py
```diff
@@ -0,0 +1,39 @@
...
+
+    :param str file_path: the path of the file locally where the artifact file is stored
+    :param str key_prefix: the logical location of the file in the S3 bucket
+    :param str key_prefix: the name of the file in S3 bucket
```
JAVGan, line 20: `key_prefix` 2 times? I believe it's a typo, the last one should be `file_key` IMO

iib/workers/s3_utils.py
```diff
@@ -0,0 +1,39 @@
...
+    :raises IIBError: when unable to upload file to the S3 bucket
+    """
+    try:
+        log.info(
+            'Uploading file %s/%s to S3 bucket: %s',
+            key_prefix,
+            file_key,
+            conf['iib_aws_s3_bucket_name'],
```
JAVGan, line 28: Here you are uploading whatever `{key_prefix}/{file_key}` to S3 which can lead to any function creating structures as they want and eventually causing mess on S3 tree.

Wouldn't it be better to limit some specific `key_prefix` which the function accepts? Then you prevent mess on S3.

Let me know what do you think about it.

iib/web/app.py
```diff
@@ -106,6 +106,33 @@ def validate_api_config(config):
...
+    if config['IIB_AWS_S3_BUCKET_NAME']:
+        if not isinstance(config['IIB_AWS_S3_BUCKET_NAME'], str):
+            raise ConfigError(
+                '"IIB_AWS_S3_BUCKET_NAME" must be set to a valid string. '
+                'This is used for read/write access to the s3 bucket by IIB'
+            )
+        if (
+            not os.getenv('AWS_ACCESS_KEY_ID')
+            or not os.getenv('AWS_SECRET_ACCESS_KEY')
+            or not os.getenv('AWS_DEFAULT_REGION')
```
JAVGan, line 127: Please add the new environment vars to `README`

iib/web/config.py
```diff
@@ -10,6 +10,7 @@ class Config(object):
 
     # Additional loggers to set to the level defined in IIB_LOG_LEVEL
     IIB_ADDITIONAL_LOGGERS = []
+    IIB_AWS_S3_BUCKET_NAME = None
```
JAVGan, line 13: Please add to `README`

iib/web/api_v1.py
```diff
@@ -170,22 +171,44 @@ def get_build_logs(request_id):
...
+    s3_bucket_name = flask.current_app.config['IIB_AWS_S3_BUCKET_NAME']
+    if not s3_bucket_name and not request_log_dir:
         raise NotFound()
 
     request = Request.query.get_or_404(request_id)
-    log_file_path = os.path.join(request_log_dir, f'{request_id}.log')
-    if not os.path.exists(log_file_path):
+
+    finalized = request.state.state_name in RequestStateMapping.get_final_states()
+    if not finalized:
```
lipoja, line 181: With this we are changing behavior also for loading log files from localhost (VM), right?
We get logs only in final state.

iib/web/api_v1.py
```diff
@@ -170,22 +171,44 @@ def get_build_logs(request_id):
...
+        raise ValidationError(
+            f'The request {request_id} is not complete yet.'
+            ' logs will be available once the request is complete.'
+        )
+
+    # If S3 bucket is configured, fetch the log file from the S3 bucket.
+    # Esle, check if logs are stored on the system itself and return them.
```
lipoja, line 188: Typo: `Esle`

iib/web/api_v1.py
```diff
@@ -170,22 +171,44 @@ def get_build_logs(request_id):
...
+    if not s3_bucket_name and not request_log_dir:
         raise NotFound()
 
     request = Request.query.get_or_404(request_id)
-    log_file_path = os.path.join(request_log_dir, f'{request_id}.log')
-    if not os.path.exists(log_file_path):
+
+    finalized = request.state.state_name in RequestStateMapping.get_final_states()
+    if not finalized:
+        raise ValidationError(
```
lipoja, line 182: Could you add `:raise ValidationError....` to docstring?

iib/web/api_v1.py
```diff
@@ -217,17 +241,38 @@ def get_related_bundles(request_id):
             ' related_bundles will be available once the request is complete.'
         )
 
+    # If S3 bucket is configured, fetch the related buyndles file from the S3 bucket.
+    # Esle, check if related bundles are stored on the system itself and return them.
```
lipoja, line 245: Typo Esle

iib/web/api_v1.py
```diff
@@ -217,17 +241,38 @@ def get_related_bundles(request_id):
...
         request_related_bundles_dir, f'{request_id}_related_bundles.json'
     )
     if not os.path.exists(related_bundles_file_path):
         expired = request.temporary_data_expiration < datetime.utcnow()
         if expired:
             raise Gone(f'The related_bundles for the build request {request_id} no longer exist')
+        if request.organization:
+            raise IIBError(
+                'IIB is done processing the request and cannot find related_bundles. Please make '
+                f'sure the iib_organizaiton_customizations for organization {request.organization}'
```
lipoja, line 269: typo? `iib_organizaiton_customizations`

iib/web/s3_utils.py
```diff
@@ -0,0 +1,33 @@
...
+
+def get_file_from_s3_bucket(file_key, bucket_name):
+    """
+    Get artifact file from AWS S3 bucket.
+
+    :param str file_key: the key of the file to be fetched from the S3 bucket
+    :param str bucket_name: the name of the S3 bucket to fetch the file from
+    :return: the body of the S3 object or None
+    :rtype: botocore.response.StreamingBody
+    """
```
JAVGan, line 17: Here `file_key` is the combination of `key_prefix` and `file_prefix` to fetch the object stored on  S3, right?  Something like `related_bundles/12345_related_bundles.json` for example.

I think it's a bit inconsistent to store the logs using `key_prefix and `file_prefix` then retrieving them by using a single parameter, can you please elaborate why are we doing this?

Finally, if we're using `conf['iib_aws_s3_bucket_name']` for the bucket, why do we need `bucket_name` as parameter?

iib/workers/config.py
```diff
@@ -11,6 +11,7 @@ class Config(object):
 
     # When publishing a message, don't continuously retry or else the HTTP connection times out
     broker_transport_options = {'max_retries': 10}
+    iib_aws_s3_bucket_name = None
```
JAVGan, line 14: `README` again

iib/web/api_v1.py
```diff
@@ -170,22 +171,44 @@ def get_build_logs(request_id):
...
+        raise ValidationError(
+            f'The request {request_id} is not complete yet.'
+            ' logs will be available once the request is complete.'
+        )
+
+    # If S3 bucket is configured, fetch the log file from the S3 bucket.
+    # Esle, check if logs are stored on the system itself and return them.
+    # Otherwise, raise an IIBError.
+    if s3_bucket_name:
+        log_file = get_file_from_s3_bucket(f'request_logs/{request_id}.log', s3_bucket_name)
```
JAVGan, line 191: Typo on `Else` hehe

Besides that, here's the confirmation that we're storing the S3 object using `key_prefix` and `file_prefix` but retrieving them as a single concatenated parameter. Not sure if this is the best approach, IMO is inconsistent.

iib/web/api_v1.py
```diff
@@ -217,17 +241,38 @@ def get_related_bundles(request_id):
             ' related_bundles will be available once the request is complete.'
         )
 
+    # If S3 bucket is configured, fetch the related buyndles file from the S3 bucket.
+    # Esle, check if related bundles are stored on the system itself and return them.
+    # Otherwise, raise an IIBError.
+    if s3_bucket_name:
+        related_bundles_file = get_file_from_s3_bucket(
+            f'related_bundles/{request_id}_related_bundles.json', s3_bucket_name
+        )
```
JAVGan, line 250: Typos on `bundles` and `Else`

Another point is that it looks like you're repeating yourself for the keys `request_logs` and `related_bundles`.

The content inside this `if` looks similar to the other one above, just changing the `key_prefix` and whether is `text/plain` or `application/json`.

Maybe there's a more generic way to do that...

JAVGan (changes requested):
Please check the comments for typos, S3 store/retrieve inconsistency and the "don't repeat yourself" on `get_build_logs`

iib/web/api_v1.py
yashvardhannanavati: yes, that's correct

iib/workers/s3_utils.py
yashvardhannanavati: So S3 doesn't have a tree. So when we create something like `{key_prefix}/{file_key}` it's technically a file with name `{key_prefix}/{file_key}`. So I think it wouldn't matter if we limit some specific values or not. That's my two cents. please let me know if you feel strongly against it.

JAVGan: No problem at all.

S3 uses APIs to manage the FS but if you look at AWS Console you'll see that when sending a file with a `{key_prefix}` before it S3 kinda "creates" a directory, it's not exactly a directory, but look likes on console.

Anyway it's not a huge problem for me, you can leave it as it is.

iib/web/s3_utils.py
yashvardhannanavati: No specific reason, I can change it to accept two params. Good point!

Because we have `conf['iib_aws_s3_bucket_name']` in the worker config.

 And here, we have it in the Flask config. I can `import flask` and then use `flask.current_app.config['IIB_AWS_S3_BUCKET_NAME']` to fetch the bucket name but that would be unnecessary operation because we are already doing it in `get_request_logs` and `get_related_bundles` which are the functions calling get_file_from_s3_bucket. Hence, the bucket name as a parameter

iib/web/app.py
yashvardhannanavati: Yes, I'll add it with the tests for api_v1.py

iib/web/api_v1.py
yashvardhannanavati: I did think about this and was torn between the option of creating a helper function or not and decided against it. But now that you have mentioned it, I'll go ahead and do it ;)

yashvardhannanavati:
@lipoja  @JAVGan I have made the requested changes and added the tests for api_v1.py. Could you please review it again?

Ah! I just realized I will have to add boto3 to py38 requirements.txt too. I'll add it tomorrow morning. In the meanwhile could you please take a look at the rest of the PR?

JAVGan (approved):

iib/web/api_v1.py
```diff
@@ -168,24 +192,43 @@ def get_build_logs(request_id):
     :rtype: flask.Response
     :raise NotFound: if the request is not found or there are no logs for the request
     :raise Gone: if the logs for the build request have been removed due to expiration
+    :raise ValidationError: if the request has not completed yet
     """
     request_log_dir = flask.current_app.config['IIB_REQUEST_LOGS_DIR']
-    if not request_log_dir:
+    s3_bucket_name = flask.current_app.config['IIB_AWS_S3_BUCKET_NAME']
+    if not s3_bucket_name and not request_log_dir:
         raise NotFound()
```
lipoja, line 200: I know that you did not change it but is this really NotFound kind of error?  
For me it is more like the error you are raising below:
```
raise IIBError(
      'IIB is done processing the request and could not find logs.'
      ' Please make sure either an S3 bucket is configured or the logs are'
      ' stored locally in a directory by specifying IIB_REQUEST_LOGS_DIR'
)
```
 
But I am fine with keeping current behavior.

iib/workers/s3_utils.py
```diff
@@ -0,0 +1,38 @@
...
+log = logging.getLogger(__name__)
+
+
+def upload_file_to_s3_bucket(file_path, key_prefix, file_key):
+    """
+    Upload artifact file to AWS S3 bucket.
+
+    :param str file_path: the path of the file locally where the artifact file is stored
+    :param str key_prefix: the logical location of the file in the S3 bucket
+    :param str file_key: the name of the file in S3 bucket
```
lipoja, line 18: since it `file_key` should be name, what about naming this parameter differently?
What about `s3_file_name`? And maybe update second parameter this in similar way `s3_key_prefix` or just `s3_prefix`?

Why I am asking to think about it is the fact that I had to read what `file_key` is to understand that it is just file name on s3.

tests/test_web/test_api_v1.py
```diff
@@ -225,22 +227,21 @@ def test_get_unique_bundles(bundles, exp_bundles, app):
...
-        (None, False, False, {'status': 200, 'mimetype': 'text/plain', 'data': ''}),
+        ('foobar', False, True, {'status': 200, 'mimetype': 'text/plain', 'data': 'foobar'}),
+        ('foobar', True, True, {'status': 200, 'mimetype': 'text/plain', 'data': 'foobar'}),
+        ('', False, True, {'status': 200, 'mimetype': 'text/plain', 'data': ''}),
+        ('', True, True, {'status': 200, 'mimetype': 'text/plain', 'data': ''}),
         (
             None,
             True,
-            False,
+            True,
```
lipoja, line 237: Is it necessary to keep `finalized` column/parameter since it is always `True`?

On the other hand, is it possible to test `get_build_logs` on incomplete request or logs not found? (Or creating JIra issue if it not in scope)

iib/workers/s3_utils.py
```diff
@@ -0,0 +1,38 @@
...
+            file_key,
+            conf['iib_aws_s3_bucket_name'],
+        )
+        s3 = boto3.resource(service_name='s3')
+        s3.meta.client.upload_file(
+            Filename=file_path,
+            Bucket=conf['iib_aws_s3_bucket_name'],
+            Key=f'{key_prefix}/{file_key}',
+        )
+    except Exception as error:
```
lipoja, line 35: Do you have somewhere list of Exceptions that can be raised? I did not find one quickly. 
I know that it is not good practice to catch broad Exception.

README.md
```diff
@@ -209,6 +214,16 @@ The custom configuration options for AMQP 1.0 messaging are listed below:
...
   must be set if messaging is enabled.
 
+If you wish to configure AWS S3 bucket for storing artifact files, the following **environment variables**
+must be set along with `IIB_AWS_S3_BUCKET_NAME` config variable:
+
+* `AWS_ACCESS_KEY_ID` - the secret access key ID used to access the AWS S3 bucket.
+* `AWS_SECRET_ACCESS_KEY` - the secret access key used to access the AWS S3 bucket.
+* `AWS_DEFAULT_REGION` - the default region of the AWS S3 bucket
+
+More info on these environment variables can be found here:
```
lcarva, line 224: Maybe this is a little nicer?

> "... can be found in the [AWS User Guide](https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-envvars.html)."

Same for the worker config section.

README.md
```diff
@@ -226,6 +241,10 @@ The custom configuration options for the Celery workers are listed below:
 * `iib_api_timeout` - the timeout in seconds for HTTP requests to the REST API. This defaults to
   `30` seconds.
 * `iib_api_url` - the URL to the IIB REST API (e.g. `https://iib.domain.local/api/v1/`).
+* `iib_aws_s3_bucket_name` - the name of the AWS S3 bucket used to store artifact files like logs
+  and related_bundles if specified. `iib_request_logs_dir` and `iib_request_related_bundles_dir`.
```
lcarva, line 245: Trailing "." at the end of this sentence?

iib/web/api_v1.py
lcarva: This is only ever executed when IIB is not configured to serve logs at all. In this case, it's expected that it will always return the same status code regardless of the status of the request. Since this configuration is optional, it shouldn't be a server error.

The argument for treating it as a client error is that the client made a request to an endpoint which is not supported by the IIB instance. Of course, the client has no way of knowing that beforehand. I do think 404 makes the most sense here.

iib/web/api_v1.py
```diff
@@ -168,24 +192,43 @@ def get_build_logs(request_id):
...
             raise Gone(f'The logs for the build request {request_id} no longer exist')
-        finalized = request.state.state_name in RequestStateMapping.get_final_states()
-        if finalized:
-            raise NotFound()
-        # The request may not have been initiated yet. Return empty logs until it's processed.
-        return flask.Response('', mimetype='text/plain')
+        raise IIBError(
+            'IIB is done processing the request and could not find logs.'
+            ' Please make sure either an S3 bucket is configured or the logs are'
+            ' stored locally in a directory by specifying IIB_REQUEST_LOGS_DIR'
```
lcarva, line 228: I don't think it makes sense to mention `IIB_REQUEST_LOGS_DIR` to users. Maybe the raised error should be just the first sentence. We can log a warning with more information that would help the IIB maintainers debug issues.

iib/web/s3_utils.py
```diff
@@ -0,0 +1,27 @@
...
+    file_name = f'{key_prefix}/{file_key}'
+    try:
+        log.info('getting file from s3 : %s', file_name)
+        s3 = boto3.resource(service_name='s3')
+        response = s3.meta.client.get_object(Bucket=bucket_name, Key=file_name)
+    except Exception as error:
+        log.debug('Unable to fetch object %s from bucket %s: %s', file_name, bucket_name, error)
```
lcarva, line 25: Is `debug` the right level here? This looks like it should be `exception` level.

iib/workers/s3_utils.py
lcarva: Any mess that we create on s3 will be self-inflicted since the `key_prefix` and `file_key` are not user provided parameters. I don't think restricting to certain values will significantly reduce the chances of us making a mess.

iib/workers/s3_utils.py
```diff
@@ -0,0 +1,38 @@
...
+        raise IIBError(
```
lcarva, line 36: How about logging in exception level so we get a the full traceback?

iib/workers/tasks/utils.py
```diff
@@ -695,6 +696,8 @@ def wrapper(*args, **kwargs):
         finally:
             if request_log_handler:
                 logger.removeHandler(request_log_handler)
+                if worker_config['iib_aws_s3_bucket_name']:
+                    upload_file_to_s3_bucket(log_file_path, f'{request_id}.log', 'request_logs')
```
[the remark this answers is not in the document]
lcarva: I think before making this call, we need to flush the log handler, something like `request_log_handler.flush()`. Actually, it might be a good idea to do that always, right after the `removeHandler` call.

iib/web/api_v1.py
lipoja: @lcarva Thanks, I agree.

iib/workers/s3_utils.py
yashvardhannanavati: I agree with the point of it not being a good practice but it's a long list so I decided to catch the general exception. Please let me know if you think there's something else I should do

(https://docs.aws.amazon.com/AmazonS3/latest/API/ErrorResponses.html) Check out "List of Error Codes" section.

iib/workers/tasks/utils.py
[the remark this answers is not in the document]
lcarva: My comment is addressed and can be resolved. I won't resolve this conversation to make it easier for @lipoja to review the changes to also verify his comment is addressed.

(Apologies for mixing conversations. I thought I had created a new one)

iib/workers/s3_utils.py
lipoja: @yashvardhannanavati what about getting inspiration from this example: https://boto3.amazonaws.com/v1/documentation/api/latest/guide/s3-uploading-files.html#uploading-files 
catching `ClientError` from `botocore.exceptions` might be better then catching general `Exception`.

More on error handling: https://boto3.amazonaws.com/v1/documentation/api/latest/guide/error-handling.html#catching-botocore-exceptions  
> The most common botocore exception you’ll encounter is ClientError. This is a general exception when an error response is provided by an AWS service to your Boto3 client’s request.

iib/workers/tasks/utils.py
```diff
@@ -695,6 +696,9 @@ def wrapper(*args, **kwargs):
         finally:
             if request_log_handler:
                 logger.removeHandler(request_log_handler)
+                request_log_handler.flush()
```
lipoja, line 699: should not we flush before removing the handler? :thinking:

tests/test_web/test_api_v1.py
lipoja: :grey_question:

iib/workers/tasks/utils.py
lcarva: An argument for doing it after is that we have a better guarantee that nothing else will be written to the handler after flushing it. Since these calls are done right next to each other, it probably doesn't make a difference.

tests/test_web/test_api_v1.py
yashvardhannanavati: I already added the test for incomplete request in this PR :)

yashvardhannanavati: So I believe we don't have to file a separate Jira for it anymore

iib/workers/s3_utils.py
yashvardhannanavati: Done

iib/web/api_v1.py
lipoja: @yashvardhannanavati Do you agree with Luiz proposal?

iib/workers/tasks/utils.py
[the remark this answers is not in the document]
lipoja: Thanks, changes were addressed.

iib/web/api_v1.py
yashvardhannanavati: So, we have a similar error message in `related_bundles` which is why I decided to keep it consistent here. But if you feel otherwise strongly, I can change it :)

yashvardhannanavati: Apologies for not replying to this earlier

lipoja (approved):

iib/web/api_v1.py
yashvardhannanavati: Changed it :)

lipoja: Thanks @yashvardhannanavati !

lipoja (approved):

JAVGan (approved):

PR #340, opened 2021-12-10
[WIP] Adding support for deprecate bundles for FBC image

lipoja:


pbortlov (changes requested):
Tests are failing

lipoja:
> Tests are failing

Tests are fixed.

iib/workers/tasks/opm_operations.py
```diff
@@ -193,6 +193,130 @@ def _serve_cmd_at_port(serve_cmd, cwd, port, max_tries, wait_time):
...
+    index_db_file = os.path.join(base_dir, get_worker_config()['temp_index_db_path'])
+
+    if os.path.exists(index_db_file):
+        log.debug('Temp index.db already exist for %s', from_index)
+        return index_db_file
+
+    log.info('Temp index.db does not exist yet for %s', from_index)
+    if from_index:
+        log.info('Using the existing database from %s', from_index)
+        return get_hidden_index_database(from_index, base_dir)
```
pbortlov, line 217: How do you decide if we have to reach out to a hidden index database or not? Is it defined by from_index?

iib/workers/tasks/opm_operations.py
```diff
@@ -193,6 +193,130 @@ def _serve_cmd_at_port(serve_cmd, cwd, port, max_tries, wait_time):
...
+def opm_registry_deprecatetruncate(
+    base_dir,
+    index_db,
+    bundles,
+    from_index=None,
+    overwrite_target_index_token=None,
+    container_tool=None,
+):
+    """
+    Migrate SQLite database to File-Based catalog using opm command.
```
pbortlov, line 238: in the name of the function is nothing mention about migration to FBC

iib/workers/tasks/opm_operations.py
```diff
@@ -193,6 +193,130 @@ def _serve_cmd_at_port(serve_cmd, cwd, port, max_tries, wait_time):
     raise IIBError(f'Index registry has not been initialized after {max_tries} tries')
 
 
+def _get_temp_index_db_file(base_dir, from_index):
```
pbortlov, line 196: What is the use of the function? Could we just get the file name from the "main" function  `with tempfile.TemporaryDirectory(prefix='iib-') as temp_dir:`

iib/workers/tasks/build.py
```diff
@@ -812,15 +817,25 @@ def handle_add_request(
...
-            deprecate_bundles(
-                deprecation_bundles,
-                temp_dir,
-                prebuild_info['binary_image'],
-                intermediate_image_name,
-                overwrite_from_index_token,
-                # Use podman so opm can find the image locally
-                container_tool='podman',
-            )
+            if is_fbc:
```
pbortlov, line 820: Can we remove lines 810-813 in favor of this one?

iib/workers/tasks/build_merge_index_image.py
```diff
@@ -196,7 +197,8 @@ def handle_merge_request(
     target_index_resolved = prebuild_info['target_index_resolved']
 
     with tempfile.TemporaryDirectory(prefix='iib-') as temp_dir:
-        if is_image_fbc(source_from_index_resolved):
+        is_fbc = is_image_fbc(source_from_index_resolved)
+        if is_fbc:
```
pbortlov, line 201: Could we remove this in favor line 254?

pbortlov (changes requested):

iib/workers/tasks/build.py
lipoja: :facepalm: Thanks!

iib/workers/tasks/build_merge_index_image.py
lipoja: No because merge index image is not yet supported. It should be removed in https://github.com/release-engineering/iib/pull/343

iib/workers/tasks/opm_operations.py
lipoja: it is not getting name but making sure that the file exist on returned path

iib/workers/tasks/opm_operations.py
lipoja: This should be used for FBC only. That is because in some cases for FBC we want to create empty index.db file.
I've added better docstring, renamed and added assert.

iib/workers/tasks/opm_operations.py
lipoja: Thanks

iib/workers/tasks/opm_operations.py
pbortlov: could we name that function accordingly?

iib/workers/tasks/build.py
```diff
@@ -802,25 +807,31 @@ def handle_add_request(
 
         arches = prebuild_info['arches']
         if deprecation_bundles:
-            if is_fbc:
```
pbortlov, line 805: Cool

iib/workers/tasks/build_merge_index_image.py
pbortlov: ok, I will check the next PR

iib/workers/tasks/opm_operations.py
pbortlov: I see you did.

lipoja: I renamed it already... is it not better?

iib/workers/tasks/opm_operations.py
```diff
@@ -193,6 +193,134 @@ def _serve_cmd_at_port(serve_cmd, cwd, port, max_tries, wait_time):
...
+        assert is_image_fbc(from_index), 'Only File-Based Catalog index images are supported'
+        log.info('Using the existing database from %s', from_index)
+        return get_hidden_index_database(from_index, base_dir)
+    else:
+        log.info('Creating empty database file %s', index_db_file)
+        index_db_dir = os.path.dirname(index_db_file)
+        if not os.path.exists(index_db_dir):
+            os.makedirs(index_db_dir, exist_ok=True)
+        with open(index_db_file, 'w'):
+            pass
```
JAVGan, line 228: > This function accepts FBC images only because we do not want to create empty index.db
    in any other case.
    
 I have a doubt here... IIUC if the function receives the parameter `from_index` you check it to make sure it's FBC using the `assert` statement, however,  if the function doesn't receive any `from_index` at all you create an empty index.db anyways. So how does this function prevents another one to pass to it a SQLite image base-dir?

iib/workers/tasks/opm_operations.py
```diff
@@ -193,6 +193,134 @@ def _serve_cmd_at_port(serve_cmd, cwd, port, max_tries, wait_time):
...
+
+
+def opm_registry_deprecatetruncate(
+    base_dir,
+    index_db,
+    bundles,
+    from_index=None,
+    overwrite_target_index_token=None,
+    container_tool=None,
+):
```
JAVGan, line 240: Will this replace `iib.workers.tasks.utils.deprecate_bundles`?

Because this function seems to be a "clone" of it and if we want to move it from here I believe we need to remove the `deprecate_bundles` from `utils.py` an then including it here.

iib/workers/tasks/opm_operations.py
```diff
@@ -193,6 +193,134 @@ def _serve_cmd_at_port(serve_cmd, cwd, port, max_tries, wait_time):
...
+def deprecate_bundles_fbc(
+    bundles,
+    base_dir,
+    binary_image,
+    from_index,
+    overwrite_target_index_token=None,
+    container_tool=None,
+):
+    """
+    Deprecate the specified bundles from the index image.
```
JAVGan, line 288: I know it might seem a bit redundant but can you please change this description to something like: "Deprecate the specified bundles from the index image when it's FBC?"

Because this text is exactly the same from `iib.workers.tasks.utils.deprecate_bundles` and I believe some confusion might appear.

Also please consider removing the `deprecate_bundles` from `utils.py` as mentioned in the other comment.

tests/test_workers/test_tasks/test_opm_operations.py
```diff
@@ -328,3 +329,103 @@ def test_opm_registry_add_fbc(
...
+    if db_exist:
+        create_index_db()
+
+    index_db_file = opm_operations._get_temp_index_db_fbc(base_dir=tmpdir, from_index=from_index)
+    assert os.path.isfile(index_db_file)
+
+
+@pytest.mark.parametrize('from_index', (None, 'some_index:latest'))
+@pytest.mark.parametrize('bundles', (['bundle:1.2', 'bundle:1.3'], []))
+@pytest.mark.parametrize('container_tool', (None, 'podwoman'))
```
JAVGan, line 358: `podwoman`, LOL :smile: 

Made my day!

JAVGan (changes requested):
Even though `iib.workers.tasks.opm_operations.opm_registry_deprecatetruncate` and `iib.workers.tasks.utils.deprecate_bundles` have different parameters they're too similar.

Please consider removing one of them and making the other one more generic to cover both cases.

iib/workers/tasks/opm_operations.py
```diff
@@ -193,6 +193,134 @@ def _serve_cmd_at_port(serve_cmd, cwd, port, max_tries, wait_time):
     raise IIBError(f'Index registry has not been initialized after {max_tries} tries')
 
 
+def _get_temp_index_db_fbc(base_dir, from_index):
```
yashvardhannanavati, line 196: should from_index be optional? Since you are checking for the presence of that param?

iib/workers/tasks/build.py
```diff
@@ -802,25 +807,31 @@ def handle_add_request(
...
-                deprecation_bundles,
-                temp_dir,
-                prebuild_info['binary_image'],
-                intermediate_image_name,
-                overwrite_from_index_token,
-                # Use podman so opm can find the image locally
-                container_tool='podman',
-            )
+            if is_fbc:
+                deprecate_bundles_fbc(
```
chandwanitulsi, line 817: Do we need log message suggesting FBC image processing here?

iib/workers/tasks/opm_operations.py
chandwanitulsi: +1, and there is also creation index.db file if from_index is not set

iib/workers/tasks/opm_operations.py
chandwanitulsi: deprecate_bundles in utils.py is a different function, it runs opm index deprecatetruncate commands directly on from_index whereas in opm_registry_deprecatetruncate we are running opm registry deprecatetruncate which works on index db. The functionality of both functions sounds similar but technically they are not

iib/workers/tasks/opm_operations.py
```diff
@@ -193,6 +193,134 @@ def _serve_cmd_at_port(serve_cmd, cwd, port, max_tries, wait_time):
...
+    :param list bundles: pull specifications of bundles to deprecate.
+    :param str base_dir: base directory where operation files will be located.
+    :param str binary_image: binary image to be used by the new index image.
+    :param str from_index: index image, from which the bundles will be deprecated.
+    :param str overwrite_target_index_token: the token used for overwriting the input
+        ``from_index`` image. This is required to use ``overwrite_target_index``.
+        The format of the token must be in the format "user:password".
+    :param str container_tool: the container tool to be used to operate on the index image
+    """
+    index_db_file = _get_temp_index_db_fbc(base_dir=base_dir, from_index=from_index)
```
chandwanitulsi, line 301: Should we use _get_temp_index_db_fbc instead in all add, rm, create functionalities?

iib/workers/tasks/build.py
lipoja: I think we already have this info present in log file from line 756.

iib/workers/tasks/opm_operations.py
lipoja: Yes

iib/workers/tasks/opm_operations.py
lipoja: OK I do not have strong opinion on this one - lets make the function universal - accept bot image types; create empty; and has better name

iib/workers/tasks/opm_operations.py
lipoja: Updated docstring, I can not remove different function - see the other comment (response from Tulsi)

iib/workers/tasks/opm_operations.py
lipoja: I changed this function a lot, please check if it is better now.

JAVGan (dismissed):

iib/workers/tasks/build.py
```diff
@@ -802,25 +807,29 @@ def handle_add_request(
...
+                    bundles=deprecation_bundles,
+                    base_dir=temp_dir,
+                    binary_image=prebuild_info['binary_image'],
+                    from_index=intermediate_image_name,
```
yashvardhannanavati, line 821: Looks like overwrite_target_index_token is missing here. Is it intentional?

iib/workers/tasks/opm_operations.py
```diff
@@ -193,6 +193,102 @@ def _serve_cmd_at_port(serve_cmd, cwd, port, max_tries, wait_time):
...
+    Dockerfile is created only, no build is performed.
+
+    :param list bundles: pull specifications of bundles to deprecate.
+    :param str base_dir: base directory where operation files will be located.
+    :param str binary_image: binary image to be used by the new index image.
+    :param str from_index: index image, from which the bundles will be deprecated.
+    """
+    index_db_file = _get_or_create_temp_index_db_file(base_dir=base_dir, from_index=from_index)
+
+    opm_registry_deprecatetruncate(
```
yashvardhannanavati, line 276: If we already know that index_db_file is empty, why bother calling this in those scenarios. But I believe we can keep it this way and later on improve it. I can look the other way on this one :)

iib/workers/tasks/build.py
yashvardhannanavati: No, hang on. The function definition says it doesn't take overwrite_target_index. That means the function call in merge_index_image.py is wrong.

iib/workers/tasks/build_merge_index_image.py
```diff
@@ -249,13 +251,22 @@ def handle_merge_request(
...
-                intermediate_image_name,
-                overwrite_target_index_token,
-            )
+            if is_fbc:
+                deprecate_bundles_fbc(
+                    bundles=deprecation_bundles,
+                    base_dir=temp_dir,
+                    binary_image=prebuild_info['binary_image'],
+                    from_index=intermediate_image_name,
+                    overwrite_target_index_token=overwrite_target_index_token,
```
yashvardhannanavati, line 260: The function definition doesn't have `overwrite_target_index_token `. Should we remove it from here?

yashvardhannanavati (changes requested):

iib/workers/tasks/build.py
lipoja: My bad - Thanks

iib/workers/tasks/build_merge_index_image.py
lipoja: removed

iib/workers/tasks/opm_operations.py
lipoja: I believe that in this deprecatetruncate scenario it will not be created empty but copied from index. I want to use `_get_or_create_temp_index_db_file` also in add and rm once this is merged. Because it shares same code and in add we can hit issue when from_index is not set.

iib/workers/tasks/build.py
```diff
@@ -41,7 +42,6 @@
 
 __all__ = ['handle_add_request', 'handle_rm_request']
 
-
```
yashvardhannanavati, line 44: Accidental?

iib/workers/tasks/build.py
```diff
@@ -808,24 +808,30 @@ def handle_add_request(
...
+                    base_dir=temp_dir,
+                    binary_image=prebuild_info['binary_image'],
+                    from_index=from_index_resolved,
+                )
+            else:
+                # opm can only deprecate a bundle image on an existing index image. Build and
+                # push a temporary index image to satisfy this requirement. Any arch will do.
+                arch = sorted(arches)[0]
+                log.info('Building a temporary index image to satisfy the deprecation requirement')
+                _build_image(temp_dir, 'index.Dockerfile', request_id, arch)
```
yashvardhannanavati, line 822: Won't this be required for FBC?

iib/workers/tasks/build.py
```diff
@@ -808,24 +808,30 @@ def handle_add_request(
...
+                intermediate_image_name = _get_local_pull_spec(
+                    request_id, arch, include_transport=True
+                )
+                deprecate_bundles(
+                    bundles=deprecation_bundles,
+                    base_dir=temp_dir,
+                    binary_image=prebuild_info['binary_image'],
+                    from_index=intermediate_image_name,
```
yashvardhannanavati, line 830: If we are building and pushing an intermediate index, why are we using a local pull spec? 

I was chatting with @chandwanitulsi and I realized there could be a scenario

All the workers use the same virtualenv. What if, when a worker X is processing the request, another request is being processed by a different worker say Y and worker Y runs _cleanup before X gets a chance to use deprecatetruncate on its intermediate image. Because when Y runs cleanup, X's intermediate image would be wiped out, no?

iib/workers/tasks/build.py
lipoja: My guess is no... we are using `opm registry deprecatetruncate` which is working directly with index.db file. 
So my guess is we just need to provide index.db with latest changes (e.g. with newly added bundles).

iib/workers/tasks/build.py
lipoja: 1. I think _clenup() in IIB has nothing in common with python virtualenv (if you are referring to that)
2. Clean up runs  `podman rmi --all --force`. I might be wrong here but I thought that this removes all images created by the user that is running the command. And we have multiple users running in parallel but processing requests sequentially. Aren't those images locate in `~/.local/share/containers/storage/` when we are running podman as users and not root?

iib/workers/tasks/build.py
yashvardhannanavati: ack, makes sense

iib/workers/tasks/build.py
yashvardhannanavati: ack, this makes sense too.

iib/workers/tasks/build.py
JAVGan: Maybe OCD? :stuck_out_tongue: 

But for me it's ok to remove this line, black doesn't complain as well.

JAVGan (dismissed):
LGTM

yashvardhannanavati (dismissed):

iib/workers/tasks/build.py
```diff
@@ -808,24 +808,30 @@ def handle_add_request(
         arches = prebuild_info['arches']
         if deprecation_bundles:
             if is_fbc:
```
pbortlov, line 810: I have problem with `is_fbc` variable.. the definition is on `lower` level (spaces, tabs- my brain doesn't work) and then you use it on different

lipoja: ok, I'll fix that - but I guess it must be working since this is already on PROD :D

lipoja: fixed

tests/test_workers/test_tasks/test_build_merge_index_image.py
```diff
@@ -10,6 +10,7 @@
 from iib.workers.tasks.utils import RequestConfigMerge
 
 
+@pytest.mark.parametrize('source_fbc, target_fbc', [(False, False), (False, True), (True, True)])
```
JAVGan, line 13: What about the case `(True, False)` when source == FBC and target != FBC?

tests/test_workers/test_tasks/test_build_merge_index_image.py
```diff
@@ -150,14 +180,20 @@ def side_effect(_, temp_dir, *args, **kwargs):
     )
 
 
+@pytest.mark.parametrize('source_fbc, target_fbc', [(False, False), (False, True), (True, True)])
```
JAVGan, line 183: Same question when source == FBC and target != FBC

JAVGan (commented):
Just small questions on tests

iib/workers/tasks/build_merge_index_image.py
```diff
@@ -279,25 +280,31 @@ def handle_merge_request(
...
-
-            deprecate_bundles(
-                deprecation_bundles,
-                temp_dir,
-                prebuild_info['binary_image'],
-                intermediate_image_name,
-                overwrite_target_index_token,
-            )
+            # we can check if source index is FBC or not because intermediate_image
+            # will be always the same type because it is built
```
chandwanitulsi, line 284: Can you elaborate what you mean by same type?

tests/test_workers/test_tasks/test_build_merge_index_image.py
lipoja: Already covered by https://github.com/release-engineering/iib/blob/50c78e92eb983c9fd3db76276c49c6b7b50cdfad/tests/test_workers/test_tasks/test_build_merge_index_image.py#L608

tests/test_workers/test_tasks/test_build_merge_index_image.py
lipoja: Already covered by https://github.com/release-engineering/iib/blob/50c78e92eb983c9fd3db76276c49c6b7b50cdfad/tests/test_workers/test_tasks/test_build_merge_index_image.py#L608

tests/test_workers/test_tasks/test_build_merge_index_image.py
JAVGan: Tks!

iib/workers/tasks/build_merge_index_image.py
lipoja: by type I meant  SQLite or FBC

JAVGan (approved):

iib/workers/tasks/build_merge_index_image.py
lipoja: should I rephrase it ?

chandwanitulsi: I meant is type will be one exact thing or same as Source
like if source is X (fbc/sqlite) intermediate will be X
I think it got changed after our discussion, so was confirming that

chandwanitulsi (approved):
Merge and add api with deprecation also works fine

PR #343, opened 2021-12-14
Adding support of FBC image to merge index image API endpoint

lipoja:
Adding support of FBC image to merge index image API endpoint

refers to CLOUDDST-8752

### DIFF
--- a/iib/workers/tasks/build_merge_index_image.py
+++ b/iib/workers/tasks/build_merge_index_image.py
@@ -1,9 +1,16 @@
 # SPDX-License-Identifier: GPL-3.0-or-later
 import itertools
 import logging
+import os
 import stat
 import tempfile
 
+from iib.workers.config import get_worker_config
+from iib.workers.tasks.opm_operations import (
+    opm_registry_add_fbc,
+    opm_migrate,
+    opm_generate_dockerfile,
+)
 from packaging.version import Version
 
 from iib.exceptions import IIBError
@@ -118,16 +125,26 @@ def _add_bundles_missing_in_source(
             '%s bundles have invalid version label and will be deprecated.', len(invalid_bundles)
         )
 
-    _opm_index_add(
-        base_dir,
-        missing_bundle_paths,
-        binary_image,
-        from_index=source_from_index,
-        overwrite_from_index_token=overwrite_target_index_token,
-        # Use podman until opm's default mechanism is more resilient:
-        #   https://bugzilla.redhat.com/show_bug.cgi?id=1937097
-        container_tool='podman',
-    )
+    is_source_fbc = is_image_fbc(source_from_index)
+    if is_source_fbc:
+        opm_registry_add_fbc(
+            base_dir=base_dir,
+            bundles=missing_bundle_paths,
+            binary_image=binary_image,
+            from_index=source_from_index,
+            container_tool='podman',
+        )
+    else:
+        _opm_index_add(
+            base_dir=base_dir,
+            bundles=missing_bundle_paths,
+            binary_image=binary_image,
+            from_index=source_from_index,
+            overwrite_from_index_token=overwrite_target_index_token,
+            # Use podman until opm's default mechanism is more resilient:
+            #   https://bugzilla.redhat.com/show_bug.cgi?id=1937097
+            container_tool='podman',
+        )
     _add_label_to_index(
         'com.redhat.index.delivery.version', ocp_version, base_dir, 'index.Dockerfile'
     )
@@ -194,10 +211,22 @@ def handle_merge_request(
     _update_index_image_build_state(request_id, prebuild_info)
     source_from_index_resolved = prebuild_info['source_from_index_resolved']
     target_index_resolved = prebuild_info['target_index_resolved']
+    dockerfile_name = 'index.Dockerfile'
 
     with tempfile.TemporaryDirectory(prefix='iib-') as temp_dir:
-        if is_image_fbc(source_from_index_resolved):
-            err_msg = 'File-Based catalog image type is not supported yet.'
+        source_fbc = is_image_fbc(source_from_index_resolved)
+        target_fbc = is_image_fbc(target_index_resolved)
+
+        # do not remove - logging requested by stakeholders
+        if source_fbc:
+            log.info("Processing source index image as File-Based Catalog image")
+        if target_fbc:
+            log.info("Processing target index image as File-Based Catalog image")
+
+        if source_fbc and not target_fbc:
+            err_msg = (
+                'Cannot merge source File-Based Catalog index image into target SQLite index image.'
+            )
             log.error(err_msg)
             raise IIBError(err_msg)
 
@@ -232,7 +261,6 @@ def handle_merge_request(
         missing_bundle_paths = [bundle['bundlePath'] for bundle in missing_bundles]
         if missing_bundle_paths:
             add_max_ocp_version_property(missing_bundle_paths, temp_dir)
-
         set_request_state(request_id, 'in_progress', 'Deprecating bundles in the deprecation list')
         log.info('Deprecating bundles in the deprecation list')
         intermediate_bundles = missing_bundle_paths + source_index_bundles_pull_spec
@@ -247,16 +275,21 @@ def handle_merge_request(
         ]
 
         if deprecation_bundles:
+            intermediate_image_name = _get_external_arch_pull_spec(
+                request_id, arch, include_transport=False
+            )
+
+            if is_image_fbc(intermediate_image_name):
+                err_msg = 'File-Based catalog image type is not supported for deprecation yet.'
+                log.error(err_msg)
+                raise IIBError(err_msg)
+
             # opm can only deprecate a bundle image on an existing index image. Build and
             # push a temporary index image to satisfy this requirement. Any arch will do.
             # NOTE: we cannot use local builds because opm commands fails,
             # index image has to be pushed to registry
             _build_image(temp_dir, 'index.Dockerfile', request_id, arch)
             _push_image(request_id, arch)
-
-            intermediate_image_name = _get_external_arch_pull_spec(
-                request_id, arch, include_transport=False
-            )
 
             deprecate_bundles(
                 deprecation_bundles,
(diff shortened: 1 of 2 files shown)


pbortlov (changes requested):
Please check tests

lipoja:
> Please check tests

Tests are fixed

iib/workers/tasks/build_merge_index_image.py
```diff
@@ -118,16 +121,26 @@ def _add_bundles_missing_in_source(
...
-        base_dir,
-        missing_bundle_paths,
-        binary_image,
-        from_index=source_from_index,
-        overwrite_from_index_token=overwrite_target_index_token,
-        # Use podman until opm's default mechanism is more resilient:
-        #   https://bugzilla.redhat.com/show_bug.cgi?id=1937097
-        container_tool='podman',
-    )
+    if is_image_fbc(source_from_index):
```
pbortlov, line 124: What if the target image is not FBC?

pbortlov (changes requested):

iib/workers/tasks/build_merge_index_image.py
lipoja: Then the execution of the code will take the `else` branch.

pbortlov: Let me try to ask a better question: What will happen if I try to merge the FBC index(source) with not FBC index (target)? I would suggest checking both the `source` and `target` indexes. Or drop a comment if we merge FBC with non-FBC the result will be "something"

lipoja: Yeah everything is all about the source index. If that one is FBC the result will be FBC as well. 
So I think you can merge FBC source and SQLite target. 

I forget about this one so than you for bringing this up. 
cc @yashvardhannanavati @chandwanitulsi 

We can discuss this behavior if it is OK or if we want to merge FBC indexes only. 
Thanks.

yashvardhannanavati: The approach I would suggest is to check in the beginning if both `source` and `target` indexes are of the same type. If they're not, we simply throw an IIBError indicating the same. 

Thoughts?

lipoja: Yes, however this will not help us when we create this first FBC image, right? 
I mean we would use 4.10 as a source and 4.11 (new empty FBC image) as a target. And it would fail for us. 
Do we want it this way? What if we just check if target is FBC then the result will be FBC as well? 

Thoughts?

chandwanitulsi: +1 from me to o/p FBC if the target is FBC index.
Other option we can consider is using a flag to specify the type of resultant image

tests/test_workers/test_tasks/test_build_merge_index_image.py
```diff
@@ -259,7 +260,7 @@ def test_handle_merge_request_no_deprecate(
 @mock.patch('iib.workers.tasks.build_merge_index_image._opm_index_add')
 @mock.patch('iib.workers.tasks.build_merge_index_image.set_request_state')
 def test_add_bundles_missing_in_source(
-    mock_srs, mock_oia, mock_aolti, mock_bi, mock_pi, mock_capml, mock_gil
+    mock_srs, mock_oia, mock_aolti, mock_bi, mock_pi, mock_capml, mock_gil, mock_iifbc
```
chandwanitulsi, line 263: May I suggest adding a test when is_image_fbc is true

iib/workers/tasks/build_merge_index_image.py
pbortlov: I am fine with just checking a target image. But it would be helpful having a comment above.

iib/workers/tasks/build_merge_index_image.py
```diff
@@ -249,6 +281,11 @@ def handle_merge_request(
         )
 
         if deprecation_bundles:
+            if is_image_fbc(intermediate_image_name):
+                err_msg = 'File-Based catalog image type is not supported for deprecation yet.'
```
chandwanitulsi, line 285: Intermediate image would be FBC when source_image is FBC?

iib/workers/tasks/build_merge_index_image.py
```diff
@@ -249,6 +281,11 @@ def handle_merge_request(
...
+                log.error(err_msg)
+                raise IIBError(err_msg)
+
             deprecate_bundles(
```
chandwanitulsi, line 289: Deprecate_bundles is only run with opm index commands, we don't need to run opm registry deprecatetruncate?

iib/workers/tasks/build_merge_index_image.py
lipoja: yes

iib/workers/tasks/build_merge_index_image.py
lipoja: we need but when I was creating Jiras I thought that it would be easier to make deprecation separately. 
So that is why we have the code for deprecation FBC here https://github.com/release-engineering/iib/pull/340

chandwanitulsi: ah, yes I assumed this PR was based on 340, was directly checking files changed

tests/test_workers/test_tasks/test_build_merge_index_image.py
lipoja: I will add it in deprecate PR - I am not able to do it here because it is raising IIBError because deprecation does not support FBC yet.

iib/workers/tasks/build_merge_index_image.py
```diff
@@ -194,13 +212,24 @@ def handle_merge_request(
...
+        target_fbc = is_image_fbc(target_index_resolved)
+
+        if source_fbc and not target_fbc:
+            err_msg = (
+                'Cannot merge source File-Based Catalog index image into target SQLite index image.'
+            )
             log.error(err_msg)
             raise IIBError(err_msg)
 
+        if source_fbc:
```
pbortlov, line 227: is this needed?

lipoja: yes - we want to have this information in logfiles.

chandwanitulsi: Minor: These 2 logs are useful, how about we move it before raising  above error "Cannot merge source File-Based Catalog index image into target SQLite index image". It will ease reading logs more

lipoja: I think in that error message you have this info as well:
'Cannot merge **source File-Based Catalog** index image into **target SQLite** index image.'

But I can move it up if you think it is better.

iib/workers/tasks/build_merge_index_image.py
```diff
@@ -247,17 +274,22 @@ def handle_merge_request(
         ]
 
         if deprecation_bundles:
+            intermediate_image_name = _get_external_arch_pull_spec(
```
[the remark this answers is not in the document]
lipoja: this function returns string only (pull spec - URL) it will not be changed unless we use different parameters.
If you are referring to actual image and that we are checking `is_image_fbc` before build and push then I would say it will not be changed - the image will be still FBC or SQLite only content will be different.
And this `is_image_fbc` will be removed once we merge the deprecate - so I can move this check down a bit in that PR. 

I put it here at top because at  that time I thought it is better to fail before time consuming build and push.

iib/workers/tasks/build_merge_index_image.py
```diff
@@ -247,17 +274,22 @@ def handle_merge_request(
...
+                raise IIBError(err_msg)
+
             # opm can only deprecate a bundle image on an existing index image. Build and
             # push a temporary index image to satisfy this requirement. Any arch will do.
             # NOTE: we cannot use local builds because opm commands fails,
             # index image has to be pushed to registry
             _build_image(temp_dir, 'index.Dockerfile', request_id, arch)
             _push_image(request_id, arch)
 
-            intermediate_image_name = _get_external_arch_pull_spec(
```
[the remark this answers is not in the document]
lipoja: no

iib/workers/tasks/build_merge_index_image.py
```diff
@@ -266,22 +298,41 @@ def handle_merge_request(
                 overwrite_target_index_token,
             )
 
+        if target_fbc:
```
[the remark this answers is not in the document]
lipoja: No - I want to have the log message before we start processing things.
And this part hast to be done as last one before build/push the final image

tests/test_workers/test_tasks/test_build_merge_index_image.py
chandwanitulsi: It would have been better to have a test without the deprecation involved, but its not a blocker for me

chandwanitulsi (dismissed):
LGTM, just a minor comment on the logs ordering

iib/workers/tasks/build_merge_index_image.py
pbortlov: Can we add comment that was requested

iib/workers/tasks/build_merge_index_image.py
```diff
@@ -232,7 +260,6 @@ def handle_merge_request(
         missing_bundle_paths = [bundle['bundlePath'] for bundle in missing_bundles]
         if missing_bundle_paths:
             add_max_ocp_version_property(missing_bundle_paths, temp_dir)
-
```
pbortlov, line 235: do we need it?

iib/workers/tasks/build_merge_index_image.py
```diff
@@ -118,16 +125,26 @@ def _add_bundles_missing_in_source(
...
-        base_dir,
-        missing_bundle_paths,
-        binary_image,
-        from_index=source_from_index,
-        overwrite_from_index_token=overwrite_target_index_token,
-        # Use podman until opm's default mechanism is more resilient:
-        #   https://bugzilla.redhat.com/show_bug.cgi?id=1937097
-        container_tool='podman',
-    )
+    is_fbc = is_image_fbc(source_from_index)
```
pbortlov, line 128: Can we rename `source_fbc` if we check the source image? Not important

iib/workers/tasks/build_merge_index_image.py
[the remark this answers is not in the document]
pbortlov: DIscussed in a meeting.. I understand now

pbortlov (dismissed):
Just renaming a variable

iib/workers/tasks/build_merge_index_image.py
lipoja: done

pbortlov (approved):
+1

PR #344, opened 2021-12-14
Add olm.maxOpenShiftVersion property to unique bundles

pbortlov:
- get unique bundles to add olm.maxOpenShiftVersion property
- add logging

### DIFF
--- a/iib/workers/tasks/utils.py
+++ b/iib/workers/tasks/utils.py
@@ -91,11 +91,19 @@ def add_max_ocp_version_property(resolved_bundles, temp_dir):
         return
 
     # Filter index image bundles to get pull spec for bundles in the request
-    updated_bundles = list(
-        filter(lambda b: b['bundlePath'] in resolved_bundles, get_bundle_json(raw_bundles))
-    )
+    present_bundles = get_bundle_json(raw_bundles)
+    unique_present_bundles_pull_spec = []
+    unique_present_bundles = []
 
-    for bundle in updated_bundles:
+    for bundle in present_bundles:
+        bundle_path = bundle['bundlePath']
+        if bundle_path in resolved_bundles:
+            if bundle_path in unique_present_bundles_pull_spec:
+                continue
+            unique_present_bundles.append(bundle)
+            unique_present_bundles_pull_spec.append(bundle_path)
+
+    for bundle in unique_present_bundles:
         if _requires_max_ocp_version(bundle['bundlePath']):
             log.info('adding property for %s', bundle['bundlePath'])
             max_openshift_version_property = {
@@ -152,6 +160,7 @@ def _requires_max_ocp_version(bundle):
     :param str bundle: a string representing the bundle pull specification
     :returns bool:
     """
+    log.info('Checking presence of maxOpenshiftVersion in bundle %s', bundle)
     cmd = [
         'operator-sdk',
         'bundle',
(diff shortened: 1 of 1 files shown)


iib/workers/tasks/utils.py
```diff
@@ -91,11 +91,19 @@ def add_max_ocp_version_property(resolved_bundles, temp_dir):
         return
 
     # Filter index image bundles to get pull spec for bundles in the request
-    updated_bundles = list(
-        filter(lambda b: b['bundlePath'] in resolved_bundles, get_bundle_json(raw_bundles))
-    )
+    present_bundles = get_bundle_json(raw_bundles)
```
chandwanitulsi, line 94: May I suggest adding a test and description in the commit/PR for this change?

lipoja: Yep, that would be nice. Otherwise it looks OK.

iib/workers/tasks/utils.py
```diff
@@ -789,7 +798,8 @@ def get_image_arches(pull_spec):
         arches.add(skopeo_out['architecture'])
     else:
         raise IIBError(
-            f'The pull specification of {pull_spec} is neither a v2 manifest list nor a v2 manifest'
+            f'The pull specification of {pull_spec} is neither a v2 manifest list nor a v2 manifest.'
+            f'Media Type: {skopeo_raw.get("mediaType")}'
```
JAVGan, line 802: :heart:

iib/workers/tasks/utils.py
```diff
@@ -91,11 +91,19 @@ def add_max_ocp_version_property(resolved_bundles, temp_dir):
         return
 
     # Filter index image bundles to get pull spec for bundles in the request
```
JAVGan, line 93: I believe it's good to change a bit this comment to reflect that it not only filters the index image bundles to get the pull spec but also it get unique ones.

Maybe something like:
```
# Filter index image bundles to get pull spec for unique bundles in the request
``` 

Anyway, it's a suggestion :smiley:

iib/workers/tasks/utils.py
JAVGan: I agree with adding a test to make sure the bundles are unique and functions like `_requires_max_ocp_version` and `_add_property_to_index` are called as expected.

JAVGan (changes requested):

iib/workers/tasks/utils.py
```diff
@@ -90,23 +90,44 @@ def add_max_ocp_version_property(resolved_bundles, temp_dir):
...
+                    'type': 'olm.maxOpenShiftVersion',
+                    'value': '4.8',
+                    'operatorbundle_name': bundle['csvName'],
+                    'operatorbundle_version': bundle['version'],
+                    'operatorbundle_path': bundle['bundlePath'],
+                }
+                _add_property_to_index(db_path, max_openshift_version_property)
+                log.info('property added for %s', bundle['bundlePath'])
+
+    # for bundle in unique_present_bundles:
```
yashvardhannanavati, line 119: Could you please remove this commented code?

iib/workers/tasks/utils.py
yashvardhannanavati: I don't think we get duplicate bundles when fetching bundles from the index. Because OPM doesn't allow adding duplicates in the first place.

yashvardhannanavati (changes requested):

pbortlov:
yes, @yashvardhannanavati  it does. Look at those logs 'Raw bundles': https://issues.redhat.com/secure/attachment/12670071/12670071_iib-logs-15.txt

iib/workers/tasks/utils.py
lipoja: @xDaile ^^

yashvardhannanavati:
I wonder what happened to the associated ticket. I am going to close this PR as it is stale. Please open a new one if this is still required.

PR #345, opened 2021-12-16
Add QE tag when building IIB API image in github workflow

yashvardhannanavati:


pbortlov (approved):

.github/workflows/build.yml
```diff
@@ -30,7 +30,7 @@ jobs:
         uses: redhat-actions/buildah-build@v2
         with:
           image: iib-api
-          tags: ${{ github.event.release.tag_name }} latest
+          tags: ${{ github.event.release.tag_name }} latest qe
```
lipoja, line 33: But this means that it will push iib-api image with three tags (v5.0.0, lates, qe) ... 
If you want to build just QE we should modify this file. 
Because if you trigger this action, it will build and push image and once it is public and it will be used by other team.
If the image is tested then I agree.  But it looks like you want to used it for QE deployment.

yashvardhannanavati: Good point. I added another workflow which will build and push on tag push to the repo

JAVGan (approved):
LGTM.

PR #349, opened 2022-01-05
Update Celery to 5.2.2 for Python 3.8

JAVGan:
Update Celery to remediate CVE-2021-23727 (Stored Command Injection security vulnerability).

### DIFF
--- a/requirements-py38.txt
+++ b/requirements-py38.txt
@@ -25,9 +25,9 @@ botocore==1.23.23 \
     # via
     #   boto3
     #   s3transfer
-celery==5.2.1 \
-    --hash=sha256:b41a590b49caf8e6498a57db628e580d5f8dc6febda0f42de5d783aed5b7f808 \
-    --hash=sha256:cc63ea6572d558be65297ba6db7a7979e64c0a3d0d61212d6302ef1ca05a0d22
+celery==5.2.2 \
+    --hash=sha256:2844eb040e915398623a43253a8e1016723442ece6b0751a3c416d8a2b34216f \
+    --hash=sha256:5a68a351076cfac4f678fa5ffd898105c28825a2224902da006970005196d061
     # via iib (setup.py)
 certifi==2019.11.28 \
     --hash=sha256:017c25db2a153ce562900032d5bc68e9f191e44e9a0f762f373977de9df1fbb3 \
--- a/requirements-test-py38.txt
+++ b/requirements-test-py38.txt
@@ -36,9 +36,9 @@ botocore==1.23.23 \
     #   -r requirements-py38.txt
     #   boto3
     #   s3transfer
-celery==5.2.1 \
-    --hash=sha256:b41a590b49caf8e6498a57db628e580d5f8dc6febda0f42de5d783aed5b7f808 \
-    --hash=sha256:cc63ea6572d558be65297ba6db7a7979e64c0a3d0d61212d6302ef1ca05a0d22
+celery==5.2.2 \
+    --hash=sha256:2844eb040e915398623a43253a8e1016723442ece6b0751a3c416d8a2b34216f \
+    --hash=sha256:5a68a351076cfac4f678fa5ffd898105c28825a2224902da006970005196d061
     # via -r requirements-py38.txt
 certifi==2019.11.28 \
     --hash=sha256:017c25db2a153ce562900032d5bc68e9f191e44e9a0f762f373977de9df1fbb3 \


requirements-py38.txt
```diff
@@ -401,6 +401,8 @@ werkzeug==2.0.2 \
     --hash=sha256:aa2bb6fc8dee8d6c504c0ac1e7f5f7dc5810a9903e793b6f715a9f015bdadb9a
     # via flask
 
-# WARNING: The following packages were not pinned, but pip requires them to be
-# pinned when the requirements file includes hashes. Consider using the --allow-unsafe flag.
-# setuptools
+# The following packages are considered to be unsafe in a requirements file:
```
pbortlov, line 404: this can be removed.

requirements-py38.txt
```diff
@@ -2,14 +2,14 @@
 # This file is autogenerated by pip-compile with python 3.8
 # To update, run:
 #
-#    pip-compile --generate-hashes --output-file=requirements-py38.txt
+#    pip-compile --allow-unsafe --generate-hashes --output-file=requirements-py38.txt
```
pbortlov, line 5: we shouldn't allow unsafe

requirements-test-py38.txt
```diff
@@ -537,6 +537,10 @@ werkzeug==2.0.2 \
     #   -r requirements-py38.txt
     #   flask
 
-# WARNING: The following packages were not pinned, but pip requires them to be
-# pinned when the requirements file includes hashes. Consider using the --allow-unsafe flag.
-# setuptools
+# The following packages are considered to be unsafe in a requirements file:
```
pbortlov, line 540: Let's dig deep why and not add --allow-unsafe

JAVGan: Without `--allow-unsafe` pip gives me this error while building:

```
[...]
Collecting werkzeug==2.0.2
  Using cached Werkzeug-2.0.2-py3-none-any.whl (288 kB)
Collecting setuptools<59.7.0,>=59.1.1
ERROR: In --require-hashes mode, all requirements must have their versions pinned with ==. These do not:
    setuptools<59.7.0,>=59.1.1 from https://files.pythonhosted.org/packages/b0/3a/88b210db68e56854d0bcf4b38e165e03be377e13907746f825790f3df5bf/setuptools-59.6.0-py3-none-any.whl#sha256=4ce92f1e1f8f01233ee9952c04f6b81d1e02939d6e1b488428154974a4d0783e (from celery==5.2.3->-r requirements-test-py38.txt (line 39))
[...]
ERROR:   py38: could not install deps [-rrequirements-test-py38.txt]; v = InvocationError('/home/jgangi/Documents/RedHat/exd-sp/cloud-dist/iib/.tox/py38/bin/python -m pip install -rrequirements-test-py38.txt', 1)
```

requirements-py38.txt
JAVGan: I agree, but it's not building.. Still didn't find a way to include `setuptools` without it, unless I do it manually....

requirements-py38.txt
JAVGan: Sure, it's automatically generated but can be removed manually

requirements-test-py38.txt
JAVGan: I figured it out, let's see if this build passes

requirements-py38.txt
JAVGan: No longer needed, fix in new version

requirements-test-py38.txt
JAVGan: From `pip-compile --help` I found this:

```
--allow-unsafe / --no-allow-unsafe
                                  Pin packages considered unsafe: distribute,
                                  pip, setuptools.
                                  
                                  WARNING: Future versions of pip-tools will
                                  enable this behavior by default. Use --no-
                                  allow-unsafe to keep the old behavior. It is
                                  recommended to pass the --allow-unsafe now
                                  to adapt to the upcoming change.

```

Basically the "unsafe" packages are `pip`, `setuptools` and `distribute` but this new version of celery requires it.

I'll upload a new version, removing the comments, but since this behavior will be enabled in future versions it don't seem to be "unsafe" by me.

@yashvardhannanavati @chandwanitulsi @pbortlov what do you think?

JAVGan: **Note:** I've sent a new push generated with `--allow-unsafe` and manually edited to remove this line from header after doing `pip-compile`

requirements-py38.txt
JAVGan: [PTAL](https://app.travis-ci.com/github/release-engineering/iib/jobs/554399473) on Travis without `--allow-unsafe`, it is, without `setuptools`

requirements-test-py38.txt
chandwanitulsi: FWIW, I think its better to have the requirements*.txt file be auto generated instead of having that manually changed. If we do want to have it manually edited, how about adding a comment near setuptools indicating the same

JAVGan: Good point!

So what should we do next?

1. Auto-generate the file and let it be with `--allow-unsafe`
2. Generate the file and edit manually leaving a comment near `setuptools`
3. Skip this update and let security complain for Celery

As I saw in the `--help` and other sources it's not possible to include `setuptools` on `requirements*` without `--allow-unsafe` by using `pip-compile`

requirements-py38.txt
pbortlov: We can try to pin setuptools in iib-playbooks

JAVGan:
Updated the requirements for Python3.8 using the following commands:

```{bash}
make build
make up
podman exec iib_iib-worker_1 pip install --upgrade pip pip-tools
podman exec iib_iib-worker_1 pip-compile --generate-hashes --output-file=requirements-py38.txt --upgrade-package "celery"
podman exec iib_iib-worker_1 pip-compile --generate-hashes --output-file=requirements-test-py38.txt requirements-test-py38.in
make test
```

JAVGan:
@pbortlov @chandwanitulsi @yashvardhannanavati [PTAL on this](https://docs.celeryproject.org/projects/kombu/en/stable/changelog.html#version-5-2-3).

It seems Celery is **really** pinning `setuptools` and without it our `requirements-py38*` always fails because it's on `require-hashes` mode.

The only (and ugly) solution I see is so far is to `pip-compile --allow-unsafe [...]`

 I've tried different things so and nothing work.

What do you think?

JAVGan:
> @pbortlov @chandwanitulsi @yashvardhannanavati [PTAL on this](https://docs.celeryproject.org/en/v5.2.3/changelog.html#version-5-2-3).
> 
> It seems Celery is **really** pinning `setuptools` and without it our `requirements-py38*` always fails because it's on `require-hashes` mode.
> 
> The only (and ugly) solution I see is so far is to `pip-compile --allow-unsafe [...]`
> 
> I've tried different things and nothing works.
> 
> What do you think?

**Nevermind...**

By looking at a bit low in the [link](https://docs.celeryproject.org/en/v5.2.3/changelog.html#version-5-2-2) I figured it out I could just upgrade to `5.2.2` instead of `5.2.3` to solve the CVE and avoid all this headache of `setuptools`.

The [tests have passed](https://app.travis-ci.com/github/release-engineering/iib/jobs/554733399), I'll update the PR title.

JAVGan:
Update commands:

```
make build
make up
podman exec iib_iib-worker_1 pip install --upgrade pip pip-tools
podman exec iib_iib-worker_1 pip-compile --generate-hashes --output-file=requirements-py38.txt --upgrade-package "celery==5.2.2"
podman exec iib_iib-worker_1 pip-compile --generate-hashes --output-file=requirements-test-py38.txt requirements-test-py38.in
make test
```

lipoja (approved):

pbortlov (approved):

PR #353, opened 2022-01-12
Static Application Security Testing

JAVGan:
# Static Application Security Testing

This PR includes `bandit` on `tox.ini` file to perform SAST in our CI

## Implementation details
* The new file `.bandit` contains the main configuration for it. In our case we're just excluding some unwanted directories.
* Among the excluded directories we have `/tests` and `/docker` since:
  - The main goal of SAST is to scan the application, not its tests or Dockerfile(s).
  - The `/tests` dir contains many functions with  `assert` keys and other stuff that get incorrectly caught by `bandit` as vulnerabilities thus even the [project's main page](https://github.com/PyCQA/bandit#per-project-command-line-args) suggests to not scan the tests directory.
* The `bandit` scan was set to just trigger failures for severity **Medium or higher**
* There are no tests skipped nor any additional plugins besides the defaults which comes with `bandit` from `pip`

## Notes
1. The following "problems" were caught by `bandit` and we need to discuss whether they should be just ignored or treated:

```
Test results:
>> Issue: [B102:exec_used] Use of exec detected.
   Severity: Medium   Confidence: High
   Location: ./iib/workers/config.py:171:12
   More Info: https://bandit.readthedocs.io/en/latest/plugins/b102_exec_used.html
170	        with open(prod_config_file_path, mode='rb') as config_file:
171	            exec(compile(config_file.read(), prod_config_file_path, 'exec'), _user_config)
172	

--------------------------------------------------
>> Issue: [B103:set_bad_file_permissions] Chmod setting a permissive mask 0o775 on file (log_file_path).
   Severity: Medium   Confidence: High
   Location: ./iib/workers/tasks/utils.py:700:12
   More Info: https://bandit.readthedocs.io/en/latest/plugins/b103_set_bad_file_permissions.html
699	            request_log_handler.setFormatter(log_formatter)
700	            os.chmod(log_file_path, 0o775)

```
  
Refers to CLOUDDST-10707

PR #355, opened 2022-01-14
Drop support for Python3.6

JAVGan:
# Drop support for Python 3.6
This PR completely removes the Python 3.6 dependencies from IIB and uses Python 3.8 as default.

## Requirements files
The first goal was to replace the files `requirements.txt` and `requirements-test.txt` with the content of `requirements-py38.txt` and `requirements-test-py38.txt` respectively.

In order to do that the following steps were executed:

* Run `pip-compile` using a `virtualenv` with Python 3.8 to regenerate the `requirements(-test).txt` files
* Make sure the generated files has the same content as `requirements(-test)-py38.txt`
* Remove the files requirements(-test)-py38.txt (since the content is the same) and create one symlink  `requirements-py38.txt -> requirements.txt` to not break the ansible-role `iib-playbook`.

Then, we have the files `requirements.txt`, `requirements-test.txt` and `requirements-test.in` with Python 3.8 content and one symlink for the sake of compatibility.

## Tox
All references for Python 3.6 were removed from `tox.ini` and `safety`env was brought back to be just one, since we no longer need to keep an alternate environment for it.

## Other files
Besides the `requirements*` and `tox.ini` the following files were changed to remove Python 3.6:

* `.python-version`
* `.travis.yml`
* `pyproject.toml`
* `setup.py`

# Temporary fix for Travis CI
Alongside Python3.8 as default this PR includes a workaround using `# nosec` to avoid `bandit` errors in CI since we started using SAST on #353 .

These issues pointed by `bandit` will be properly addressed in future

# Notes
* The symlink can be safely removed once its references in ansible-roles get changed back to `requirements.txt`.
* Python 3.8 is the default version for IIB. 
* I was going to add support to Python 3.9 but it require to update `black` from `19.10b0` to at least `21.4b0` and then it starts complaining about a lot of stuff in this [new version](https://pypi.org/project/black/21.4b0/), so I skipped it, at least in this PR.
 

This PR refers to CLOUDDST-10861


midnightercz (dismissed):

iib/workers/tasks/utils.py
```diff
@@ -697,7 +697,8 @@ def wrapper(*args, **kwargs):
             request_log_handler = logging.FileHandler(log_file_path)
             request_log_handler.setLevel(log_level)
             request_log_handler.setFormatter(log_formatter)
-            os.chmod(log_file_path, 0o775)
+            # FIXME: Bandit complaining on too permissive logs - CLOUDDST-11307
```
pbortlov, line 700: Do we have a task to fix this?

requirements-test.txt
```diff
@@ -1,5 +1,5 @@
 #
-# This file is autogenerated by pip-compile with python 3.6
+# This file is autogenerated by pip-compile with python 3.8
 # To update, run:
 #
 #    pip-compile --generate-hashes --output-file=requirements-test.txt requirements-test.in
```
pbortlov, line 5: `requirements-test-py38.txt` and `requirements-test-py38.in`

pbortlov (changes requested):
It is a little bit confusing why it seems that you removed requirements-test-py38.in and requirements-test-py38.txt and readded. I do miss those lines 
>>> -r requirements-py38.txt
>>> coverage
>>> pytest
>>> pytest-cov

in `requirements-test-py38.in`

iib/workers/tasks/utils.py
JAVGan: Yes, the CLOUDDST-11307

requirements-test.txt
JAVGan: `requirements-test-py38.txt` is now a link to `requirements-test.txt` thus `requirements-test-py38.in` is no longer needed. In fact it's just another link to `requirements-test.in`

JAVGan:
> It is a little bit confusing why it seems that you removed requirements-test-py38.in and requirements-test-py38.txt and readded. I do miss those lines
> 
> > > > -r requirements-py38.txt
> > > > coverage
> > > > pytest
> > > > pytest-cov
> 
> in `requirements-test-py38.in`

Basically the whole idea is to use just `requirements.txt`, `requirements-test.txt` and `requirements-test.in` with Python3.8. However, since ansible roles is using now the files with suffix `*-py38*` I transformed them in symlinks to the original files. Once we revert the ansible roles to the proper files we can remove the symlinks, they have no use since we don't have a separation of Python envs anymore.

requirements-test.txt
JAVGan: If you prefer I can remove the `requirements-test-py38.in` since it has complete no use

JAVGan: To better explain what's going on:

The idea is to replace `requirements-py38[-test]*` with `requirements[-test]*` since we're dropping Python 3.6
This could be simply achieved by:
```{bash}
mv requirements-py38.txt requirements.txt
mv requirements-py38.test.txt requirements-test.txt
mv requirements-py38.test.in requirements.test.in
```

**However**, there are 2 problems:
1. if we do that we'll break `iib-playbook` because it's now expecting `requirements-py38.txt` instead of `requirements.txt`, thus I've created symlinks pointing these `-py38` files to the regular ones.
2. the content inside `requirements-py38*` refers to its names and should be changed to the regular files (without suffix `-py38`

That's why I've recompiled the files and created a symlink to them.

But thinking a bit more I figured out your point, it doesn't make sense to keep the files `requirements-py38-test*` because they're just used by DEV env, so I can safely remove them and it will break nothing.

**What I'll do:** I'll send a patched version without the symlinks for `*-py38-test*` since we're not using them and let the link for `requirements-py38.txt -> requirements.txt` exist until we put it back `requirements.txt` in `iib-playbook`

pbortlov (approved):
Please update the description of the PR about removing py38 requirements

PR #356, opened 2022-01-17
Add support for Python3.9

JAVGan:
# Add support for Python 3.9

This PR is a subsequent implementation of #355 and should be only merged **after** it.

It includes the support for Python3.9 as well as keep Python3.8 as default for IIB.

## Black update issues
The code-formatter `black` just support Python 3.9 from version [21.4b0](https://pypi.org/project/black/21.4b0/) and above. 

However, after upgrading `black` from `19.10b0` to `21.4b0` it complained about the formatting in our code, thus I had to let it reformat a series of files to be compliant with its new exigencies.

Finally `black` was updated and `py39` is fully supported.

## Changes
The following files were changed to support Python 3.9:
  * `.travis.yml`
  * `pyproject.toml`
  * `setup.py`
  * `tox.ini`
  
**All other files were just reformatted to be compliant with the new `black` constraints.**

## Observations
* This PR doesn't change any functionality on IIB, just reformat files and allow the use of Python 3.9
* The latest version of Python is 3.10 however this version is not supported on RHEL8 and RHEL9 beta uses 3.9 as default, thus the inclusion of 3.9 instead of 3.10

Refers to CLOUDDST-11351

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -28,7 +28,7 @@ pytest_command =
 description = black checks [Mandatory]
 skip_install = true
 deps =
-    black==19.10b0
+    black==21.4b0
 commands =
     black --check --diff iib tests
 
--- a/tox.ini
+++ b/tox.ini
@@ -1,16 +1,18 @@
 [tox]
 skip_missing_interpreters = true
-envlist = black,docs,flake8,py38,safety,yamllint,bandit
+envlist = black,docs,flake8,py38,py39,safety,yamllint,bandit
 downloadcache = {toxworkdir}/_download/
 
 [testenv]
 usedevelop = true
 # 3.8 is the current suppported version
+# 3.9 is the latest version available on RHEL8
 basepython =
     black: python3.8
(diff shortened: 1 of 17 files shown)


tox.ini
```diff
@@ -26,7 +28,7 @@ pytest_command =
 description = black checks [Mandatory]
 skip_install = true
 deps =
-    black==19.10b0
+    black==21.4b0
```
lipoja, line 31: Suggestion - separate PR or at least separate commit for bumping Black version - one can be lost in all those format changes which might not be relevant to the topic "Add support for Python3.9"

lipoja (changes requested):
LGTM, just separate PR/commit for "black magic"

tox.ini
JAVGan: Roger :+1:

JAVGan: Addressed

lipoja (approved):

PR #357, opened 2022-01-18
Bugfix for Buildah retries on 50X

JAVGan:
# Bugfix for Buildah retries on 50X
This PR fixes issues with the implementation delivered by #335 : 
* The regex wasn't being correctly evaluated since `re.match()` doesn't support parsing multiple lines by default.
    * This was fixed using the very same loop from the previous lines thus I've created a separate function for that to avoid code duplication. 
* `ExternalServiceError` can't inherit `IIBError` since it will be caught first by `iib.workers.tasks.general.failed_request_callback` and the retry will fail
   * This was fixed by letting `ExternalServiceError` inherit from `BaseException`
   
* New test for Buildah retry
   * The test `test_run_cmd_failed_buildah` complements `test_buildah_fail_max_retries`

Bug found [here](https://iib.stage.engineering.redhat.com/api/v1/builds/56056/logs) and [here](https://iib.stage.engineering.redhat.com/api/v1/builds/56057/logs).

Refers to CLOUDDST-10989

### DIFF
--- a/iib/exceptions.py
+++ b/iib/exceptions.py
@@ -21,5 +21,5 @@ class AddressAlreadyInUse(BaseException):
     """Adress is already used by other service."""
 
 
-class ExternalServiceError(IIBError):
+class ExternalServiceError(BaseException):
     """An external service error occurred with HTTP 50X."""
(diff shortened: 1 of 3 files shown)


lipoja (dismissed):

lipoja:
@JAVGan Let's rebase this

JAVGan:
> @JAVGan Let's rebase this

Rebased

lipoja (approved):

PR #358, opened 2022-01-19
Fix loosing olm.maxOpenShiftVersion property

lipoja:
The issue is that after adding olm.maxOpenShiftVersion property
directly to index.db we did not build new image.
When merge API was called and deprecation_bundles were set
the function deprecate_bundles worked with intermediate image
that was pushed to external registry. Unfortunately it does not
contain olm.maxOpenShiftVersion since no image with this change
was never built and pushed to that registry.

Fix - build and push image with olm.maxOpenShiftVersion property

### DIFF
--- a/iib/workers/tasks/build_merge_index_image.py
+++ b/iib/workers/tasks/build_merge_index_image.py
@@ -232,6 +232,7 @@ def handle_merge_request(
         missing_bundle_paths = [bundle['bundlePath'] for bundle in missing_bundles]
         if missing_bundle_paths:
             add_max_ocp_version_property(missing_bundle_paths, temp_dir)
+
         set_request_state(request_id, 'in_progress', 'Deprecating bundles in the deprecation list')
         log.info('Deprecating bundles in the deprecation list')
         intermediate_bundles = missing_bundle_paths + source_index_bundles_pull_spec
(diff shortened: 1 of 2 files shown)


iib/workers/tasks/build_merge_index_image.py
```diff
@@ -232,6 +232,12 @@ def handle_merge_request(
         missing_bundle_paths = [bundle['bundlePath'] for bundle in missing_bundles]
         if missing_bundle_paths:
             add_max_ocp_version_property(missing_bundle_paths, temp_dir)
+            # we have to build and push image to external registry because we have to use it
+            # when we deprecate bundles
+            # Note: since we modified only index.db LABELs are alraedy present in Dockerfile
```
pbortlov, line 237: TYPO: alraedy

pbortlov: So you say that because we didn't modify LABELs but how is it connected to index.db?

lipoja: I mean when we build the image we add LABELs to Dockerfile. 
The labels are  'com.redhat.index.delivery.version' and 'com.redhat.index.delivery.distribution_scope'.

And since we did not generated new docker file we do not have to add those labels as we do before every build.

JAVGan (changes requested):
`test_handle_merge_request` is failing

JAVGan (dismissed):

lipoja:
@JAVGan I move it to different position so we do not build and push image when it is not needed. Could you check it again, please?

JAVGan (approved):

JAVGan:
> @JAVGan I move it to different position so we do not build and push image when it is not needed. Could you check it again, please?

LGTM, probably need a rebase but it's all good!

iib/workers/tasks/build_merge_index_image.py
```diff
@@ -244,11 +245,17 @@ def handle_merge_request(
         deprecation_bundles = deprecation_bundles + [
             bundle['bundlePath'] for bundle in invalid_version_bundles
         ]
-        intermediate_image_name = _get_external_arch_pull_spec(
-            request_id, arch, include_transport=False
-        )
 
         if deprecation_bundles:
+            # opm can only deprecate a bundle image on an existing index image. Build and
```
pbortlov, line 250: could you add the comment that the index image has to be pushed to the registry and we CAN'T use it from internal registry?

pbortlov (approved):
Please add a comment otherwise it looks good

iib/workers/tasks/build_merge_index_image.py
lipoja: added

pbortlov (approved):

PR #359, opened 2022-01-19
Workaround for failing `opm index deprecatetruncate`

lipoja:
`opm index deprecatetruncate` does not work correctly with local image
We can use external registry to temporary store index image from which
we want to deprecate bundles.

Note: external registry is used in merge index API endpoint

Might be workaround for CLOUDDST-11326

### DIFF
--- a/iib/workers/tasks/build.py
+++ b/iib/workers/tasks/build.py
@@ -820,8 +820,9 @@ def handle_add_request(
                 arch = sorted(arches)[0]
                 log.info('Building a temporary index image to satisfy the deprecation requirement')
                 _build_image(temp_dir, 'index.Dockerfile', request_id, arch)
-                intermediate_image_name = _get_local_pull_spec(
-                    request_id, arch, include_transport=True
+                _push_image(request_id, arch)
+                intermediate_image_name = _get_external_arch_pull_spec(
+                    request_id, arch, include_transport=False
                 )
                 deprecate_bundles(
                     bundles=deprecation_bundles,
--- a/tests/test_workers/test_tasks/test_build.py
+++ b/tests/test_workers/test_tasks/test_build.py
@@ -672,10 +672,10 @@ def test_handle_add_request(
     if deprecate_bundles:
         # Take into account the temporarily created index image
         assert mock_bi.call_count == len(arches) + 1
+        assert mock_pi.call_count == len(arches) + 1
     else:
         assert mock_bi.call_count == len(arches)
-
-    assert mock_pi.call_count == len(arches)
+        assert mock_pi.call_count == len(arches)
 
     mock_uiips.assert_called_once()
     mock_vii.assert_not_called()
@@ -686,7 +686,7 @@ def test_handle_add_request(
             bundles=['random_bundle@sha', 'some-deprecation-bundle@sha'],
             base_dir=mock.ANY,
             binary_image=binary_image or 'some_image',
-            from_index='containers-storage:localhost/iib-build:3-amd64',
+            from_index='registry:8443/iib-build:3-amd64',
             overwrite_target_index_token=None,
             container_tool='podman',
         )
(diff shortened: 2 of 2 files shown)


iib/workers/tasks/build.py
```diff
@@ -816,7 +816,10 @@ def handle_add_request(
             arch = sorted(arches)[0]
             log.info('Building a temporary index image to satisfy the deprecation requirement')
             _build_image(temp_dir, 'index.Dockerfile', request_id, arch)
-            intermediate_image_name = _get_local_pull_spec(request_id, arch, include_transport=True)
+            _push_image(request_id, arch)
```
JAVGan, line 819: Thanks for the note of `merge-index-image`, it uses the same logic.

Now I've 2 questions:
1. Do we have a test for it somewhere to make sure it works?
2. Do we need or have a way to clean-up the registry to not let it store the intermediate images forever?

lipoja: 1. iib-api-test should cover this on any instance (but I guess they are failing right now) - maybe because of this issue. 
2. no - it should be overwritten later by push to the same pull spec

JAVGan (changes requested):
Tests failing

JAVGan:
> We should test it - it might not be related to that error

Cool, I'll keep watching it and if it solves the issue I can approve it.

lipoja:
> > We should test it - it might not be related to that error
> 
> Cool, I'll keep watching it and if it solves the issue I can approve it.

My local tests showed that it might solve the issue - I am waiting for @chandwanitulsi  doublecheck

chandwanitulsi:
I tested this change with the following command, the request completed successfully
`curl -u: --negotiate -X POST -H "Content-Type: application/json" http://localhost:8080/api/v1/builds/add -d '{"bundles": [ "registry-proxy.engineering.redhat.com/rh-osbs/amq7-amq-streams-rhel7-operator-metadata:1.0.0-1","registry-proxy.engineering.redhat.com/rh-osbs/amq7-amq-streams-rhel7-operator-metadata:1.1.0-1", "registry-proxy.engineering.redhat.com/rh-osbs/amq7-amq-streams-rhel7-operator-metadata:1.2.0-1"] , "binary_image": "quay.io/operator-framework/upstream-registry-builder:v1.5.9","from_index": "registry-proxy-stage.engineering.redhat.com/rh-osbs-stage/iib-pub-pending-qa:apitestoriginal", "add_arches":["amd64"], "deprecation_list":["registry-proxy.engineering.redhat.com/rh-osbs/amq7-amq-streams-rhel7-operator-metadata:1.0.0-1"]}'
`

Before this PR, the request would fail with `Failed to deprecate the bundles: error copying container directory Error: container \"Getting image source signatures\\nCopying blob sha256\" does not exist` error

chandwanitulsi (dismissed):
LGTM

JAVGan (dismissed):

JAVGan (dismissed):

chandwanitulsi (dismissed):

iib/workers/tasks/build.py
```diff
@@ -820,8 +820,9 @@ def handle_add_request(
                 arch = sorted(arches)[0]
                 log.info('Building a temporary index image to satisfy the deprecation requirement')
                 _build_image(temp_dir, 'index.Dockerfile', request_id, arch)
-                intermediate_image_name = _get_local_pull_spec(
-                    request_id, arch, include_transport=True
+                _push_image(request_id, arch)
+                intermediate_image_name = _get_external_arch_pull_spec(
+                    request_id, arch, include_transport=False
```
yashvardhannanavati, line 825: As long as you are pushing the image with the arch suffix, I believe it should be ok

lipoja: Thank you!

JAVGan (approved):

chandwanitulsi (approved):

PR #383, opened 2022-01-24
Use set_registry_token before checking index_image

lipoja:
There was missing _set_registry_token (adding credentials from payload `overwrite_from_index_token`) when accessing image in private registry. 

CLOUDDST-11393

### DIFF
--- a/iib/workers/tasks/build_merge_index_image.py
+++ b/iib/workers/tasks/build_merge_index_image.py
@@ -126,7 +126,8 @@ def _add_bundles_missing_in_source(
             '%s bundles have invalid version label and will be deprecated.', len(invalid_bundles)
         )
 
-    is_source_fbc = is_image_fbc(source_from_index)
+    with set_registry_token(overwrite_target_index_token, source_from_index):
+        is_source_fbc = is_image_fbc(source_from_index)
     if is_source_fbc:
         opm_registry_add_fbc(
             base_dir=base_dir,
@@ -134,6 +135,7 @@ def _add_bundles_missing_in_source(
             binary_image=binary_image,
             from_index=source_from_index,
             container_tool='podman',
+            overwrite_from_index_token=overwrite_target_index_token,
         )
     else:
         _opm_index_add(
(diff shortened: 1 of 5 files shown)


iib/workers/tasks/build_merge_index_image.py
```diff
@@ -215,8 +217,9 @@ def handle_merge_request(
     dockerfile_name = 'index.Dockerfile'
 
     with tempfile.TemporaryDirectory(prefix='iib-') as temp_dir:
-        source_fbc = is_image_fbc(source_from_index_resolved)
-        target_fbc = is_image_fbc(target_index_resolved)
+        with set_registry_token(overwrite_target_index_token, source_from_index):
```
lipoja, line 220: This is actually the fix it self - once image is pulled it will not fail - but for safety reasons I added set_registry_token when accessing the image.

iib/workers/tasks/opm_operations.py
```diff
@@ -194,7 +194,9 @@ def _serve_cmd_at_port(serve_cmd, cwd, port, max_tries, wait_time):
     raise IIBError(f'Index registry has not been initialized after {max_tries} tries')
 
 
-def _get_or_create_temp_index_db_file(base_dir, from_index=None):
+def _get_or_create_temp_index_db_file(
+    base_dir, from_index=None, overwrite_from_index_token=None, ignore_existing=False
+):
```
chandwanitulsi, line 199: Can you explain the two cases for ignore_existing? we are using it only at one place with true

lipoja: I did not want to change behavior so I added this parameter - I thought it can be used in other functions but checking it now it does not seems like it.

Do you agree with this or should I change it back?

chandwanitulsi: yeah, if we don't see a future scenario where both values would be useful, I think we should stick to one

yashvardhannanavati (commented):
+1 once Tulsi's comment is addressed

chandwanitulsi (approved):

PR #386, opened 2022-02-01
Enable send events to the broker

zanssa:
For ref, CLOUDDST-2704, and README file for the celery role.


chandwanitulsi (approved):
LGTM :+1: 
Adding https://docs.celeryproject.org/en/latest/userguide/configuration.html#std-setting-task_send_sent_event for reference

lipoja (approved):

PR #391, opened 2022-03-09
Use terminate function to shut down a process instead of kill

yashvardhannanavati:
kill function sends the signal signal.SIGKILL because of which
OPM doesn't cleanup the temp directory after a request causing
memory issues in IIB. Replacing it with terminate function instead
sends signal.SIGTERM signal thus causing OPM to cleanup after the
request.

Refers to CLOUDDST-12129


lipoja (approved):

lslebodn:
IMHO you should try SIGTERM and if it does not work than you should fall back to SIGKILL.
Better to theoretically leak data in `/tmp` than having zombie processes on system.

PR #403, opened 2022-05-16
Create empty index endpoint does not accept build_tags.

lipoja:
Removind help text for build_tags since it is not accepted by create-empty-index endpoint


JAVGan (approved):
LGTM

pbortlov (approved):

PR #407, opened 2022-06-08
Log when greenwave is not configured

JAVGan:
Add a message to logs to confirm that greenwave gating is disabled.

Refers to CLOUDDST-12595

### DIFF
--- a/iib/workers/tasks/build.py
+++ b/iib/workers/tasks/build.py
@@ -724,6 +724,8 @@ def handle_add_request(
     # Check if Gating passes for all the bundles
     if greenwave_config:
         gate_bundles(resolved_bundles, greenwave_config)
+    else:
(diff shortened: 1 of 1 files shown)


chandwanitulsi (approved):

iib/workers/tasks/build.py
```diff
@@ -724,6 +724,8 @@ def handle_add_request(
     # Check if Gating passes for all the bundles
     if greenwave_config:
         gate_bundles(resolved_bundles, greenwave_config)
+    else:
+        log.warning('Greenwave checks are not active and will be disabled in next release.')
```
yashvardhannanavati, line 728: Should we reword this to something like `Greenwave checks are disabled. Bundles will not be gated`

PR #410, opened 2022-06-30
Adding fallback from SIGTERM to SIGKILL

lipoja:
[CLOUDDST-12159]

### DIFF
--- a/iib/workers/tasks/utils.py
+++ b/iib/workers/tasks/utils.py
@@ -83,7 +83,7 @@ def add_max_ocp_version_property(resolved_bundles, temp_dir):
         ['grpcurl', '-plaintext', f'localhost:{port}', 'api.Registry/ListBundles'],
         exc_msg='Failed to get bundle data from index image',
     )
-    rpc_proc.terminate()
+    terminate_process(rpc_proc)
 
     # This branch is hit when `bundles` attribute is empty and the index image is empty.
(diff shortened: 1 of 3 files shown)


iib/workers/tasks/utils.py
```diff
@@ -701,6 +701,23 @@ def run_cmd(
...
+    """
+    Terminate given process. Fallback to SIGKILL when process is not terminated in given timeout.
+
+    :param subprocess.Popen proc: process to be terminated
+    :param int timeout: number of seconds to wait for terminating process
+    """
+    log.debug('Terminating process %s', proc)
+    proc.terminate()
+    try:
+        proc.communicate(timeout=timeout)
```
JAVGan, line 714: Just a small comment, I was going to ask why did you use `proc.communicate()` here instead of `proc.wait()` and I see that's because of the possible deadlock when using PIPE for STDIN and STDOUT as mentioned [here](https://docs.python.org/3/library/subprocess.html#subprocess.Popen.wait).


However I think it might be useful for maintainability to add a comment above explaining the use of `proc.communicate()` to avoid the deadlock, WDYT?

lipoja: @JAVGan Yeah, why not. Comment added.

JAVGan (approved):
LGTM

yashvardhannanavati (approved):

PR #412, opened 2022-07-07
Add static types to iib/workers

xDaile:
CLOUDDST-14124

Static types to methods in ./iib/workers were added
Check in the tox file was configured also
Few type inconsistencies were repaired

### DIFF
--- a/iib/workers/api_utils.py
+++ b/iib/workers/api_utils.py
@@ -1,17 +1,19 @@
 # SPDX-License-Identifier: GPL-3.0-or-later
 import json
 import logging
+from typing import Any, Dict, Optional
 
 import requests
 from requests.packages.urllib3.util.retry import Retry
 import requests_kerberos
 
 from iib.exceptions import IIBError
+from iib.workers.tasks.iib_static_types import UpdateRequestPayload
 
 log = logging.getLogger(__name__)
 
 
-def get_requests_session(auth=False):
+def get_requests_session(auth: bool = False) -> requests.Session:
     """
     Create a requests session with authentication (when enabled).
 
@@ -33,7 +35,7 @@ def get_requests_session(auth=False):
     return session
 
 
-def get_request(request_id):
+def get_request(request_id: int) -> Dict[str, Any]:
     """
     Get the IIB build request from the REST API.
 
@@ -68,7 +70,7 @@ def get_request(request_id):
     return rv.json()
 
 
-def set_request_state(request_id, state, state_reason):
+def set_request_state(request_id: int, state: str, state_reason: str) -> Dict[str, Any]:
     """
     Set the state of the request using the IIB API.
 
@@ -85,12 +87,15 @@ def set_request_state(request_id, state, state_reason):
         state,
         state_reason,
     )
-    payload = {'state': state, 'state_reason': state_reason}
+    payload: UpdateRequestPayload = {'state': state, 'state_reason': state_reason}
     exc_msg = 'Setting the state to "{state}" on request {request_id} failed'
     return update_request(request_id, payload, exc_msg=exc_msg)
 
 
-def set_omps_operator_version(request_id, omps_operator_version):
+def set_omps_operator_version(
+    request_id: int,
+    omps_operator_version: Dict[str, str],
+) -> Dict[str, Any]:
     """
     Set the set_omps_operator_version of the request using the IIB API.
 
@@ -106,13 +111,17 @@ def set_omps_operator_version(request_id, omps_operator_version):
         request_id,
         omps_operator_version_json,
     )
-    payload = {'omps_operator_version': omps_operator_version_json}
+    payload: UpdateRequestPayload = {'omps_operator_version': omps_operator_version_json}
     exc_msg = 'Setting the omps_operator_version to "{omps_operator_version}" failed'
 
     return update_request(request_id, payload, exc_msg=exc_msg)
 
 
-def update_request(request_id, payload, exc_msg=None):
+def update_request(
+    request_id: int,
+    payload: UpdateRequestPayload,
+    exc_msg: Optional[str] = None,
+) -> Dict[str, Any]:
     """
     Update the IIB build request.
 
--- a/iib/workers/config.py
+++ b/iib/workers/config.py
@@ -2,50 +2,62 @@
 import os
 import logging
 import types
+from typing import Any, Dict, List, Optional, Type, Union
+
+from celery import Celery, app
 
 from iib.exceptions import ConfigError
+from iib.workers.tasks.iib_static_types import (
+    IIBOrganizationCustomizations,
+    CSVAnnotations,
+    PackageNameSuffix,
+    ImageNameFromLabels,
+    RegistryReplacements,
+    EncloseRepo,
+    iib_organization_customizations_type,
+)
 
 
 class Config(object):
     """The base IIB Celery configuration."""
 
     # When publishing a message, don't continuously retry or else the HTTP connection times out
-    broker_transport_options = {'max_retries': 10}
+    broker_transport_options: Dict[str, int] = {'max_retries': 10}
     # Avoid infinite Celery retries when the broker is offline.
     broker_connection_max_retries: int = 10
-    iib_aws_s3_bucket_name = None
-    iib_api_timeout = 60
-    iib_docker_config_template = os.path.join(
+    iib_aws_s3_bucket_name: Optional[str] = None
+    iib_api_timeout: int = 60
+    iib_docker_config_template: str = os.path.join(
         os.path.expanduser('~'), '.docker', 'config.json.template'
     )
-    iib_greenwave_url = None
-    iib_grpc_init_wait_time = 30
-    iib_grpc_max_port_tries = 100
-    iib_grpc_max_tries = 5
-    iib_grpc_start_port = 50051
-    iib_image_push_template = '{registry}/iib-build:{request_id}'
-    iib_index_image_output_registry = None
-    iib_log_level = 'INFO'
-    iib_organization_customizations = {}
-    iib_request_logs_dir = None
-    iib_request_logs_format = (
+    iib_greenwave_url: Optional[str] = None
+    iib_grpc_init_wait_time: int = 30
+    iib_grpc_max_port_tries: int = 100
+    iib_grpc_max_tries: int = 5
+    iib_grpc_start_port: int = 50051
+    iib_image_push_template: str = '{registry}/iib-build:{request_id}'
+    iib_index_image_output_registry: Optional[str] = None
+    iib_log_level: str = 'INFO'
+    iib_organization_customizations: iib_organization_customizations_type = {}
+    iib_request_logs_dir: Optional[str] = None
+    iib_request_logs_format: str = (
         '%(asctime)s %(name)s %(processName)s {request_id} '
         '%(levelname)s %(module)s.%(funcName)s %(message)s'
     )
-    iib_request_logs_level = 'DEBUG'
-    iib_required_labels = {}
-    iib_request_related_bundles_dir = None
+    iib_request_logs_level: str = 'DEBUG'
+    iib_required_labels: Dict[str, str] = {}
+    iib_request_related_bundles_dir: Optional[str] = None
     # Configuration for dogpile.cache
     # Disabled by default (by using 'dogpile.cache.null').
     # To enable caching set 'dogpile.cache.memcached' as backend.
-    iib_dogpile_backend = 'dogpile.cache.null'
-    iib_dogpile_expiration_time = 600
-    iib_dogpile_arguments = {'url': ['127.0.0.1']}
-    iib_skopeo_timeout = '300s'
+    iib_dogpile_backend: str = 'dogpile.cache.null'
+    iib_dogpile_expiration_time: int = 600
+    iib_dogpile_arguments: Dict[str, List[str]] = {'url': ['127.0.0.1']}
+    iib_skopeo_timeout: str = '300s'
     iib_total_attempts: int = 5
     iib_retry_delay: int = 5
     iib_retry_jitter: int = 5
-    include = [
+    include: List[str] = [
         'iib.workers.tasks.build',
         'iib.workers.tasks.build_merge_index_image',
         'iib.workers.tasks.build_regenerate_bundle',
@@ -53,25 +65,25 @@ class Config(object):
         'iib.workers.tasks.general',
     ]
     # Path to hidden location of SQLite database
-    hidden_index_db_path = '/var/lib/iib/_hidden/do.not.edit.db'
+    hidden_index_db_path: str = '/var/lib/iib/_hidden/do.not.edit.db'
     # The task messages will be acknowledged after the task has been executed,
     # instead of just before
-    task_acks_late = True
+    task_acks_late: bool = True
     # Don't use the default 'celery' queue and routing key
-    task_default_queue = 'iib'
-    task_default_routing_key = 'iib'
+    task_default_queue: str = 'iib'
+    task_default_routing_key: str = 'iib'
     # Requeue the message if the worker abruptly exits or is signaled
-    task_reject_on_worker_lost = True
+    task_reject_on_worker_lost: bool = True
     # Path to index.db in our temp directories used in IIB code
-    temp_index_db_path = 'database/index.db'
+    temp_index_db_path: str = 'database/index.db'
     # For now, only allow a single process so that all tasks are processed serially
-    worker_concurrency = 1
+    worker_concurrency: int = 1
     # Don't allow the worker to fetch more messages than it can handle at a time. This is so that
     # other tasks aren't starved. This will only be useful once more workers are enabled.
-    worker_prefetch_multiplier = 1
+    worker_prefetch_multiplier: int = 1
     # Enable send events to the broker. This is needed for celery promethues exporter
-    worker_send_task_events = True
-    task_send_sent_event = True
+    worker_send_task_events: bool = True
+    task_send_sent_event: bool = True
 
 
 class ProductionConfig(Config):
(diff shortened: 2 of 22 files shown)


iib/workers/config.py
```diff
@@ -170,13 +172,14 @@ def configure_celery(celery_app):
...
         # a hack taken from flask.config.from_pyfile.
-        _user_config = {}
+        _user_config: Dict[str, Any] = {}
         with open(prod_config_file_path, mode='rb') as config_file:
             exec(compile(config_file.read(), prod_config_file_path, 'exec'), _user_config)  # nosec
 
         # Celery doesn't support configuring from multiple objects, so this is a way for
         # the configuration in prod_config_file_path to override the defaults in ProductionConfig
-        config = ProductionConfig()
+        # TODO is there some difference between ProductionConfig() and ProductionConfig - try debug
```
xDaile, line 181: I am not sure If this is okay, however tests passed

xDaile:
@exd-hello-operator-guild PR is ready to be reviewed.

yashvardhannanavati:
@release-engineering/exd-guild-hello-operator

iib/workers/api_utils.py
```diff
@@ -5,13 +5,14 @@
 import requests
 from requests.packages.urllib3.util.retry import Retry
 import requests_kerberos
+from typing import Any, Dict, Optional
```
yashvardhannanavati, line 8: looks like `typing` is a python standard library. This line should go after line 3 and before the empty line

iib/workers/config.py
```diff
@@ -138,8 +140,8 @@ class DevelopmentConfig(Config):
         ],
     }
     iib_registry = 'registry:8443'
```
yashvardhannanavati, line 142: I see such strings not having static types. Is there a reason for that?

iib/workers/config.py
```diff
@@ -170,12 +172,13 @@ def configure_celery(celery_app):
...
         # Celery doesn't support importing config files that aren't part of a Python path. This is
         # a hack taken from flask.config.from_pyfile.
-        _user_config = {}
+        _user_config: Dict[str, Any] = {}
         with open(prod_config_file_path, mode='rb') as config_file:
             exec(compile(config_file.read(), prod_config_file_path, 'exec'), _user_config)  # nosec
 
         # Celery doesn't support configuring from multiple objects, so this is a way for
         # the configuration in prod_config_file_path to override the defaults in ProductionConfig
+        # TODO is there some difference between ProductionConfig() and ProductionConfig - try debug
```
yashvardhannanavati, line 181: This line should probably be removed?

iib/workers/config.py
```diff
@@ -256,7 +259,6 @@ def _validate_iib_org_customizations(iib_org_customizations):
         'image_name_from_labels': {'template'},
         'enclose_repo': {'enclosure_glue', 'namespace'},
     }
-
```
yashvardhannanavati, line 259: This is probably a mistake. Could you please undo this

iib/workers/greenwave.py
```diff
@@ -4,6 +4,7 @@
 import logging
 
 import requests
+from typing import Any, Dict, List
```
yashvardhannanavati, line 7: Same as above. This is a standard library

iib/workers/tasks/build_regenerate_bundle.py
```diff
@@ -457,7 +484,7 @@ def _replace_csv_pullspecs(bundle_metadata, replacement_pullspecs):
     for old_pullspec, new_pullspec in replacement_pullspecs.items():
         log.debug(
             '%s will be replaced with %s in the bundle CSVs',
-            old_pullspec.to_str(),
+            old_pullspec.__repr__(),
```
yashvardhannanavati, line 487: Could you please change this back to as it was. If you want this change to come in, could you please add it in a different PR

yashvardhannanavati (changes requested):
Good work! This looks promising. A few comments

iib/workers/config.py
```diff
@@ -10,31 +12,31 @@ class Config(object):
...
     iib_docker_config_template = os.path.join(
         os.path.expanduser('~'), '.docker', 'config.json.template'
     )
-    iib_greenwave_url = None
+    iib_greenwave_url: Optional[str] = None
     iib_grpc_init_wait_time = 30
     iib_grpc_max_port_tries = 100
     iib_grpc_max_tries = 5
     iib_grpc_start_port = 50051
     iib_image_push_template = '{registry}/iib-build:{request_id}'
```
JAVGan, line 28: I think these values are missing the static type as well as some other config values

iib/workers/tasks/build.py
```diff
@@ -120,7 +121,11 @@ def _cleanup():
 
 
 @retry(exceptions=IIBError, tries=get_worker_config().iib_total_attempts, logger=log)
-def _create_and_push_manifest_list(request_id, arches, build_tags):
+def _create_and_push_manifest_list(
+    request_id: int,
+    arches: Iterable,
```
JAVGan, line 126: I believe this could be something like this:

```suggestion
    arches: Iterable[str],
```

iib/workers/tasks/build.py
```diff
@@ -217,7 +222,7 @@ def _update_index_image_pull_spec(
     :raises IIBError: if the manifest list couldn't be created and pushed
     """
     conf = get_worker_config()
-    if from_index and overwrite_from_index:
+    if from_index and overwrite_from_index and resolved_prebuild_from_index:
```
JAVGan, line 225: Is this logic change required?

iib/workers/tasks/build.py
```diff
@@ -641,8 +661,10 @@ def _skopeo_copy(source, destination, copy_all=False, exc_msg=None):
 
 
 def _verify_index_image(
-    resolved_prebuild_from_index, unresolved_from_index, overwrite_from_index_token=None
-):
+    resolved_prebuild_from_index: str,
+    unresolved_from_index: str,
+    overwrite_from_index_token: str = None,
```
JAVGan, line 666:
```suggestion
    overwrite_from_index_token: Optional[str] = None,
```

iib/workers/tasks/build_merge_index_image.py
```diff
@@ -397,6 +397,8 @@ def is_bundle_version_valid(bundle_path, valid_ocp_version):
         raise IIBError(f'Invalid OCP version, "{valid_ocp_version}", specified in Index Image')
     try:
         bundle_version_label = get_image_label(bundle_path, 'com.redhat.openshift.versions')
+        if bundle_version_label is None:
+            raise ValueError
```
JAVGan, line 401: I believe it would be better to raise an `IIBError` since it will be caught by API to properly log, also please consider including an error message for the error.

iib/workers/tasks/build.py
xDaile: Sure, I missed that somehow.

iib/workers/tasks/build.py
xDaile: Fixed

iib/workers/api_utils.py
xDaile: It should be okay in every file now.

iib/workers/tasks/build_merge_index_image.py
xDaile: Please check again. I made the change according to that.

iib/workers/tasks/build.py
xDaile: My bad, did not noticed it is checked in functions called after that.

iib/workers/config.py
xDaile: I intended to add static types just to method definitions, however I also added static types to class variables. Is it okay in this way?

iib/workers/config.py
xDaile: I just added static types to all class variables (most of them in config.py). 
I intended to add static types just to method definitions at first, that´s why it was without them. 
Is it okay now?

JAVGan: Yes, thanks!

iib/workers/tasks/build_merge_index_image.py
JAVGan: Great, thanks!

iib/workers/config.py
```diff
@@ -81,10 +83,10 @@ class ProductionConfig(Config):
...
     """The development IIB Celery configuration."""
 
-    broker_url = 'amqp://iib:iib@rabbitmq:5673//'
-    iib_api_url = 'http://iib-api:8080/api/v1/'
-    iib_log_level = 'DEBUG'
-    iib_organization_customizations = {
+    broker_url: str = 'amqp://iib:iib@rabbitmq:5673//'
+    iib_api_url: str = 'http://iib-api:8080/api/v1/'
+    iib_log_level: str = 'DEBUG'
+    iib_organization_customizations: Dict[str, Any] = {
```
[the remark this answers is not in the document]
xDaile: By this, you mean that iib_organization_customizations variable, or the DevelopmentConfig class?
By the way, thanks for the idea, I implemented TypeDict at different places.

iib/workers/tasks/build.py
```diff
@@ -340,21 +350,22 @@ def _get_present_bundles(from_index, base_dir):
...
         unique_present_bundles.append(bundle)
         unique_present_bundles_pull_spec.append(bundle_path)
-
     return unique_present_bundles, unique_present_bundles_pull_spec
 
 
-def _get_missing_bundles(present_bundles, bundles):
+def _get_missing_bundles(
+    present_bundles: List[BundleImage],
+    resolved_bundles: List[str],
```
yashvardhannanavati, line 367: Since the variable name changed, you probably need to change the docstring here too :)

iib/workers/tasks/build.py
```diff
@@ -340,21 +350,22 @@ def _get_present_bundles(from_index, base_dir):
         return [], []
 
     # Transform returned data to parsable json
-    unique_present_bundles = []
+    unique_present_bundles: List[BundleImage] = []
     unique_present_bundles_pull_spec = []
     present_bundles = get_bundle_json(bundles)
-
```
yashvardhannanavati, line 346: This was probably removed by accident? Could you please add the empty line back?

iib/workers/tasks/build.py
```diff
@@ -340,21 +350,22 @@ def _get_present_bundles(from_index, base_dir):
...
     for bundle in present_bundles:
         bundle_path = bundle['bundlePath']
         if bundle_path in unique_present_bundles_pull_spec:
             continue
         unique_present_bundles.append(bundle)
         unique_present_bundles_pull_spec.append(bundle_path)
-
```
yashvardhannanavati, line 353: Same as above

iib/workers/tasks/build.py
```diff
@@ -485,12 +501,12 @@ def _opm_index_rm(base_dir, operators, binary_image, from_index, overwrite_from_
...
-    request_id,
-    output_pull_spec,
-    from_index,
-    resolved_prebuild_from_index,
-    overwrite_from_index_token=None,
-):
+    request_id: int,
+    output_pull_spec: str,
+    from_index: str,
+    resolved_prebuild_from_index: Optional[str],
```
yashvardhannanavati, line 507: This probably should be `str`

iib/workers/tasks/build.py
```diff
@@ -562,15 +581,15 @@ def _update_index_image_build_state(request_id, prebuild_info):
         image.
     """
     arches_str = ', '.join(sorted(prebuild_info['arches']))
-    payload = {
+    payload: Dict[str, Union[str, Dict[str, List[str]]]] = {
```
yashvardhannanavati, line 584: Shouldn't this be `Dict[str, str]`?

iib/workers/tasks/build_merge_index_image.py
```diff
@@ -125,7 +127,6 @@ def _add_bundles_missing_in_source(
         log.info(
             '%s bundles have invalid version label and will be deprecated.', len(invalid_bundles)
         )
-
```
yashvardhannanavati, line 128: maybe indentation mistake? Could you please revert this?

iib/workers/tasks/build_regenerate_bundle.py
```diff
@@ -287,11 +300,13 @@ def _adjust_operator_bundle(
             log.info('Resolving image pull specs')
             bundle_metadata = _get_bundle_metadata(operator_manifest, pinned_by_iib)
             _resolve_image_pull_specs(bundle_metadata, labels, pinned_by_iib)
-
     return labels
 
 
-def _get_bundle_metadata(operator_manifest, pinned_by_iib):
+def _get_bundle_metadata(
+    operator_manifest: OperatorManifest,
```
yashvardhannanavati, line 307: Nice! I like this

iib/workers/tasks/build_regenerate_bundle.py
yashvardhannanavati: I believe this is still pending? We discussed to remove __repr__ here

iib/workers/tasks/iib_types.py
```diff
@@ -0,0 +1,51 @@
+from typing import TypedDict, Set, List, Dict, Optional
```
yashvardhannanavati, line 1: knitpick: Could we please rename this file to `iib_static_types`?

yashvardhannanavati (changes requested):

iib/workers/tasks/iib_types.py
yashvardhannanavati: Nice! I like this file

iib/workers/config.py
```diff
@@ -2,6 +2,8 @@
 import os
 import logging
 import types
+from typing import Any, Dict, Optional, Type, Union, List
+from celery import Celery, app
```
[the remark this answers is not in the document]
xDaile: Yes, of course. That makes sense.

iib/workers/tasks/build.py
xDaile: That´s right, thanks.

iib/workers/tasks/build.py
xDaile: Thanks, I forgot to do that.

iib/workers/tasks/build_regenerate_bundle.py
xDaile: Should be okay now.

iib/workers/tasks/build.py
xDaile: Actually, It was alright. 
The second Dict[str,List[str]] is needed for bundle_mapping assigned into the payload variable later (two lines after those selected).

xDaile: What is assigned into the payload['bundle_mapping'] in the tests looks like this:
`{'some-bundle': ['quay.io/some-bundle:v1'], 'some-bundle2': ['quay.io/some-bundle2:v1']}`

iib/workers/tasks/build.py
```diff
@@ -120,7 +122,11 @@ def _cleanup():
 
 
 @retry(exceptions=IIBError, tries=get_worker_config().iib_total_attempts, logger=log)
-def _create_and_push_manifest_list(request_id, arches, build_tags):
+def _create_and_push_manifest_list(
+    request_id: int,
+    arches: Iterable[str],
```
chandwanitulsi, line 127: Why are we using Iterable here and not List?

iib/workers/tasks/build_merge_index_image.py
```diff
@@ -153,31 +155,31 @@ def _add_bundles_missing_in_source(
     )
     _add_label_to_index(
         'com.redhat.index.delivery.distribution_scope',
-        distribution_scope,
+        distribution_scope,  # type: ignore
```
chandwanitulsi, line 158: May I ask in what situations you have used type: ignore.
I noticed its used in function calls, local variables and some others

iib/workers/tasks/build.py
xDaile: That´s a good point, I used that as i saw it elsewhere already used for arches. 
Now I changed arches, and all add_arches function parameters to have type Set - as the arches should not repeat itself. 
If I am wrong, please let me know, I will change those to List type.

iib/workers/tasks/build_merge_index_image.py
xDaile: Yes, of course. We discussed that with @yashvardhannanavati . We did not wanted to do big changes to logic of the IIB with this PR. 
Adding the types resulted in Mypy errors, which few were fixed, however few of them cannot be fixed without changing of the IIB logic (most of the times checking if the parameter is not None, I think this example is that case too).
So places where those errors were, are now annotated with those `# type: ignore` comments. After this PR is merged I will create another PR, where those comments will be removed, and fixes which are removing those errors proposed. 
Or I can make another commit which will fix those mypy errors and add it to this PR.

xDaile:
Pipeline fail is not connected with changes made in this PR (docs build is failing).

iib/workers/tasks/build.py
chandwanitulsi: Set is better :+1:

iib/workers/tasks/build_merge_index_image.py
chandwanitulsi: Another PR would be better :+1:

iib/workers/tasks/utils.py
```diff
@@ -199,15 +209,21 @@ class RequestConfig:
...
         'cnr_token',
         'overwrite_from_index_token',
         'overwrite_target_index_token',
         'registry_auths',
     ]
 
-    _attrs = ["_binary_image", "distribution_scope", "binary_image_config"]
+    _attrs: List[str] = ["_binary_image", "distribution_scope", "binary_image_config"]
     __slots__ = _attrs
+    if TYPE_CHECKING:
```
yashvardhannanavati, line 221: I don't get this. Could you please add a comment explaining?

yashvardhannanavati (changes requested):
Sorry for the last knitpick comments. But this is really taking shape now. Great job!

iib/workers/config.py
```diff
@@ -81,87 +93,114 @@ class ProductionConfig(Config):
...
+                        'quay.io': 'registry.koji.company.com',
+                    },
+                }
+            ),
+            RelatedBundles({'type': 'related_bundles'}),
+            ImageNameFromLabels({'type': 'image_name_from_labels', 'template': '{name}-{version}'}),
+            EncloseRepo(
+                {'type': 'enclose_repo', 'enclosure_glue': '----', 'namespace': "company-pending"}
+            ),
+            RegistryReplacements(
```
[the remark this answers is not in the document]
xDaile: Only way, I found (Jindra advised me) that does not need breaking the datastructure into smaller parts.

iib/workers/tasks/iib_static_types.py
```diff
@@ -0,0 +1,100 @@
...
+class BundleImage(TypedDict):
+    """Type class referencing data related to bundle image used for mypy checking."""
+
+    packageName: str
+    version: str
+    bundlePath: str
+    csvName: Optional[str]
+
+
+class ResolveImagePullspecs(TypedDict):
```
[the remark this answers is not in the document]
xDaile: It could be also done without creating the class. 
Instead of class, there could be `resolve_image_pullspecs = TypedDict('resolve_image_pullspecs', {'type':str})`.

lipoja (changes requested):

iib/workers/config.py
```diff
@@ -2,76 +2,88 @@
...
+    iib_request_logs_dir: Optional[str] = None
+    iib_request_logs_format: str = (
         '%(asctime)s %(name)s %(processName)s {request_id} '
         '%(levelname)s %(module)s.%(funcName)s %(message)s'
     )
-    iib_request_logs_level = 'DEBUG'
-    iib_required_labels = {}
-    iib_request_related_bundles_dir = None
+    iib_request_logs_level: str = 'DEBUG'
+    iib_required_labels: Dict[str, Any] = {}
```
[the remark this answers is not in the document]
xDaile: Usually? Could you specify better the case when they are not `Dict[str, str]`? So I can change those parameters in right way.

iib/workers/tasks/iib_static_types.py
```diff
@@ -0,0 +1,143 @@
...
+    bundle_image: Optional[str]
+    omps_operator_version: Optional[str]
+    target_index_resolved: Optional[str]
+    index_image_resolved: Optional[str]
+    internal_index_image_copy: Optional[str]
+    internal_index_image_copy_resolved: Optional[str]
+    index_image: Optional[str]
+    binary_image: Optional[str]
+    binary_image_resolved: Optional[str]
+    stage: Optional[str]
```
yashvardhannanavati, line 79: This should probably be removed.

iib/workers/tasks/build.py
```diff
@@ -222,7 +228,7 @@ def _update_index_image_pull_spec(
             request_id,
             output_pull_spec,
             from_index,
-            resolved_prebuild_from_index,
+            resolved_prebuild_from_index,  # type: ignore
```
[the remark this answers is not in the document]
xDaile: Of course, Actually I added to the line above comment with the mypy error message - `            # MYPY error: Argument 4 to "_overwrite_from_index" has incompatible type "Optional[str]"; expected "str"`. There is a way, however I would like to create separated PR for those fixes.

iib/workers/api_utils.py
```diff
@@ -156,7 +156,6 @@ def update_request(
         else:
             _exc_msg = f'The worker failed to update the request {request_id}'
         raise IIBError(_exc_msg)
-
```
yashvardhannanavati, line 152: Probably a mistaken delete

iib/workers/tasks/iib_static_types.py
```diff
@@ -52,49 +54,90 @@ class BundleImage(TypedDict):
     csvName: Optional[str]
 
 
-class ResolveImagePullspecs(TypedDict):
-    """TypedDict class for typing the DevelopmentConfig class."""
+class BinaryImageConfig(TypedDict):
+    """Type class referencing Binary Image configuration."""
 
-    type: str
+    prod: Dict[str, str]
```
yashvardhannanavati, line 60: I don't think `prod` is the right key name here. It's a string. Let's keep this Dict[str, Dict[str, str]]

yashvardhannanavati (changes requested):
Let's squash these into a single commit. One final review and I think this should be +1 from me :)

iib/workers/tasks/iib_static_types.py
xDaile: You mean stage parameter?

iib/workers/tasks/iib_static_types.py
xDaile: I removed the class and replaced it with Dict[str, Dict[str, str]]. Pleas check.

iib/workers/tasks/build_merge_index_image.py
```diff
@@ -153,31 +155,33 @@ def _add_bundles_missing_in_source(
     )
     _add_label_to_index(
         'com.redhat.index.delivery.distribution_scope',
-        distribution_scope,
+        # MYPY error: Argument 2 to "_add_label_to_index"
```
midnightercz, line 158: In this case _add_label_to_index method should be updated to tolerate None as label or distribution scope should be changed make distribution_scope non-null

iib/workers/tasks/opm_operations.py
```diff
@@ -171,10 +178,14 @@ def _serve_cmd_at_port(serve_cmd, cwd, port, max_tries, wait_time):
             ret = rpc_proc.poll()
             # process has terminated
             if ret is not None:
-                stderr = rpc_proc.stderr.read()
-                if 'address already in use' in stderr:
+                if rpc_proc.stderr is None:
```
chandwanitulsi, line 181: Can you please explain why we added failure when stderr is not captured?

xDaile: Of course. So this part of the code will raise Exception one way or other. However there was this MYPY error `error: Item "None" of "Optional[IO[str]]" has no attribute "read"`. Basically I just added check for value, that is used later (and can be possibly empty). 
I can make this change in separated PR, however I believe this is not change of logic - either way exception will be thrown.

iib/workers/tasks/build_merge_index_image.py
xDaile: Do I understand right that you are proposing those two solutions?
- None value should be accepted as the label_key parameter in the _add_label_to_index method
-  distribution_scope should be set to some empty value when it`s value is None
If I am right, which of that solutions is better according to you? Personally I like second one more, as it does not interfere with other part of the code - in some cases Mypy check that you are not sending None, could be useful.

yashvardhannanavati (changes requested):

iib/workers/tasks/opm_operations.py
xDaile: Should be okay now.

xDaile: @chandwanitulsi stderr can be None, according to https://docs.python.org/3/library/subprocess.html#subprocess.CompletedProcess.stderr . I am not sure if it ever happened or if it is possible, however in case that `ret` would be set and `stderr` would be `None`, there would not be relevant error message. Please check solution now (error message was changed).

tests/test_workers/test_tasks/test_build_merge_index_image.py
```diff
@@ -397,6 +397,7 @@ def test_add_bundles_missing_in_source(
         1,
         'amd64',
         '4.6',
+        'private',
```
yashvardhannanavati, line 400: distribution_scope is usually `stage`, `prod` or `dev`

yashvardhannanavati (changes requested):
Just one typo comment :)

tests/test_workers/test_tasks/test_build_merge_index_image.py
xDaile: 'Private' replaced for 'dev'. Is that okay?

iib/workers/tasks/iib_static_types.py
```diff
@@ -0,0 +1,136 @@
...
+    target_index_resolved: str
+    target_ocp_version: str
+
+
+class BundleImage(TypedDict):
+    """Type class referencing data related to bundle image used for mypy checking."""
+
+    bundlePath: str
+    csvName: Optional[str]
+    packageName: str
```
JAVGan, line 53: Don't we need `total=False` here as well since `csvName` can be optional?

iib/workers/tasks/iib_static_types.py
```diff
@@ -0,0 +1,136 @@
...
+class AllIndexImagesInfo(TypedDict):
+    """Type class referencing group of IndexImageInfo classes used for mypy checking."""
+
+    from_index: IndexImageInfo
+    source_from_index: IndexImageInfo
+    target_index: IndexImageInfo
+
+
+# MYPY check - should label be optional? If not total parameter can be removed.
+class PrebuildInfo(TypedDict, total=False):
```
JAVGan, line 31: AFAIK `labels` are indeed optional so it's fine to leave `total=False` there.

iib/workers/tasks/build_regenerate_bundle.py
```diff
@@ -457,7 +486,7 @@ def _replace_csv_pullspecs(bundle_metadata, replacement_pullspecs):
     for old_pullspec, new_pullspec in replacement_pullspecs.items():
         log.debug(
             '%s will be replaced with %s in the bundle CSVs',
-            old_pullspec.to_str(),
+            old_pullspec,
```
JAVGan, line 489: Why was this changed?

iib/workers/tasks/build_merge_index_image.py
JAVGan, line 158: It would be nice to put a comment before `_add_label_to_index` indicating the reason for this `type: ignore` like you did for the other ones.

JAVGan (commented):
Just small comments.

Most of the work LGTM.

Also, please rebase this branch and solve the conflicts with `iib/workers/tasks/utils.py`

yashvardhannanavati (dismissed):

iib/workers/tasks/iib_static_types.py
xDaile: Do we need those labels? It seems that they are passed to variable `prebuild_info` in the `handle_create_empty_index_request` method, however after that in the `_update_index_image_build_state` method they are not even used {I did not found any other place where those labels could be used}.

iib/workers/tasks/build_regenerate_bundle.py
xDaile: Type of the `old_pullspec` variable is already str.

iib/workers/tasks/build_merge_index_image.py
xDaile: Seems that it somehow slipped, this error was removed, with replacing of the `distribution_scope: Optional[str]` for `distribution_scope: str ` in the `_add_bundles_missing_in_source` method.

iib/workers/tasks/general.py
```diff
@@ -13,7 +15,12 @@
 
 @app.task
 @request_logger
-def failed_request_callback(context, exc, traceback, request_id):
+def failed_request_callback(
+    context: celery.app.task.Context,
+    exc: Exception,
+    traceback,
```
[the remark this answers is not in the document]
xDaile: Mypy `reveal_type()` method is telling me that type is Any. I am not sure with the required type too, so I would propose leaving there type `Any`? If you can think of something better please let me know.

iib/workers/tasks/iib_static_types.py
JAVGan: The element `labels` is a supported optional parameter used only in `create-empty-index` requests, as you can see [here](https://iib.qe.engineering.redhat.com/#/paths/~1builds~1create-empty-index/post). It allows external entities to add additional labels to the resulting index image and if not specified we just set [these ones](https://github.com/release-engineering/iib/blob/master/iib/workers/tasks/build_create_empty_index.py#L126-L129).

If we remove the `labels` from IIB's code it will break this feature from its API on `create-empty-index` and will cause some troubles thus we need to keep it but consider this parameter to be optional from the requester point of view.

JAVGan (approved):
LGTM

iib/workers/tasks/iib_static_types.py
xDaile: Sure, updated. Thanks. However Optional[] does not mean that class has to have total=false, Optional means that parameter could be set to None value, and total=false means that it is possible that some of the keys will be missing.

iib/workers/tasks/iib_static_types.py
```diff
@@ -37,23 +39,33 @@ class PrebuildInfo(TypedDict, total=False):
...
     ocp_version: str
     source_from_index_resolved: str
     source_ocp_version: str
     target_index_resolved: str
     target_ocp_version: str
 
 
-class BundleImage(TypedDict):
-    """Type class referencing data related to bundle image used for mypy checking."""
+class PrebuildInfo(PrebuildInfoBase, total=False):
```
yashvardhannanavati, line 49: Hm, I think we can still use `NotRequired` in Python 3.8 if we use `typing_extensions.TypedDict` instead of `typing.TypedDict`

https://peps.python.org/pep-0655/#usage-in-python-3-11

lipoja:
@xDaile There is missing `typing_extensions` dependence in `requirements-test.txt`

yashvardhannanavati (approved):

lipoja (approved):

PR #418, opened 2022-08-03
Append overwrite index token to current docker config.

lipoja:
[CLOUDDST-13158]

### DIFF
--- a/iib/workers/tasks/build_merge_index_image.py
+++ b/iib/workers/tasks/build_merge_index_image.py
@@ -60,6 +60,7 @@ def _add_bundles_missing_in_source(
     arch: str,
     ocp_version: str,
     distribution_scope: str,
+    target_index=None,
     overwrite_target_index_token: Optional[str] = None,
 ) -> Tuple[List[BundleImage], List[BundleImage]]:
     """
@@ -76,6 +77,7 @@ def _add_bundles_missing_in_source(
     :param int request_id: the ID of the IIB build request.
     :param str arch: the architecture to build this image for.
     :param str ocp_version: ocp version which will be added as a label to the image.
+    :param str target_index: the pull specification of the container image
     :param str overwrite_target_index_token: the token used for overwriting the input
         ``source_from_index`` image. This is required to use ``overwrite_target_index``.
         The format of the token must be in the format "user:password".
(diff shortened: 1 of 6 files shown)


JAVGan (changes requested):
The following tests are failing on CI:

- tests/test_workers/test_tasks/test_build.py::test_opm_index_add
- tests/test_workers/test_tasks/test_build.py::test_opm_index_rm 
- tests/test_workers/test_tasks/test_build.py::test_verify_index_image_failure
- tests/test_workers/test_tasks/test_utils.py::test_set_registry_token

Can you please check them?

Also please rebase the branch.

lipoja:
@JAVGan thank but this failed another test that I did so I will be chaning a lot here and I could not find button to mark it as draft

iib/workers/tasks/build_merge_index_image.py
```diff
@@ -126,28 +128,26 @@ def _add_bundles_missing_in_source(
...
+        else:
+            _opm_index_add(
+                base_dir=base_dir,
+                bundles=missing_bundle_paths,
+                binary_image=binary_image,
+                from_index=source_from_index,
+                # Use podman until opm's default mechanism is more resilient:
+                #   https://bugzilla.redhat.com/show_bug.cgi?id=1937097
+                container_tool='podman',
+            )
```
JAVGan, line 150: Why do we want to call `opm_registry_add_fbc` and `_opm_index_app` without passing the `overwrite_from_index_token`?

lipoja: Correct token (docker config file) is set by `with set_registry_token(overwrite_target_index_token, target_index, append=True):` on line 131. 
Unfortunately we have these settings wrongly written inside of almost all function. We agreed on guild meeting that we will support "private registry" for source index as well (if it is the same as target index).That is why I removed the overwrite token before this call to make sure it is not wrongly set for incorrect repository (from_index is source index and we set token for target index in this API, therefore call inside of this function is can set token to different repository which breaks IIB). I would need more time to do proper refactoring of this.

iib/workers/tasks/utils.py
```diff
@@ -526,7 +528,17 @@ def set_registry_token(token, container_image):
 
     registry = ImageName.parse(container_image).registry
     encoded_token = base64.b64encode(token.encode('utf-8')).decode('utf-8')
-    registry_auths = {'auths': {registry: {'auth': encoded_token}}}
+    registry_auths = {'auths': {}}
+    if append:
+        docker_config_path = os.path.join(os.path.expanduser('~'), '.docker', 'config.json')
+        if os.path.exists(docker_config_path):
+            with open(docker_config_path, 'r') as f:
+                registry_auths = json.load(f)
```
xDaile, line 536: I am not sure how config.json file is created, however I would add here TypeError catch in case that json.loads() would fail to read the file.

JAVGan:
Whenever possible please rebase and resolve the conflicts

iib/workers/tasks/utils.py
lipoja: I've added catching exception an logging it. However I raised it again because I think we want to fail horribly when the config is incorrect.

iib/workers/tasks/utils.py
```diff
@@ -68,7 +68,7 @@ def _add_property_to_index(db_path: str, property: Dict[str, str]) -> None:
     con.close()
 
 
-def add_max_ocp_version_property(resolved_bundles: List[str], temp_dir: str) -> None:
+def add_max_ocp_version_property(resolved_bundles, temp_dir):
```
JAVGan, line 71: Can you please keep the typing annotation?

lipoja: Nice catch! Thanks

iib/workers/tasks/build_merge_index_image.py
```diff
@@ -128,28 +130,26 @@ def _add_bundles_missing_in_source(
...
-            base_dir=base_dir,
-            bundles=missing_bundle_paths,
-            binary_image=binary_image,
-            from_index=source_from_index,
-            overwrite_from_index_token=overwrite_target_index_token,
-            # Use podman until opm's default mechanism is more resilient:
-            #   https://bugzilla.redhat.com/show_bug.cgi?id=1937097
-            container_tool='podman',
-        )
+        if is_source_fbc:
```
xDaile, line 135: Could not be indentation here be reduced by one tab (as in build.py case)?

lipoja: I need to have set the token correctly for these calls (is_image_fbc, opm_registry_add_fbc, _opm_index_add).
That is why it is indented this way.

xDaile (approved):

JAVGan (approved):
LGTM

yashvardhannanavati (approved):

PR #425, opened 2022-09-06
Adding ability to create single-active-consumer queues.

lipoja:
- add iib_sac_queue in config
- setting 'x-single-active-consumer' as queue_arguments (RabbitMQ)

[CLOUDDST-15267]

### DIFF
--- a/iib/workers/config.py
+++ b/iib/workers/config.py
@@ -4,6 +4,7 @@ import logging
 import types
 from typing import Any, Dict, List, Optional, Type, Union
 
+from kombu import Queue
 from celery import Celery, app
 
 from iib.exceptions import ConfigError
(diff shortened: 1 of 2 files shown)


JAVGan (approved):

iib/workers/config.py
```diff
@@ -39,6 +40,7 @@ class Config(object):
     iib_index_image_output_registry: Optional[str] = None
     iib_log_level: str = 'INFO'
     iib_organization_customizations: iib_organization_customizations_type = {}
+    iib_sac_queue = []
```
yashvardhannanavati, line 43: missing type :)

lipoja: added

yashvardhannanavati (approved):

PR #430, opened 2022-09-14
Bug Fix: Convert set to list before dumping to JSON

yashvardhannanavati:


yashvardhannanavati:
@xDaile I had to revert a few lines from #412 since sets are not serializable.

chandwanitulsi (approved):

PR #431, opened 2022-09-14
Revert "enable customizations for podman-compose"

yashvardhannanavati:
Reverts release-engineering/iib#421


yashvardhannanavati:
I was not able to run docker-compose locally with these changes. There are a few more things that will need to be changed. @arewm An easy way here could be to have docker-compose.yml and podman-compose.yml in the root directory

yashvardhannanavati:
@lipoja PTAL

PR #432, opened 2022-09-14
Revert "enable customizations for podman-compose"

yashvardhannanavati:
Reverts release-engineering/iib#421


yashvardhannanavati:
Doh! I tried something but didn't work. I had to rebase it manually. 

I was not able to run docker-compose locally with these changes. There are a few more things that will need to be changed eg: path of the dockerfile. @arewm An easy way here could be to have docker-compose.yml and podman-compose.yml in the root directory

@lipoja PTAL

arewm:
I think I fixed the behavior with this PR: https://github.com/release-engineering/iib/pull/433 ... instead of reverting the PR/commit.

lipoja (approved):

lipoja:
I will accept this revert instead of https://github.com/release-engineering/iib/pull/433 fix since docker-compose still does not work and I would like to make release

PR #435, opened 2022-09-15
IIB 6.6.1 release

lipoja:
- Add recursive-related-bundles endpoint
- Adding ability to create single-active-consumer queues
- Add static types and mypy checks
- Append overwrite index token to current docker config
- Fixed config loading

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
-## 6.6.0
+## 6.6.1
 - Add recursive-related-bundles endpoint 
 - Adding ability to create single-active-consumer queues
 - Add static types and mypy checks
--- a/setup.py
+++ b/setup.py
@@ -3,7 +3,7 @@ from setuptools import setup, find_packages
 
 setup(
     name='iib',
-    version='6.6.0',
+    version='6.6.1',
     long_description=__doc__,
     packages=find_packages(exclude=['tests', 'tests.*']),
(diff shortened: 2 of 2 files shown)


CHANGELOG.md
```diff
@@ -6,7 +6,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
-## 6.6.0
+## 6.6.1
 - Add recursive-related-bundles endpoint
```
yashvardhannanavati, line 10: Should we have two releases in the release notes? 6.6.1 only having the bug fix and the rest in 6.6.0

yashvardhannanavati (changes requested):

CHANGELOG.md
chandwanitulsi: Only 6.6.1 looks okay to me since 6.6.0 was not released, but no strong objections to having both releases here

chandwanitulsi (approved):

CHANGELOG.md
lipoja: GitHub release of 6.6.0 was never done - we have only  v6.6.0 tag present...

yashvardhannanavati: Ok, good to go with 6.6.1

yashvardhannanavati (approved):

yashvardhannanavati (approved):

CHANGELOG.md
lipoja: And I hope that v6.6.1 will be working I am planning to do the GitHub release - which will create official IIB images and release

PR #441, opened 2022-10-13
Fix Rm request private registry bug

yashvardhannanavati:
IIB wasn't using the private registry token to pull the index for fbc images in rm requests

Refers to CLOUDDST-15975

### DIFF
--- a/iib/workers/tasks/opm_operations.py
+++ b/iib/workers/tasks/opm_operations.py
@@ -550,6 +550,7 @@ def opm_registry_rm_fbc(
     from_index: str,
     operators: List[str],
     binary_image: str,
+    overwrite_from_index_token: Optional[str] = None,
 ) -> None:
     """
     Remove operator/s from a File Based Catalog index image.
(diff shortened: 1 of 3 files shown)


yashvardhannanavati:

@lipoja could we please include this in the next release?

lipoja (changes requested):
missing doctring otherwise OK

tests/test_workers/test_tasks/test_opm_operations.py
```diff
@@ -376,6 +378,7 @@ def test_opm_registry_rm_fbc(
         tmpdir,
     )
 
+    mock_srt.assert_called_once()
```
xDaile, line 381: I would personally fill overwrite_from_index_token parameter, when calling opm_registry_rm_fbc(). This way it would be possible to test if we are passing the parameter to set_registry_token() - now If I understand the code right this feature is not tested at all. If we would fill the parameter, this line could be changed to mock_srt.assert_called_once_with(overwrite_from_index_token, from_index, append). 
But I am not sure if this set_registry_token can do something to the rest of the test.

iib/workers/tasks/opm_operations.py
```diff
@@ -550,6 +550,7 @@ def opm_registry_rm_fbc(
     from_index: str,
     operators: List[str],
     binary_image: str,
+    overwrite_from_index_token: Optional[str] = None,
```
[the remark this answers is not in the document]
yashvardhannanavati: thanks for catching this

tests/test_workers/test_tasks/test_opm_operations.py
yashvardhannanavati: This piece is being tested in other parts of the code. But you're right, it doesn't hurt to have more aggressive checks. Changed :)

chandwanitulsi (approved):

lipoja (approved):

xDaile (approved):

PR #443, opened 2022-10-20
Adding retry and port check when opm serve/opm registry serve is called

lipoja:
This should prevent issue when _get_free_port_for_grpc returned false positive. Port which will be already taken by different service.

- we will try to run opm serve commands directly with port
- if port is taken we will get next port from given range and try that
- repeat this until free port is found

refers [CLOUDDST-16030]

### DIFF
--- a/iib/workers/tasks/opm_operations.py
+++ b/iib/workers/tasks/opm_operations.py
@@ -190,7 +190,7 @@ def _serve_cmd_at_port(
                 output = ''
 
             if 'api.Registry.ListBundles' in output or 'api.Registry.ListPackages' in output:
-                log.debug('Started the command "%s"', ' '.join(serve_cmd))
+                log.debug('Started the command "%s"; pid: %d', ' '.join(serve_cmd), rpc_proc.pid)
                 log.info('Index registry service has been initialized.')
                 return rpc_proc
 
--- a/iib/workers/tasks/opm_operations.py
+++ b/iib/workers/tasks/opm_operations.py
@@ -1,10 +1,10 @@
 import logging
 import os
+import random
 import shutil
-import socket
 import subprocess
 import time
-from typing import List, Optional, Tuple
+from typing import List, Optional, Tuple, Generator
 
 from retry import retry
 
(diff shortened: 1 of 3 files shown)


tests/test_workers/test_tasks/test_opm_operations.py
```diff
@@ -21,96 +21,80 @@ def mock_config():
...
+    port_range = set(range(port_start, port_end))
 
-@mock.patch('iib.workers.tasks.opm_operations.socket')
-def test_get_free_port_no_port(mock_socket):
-    mock_socket.socket().bind.side_effect = OSError()
-    with pytest.raises(IIBError, match='No free port has been found after 3 attempts.'):
-        opm_operations._get_free_port(50051, 50054)
+    with pytest.raises(IIBError, match='No free port has been found after 100 attempts.'):
+        for port in opm_operations._gen_port_for_grpc():
+            assert port in port_range
```
JAVGan, line 32: IIUC this is testing the range from config which now is 100 but it can be changed, right?

Isn't it better to change the `match` to use the `iib_grpc_max_port_tries` instead of hardcoded 100?

lipoja: @JAVGan  where do you see hardcoded values? Maybe I am overlooking something or I am not sure what are you referring to.

JAVGan: In the string `'No free port has been found after 100 attempts.`

This `100` can be changed, no?

lipoja: ah long day... yeah we can take it from config. I'll do it on Monday.

iib/workers/tasks/opm_operations.py
```diff
@@ -92,13 +78,17 @@ def opm_serve(catalog_dir: str) -> Tuple[int, subprocess.Popen]:
...
+        try:
+            cmd = ['opm', 'serve', catalog_dir, '-p', str(port), '-t', '/dev/null']
+            cwd = os.path.abspath(os.path.join(catalog_dir, os.path.pardir))
+            return (
+                port,
+                _serve_cmd_at_port_defaults(cmd, cwd, port),
+            )
+        except AddressAlreadyInUse:
+            log.debug('Port %s is already taken. Checking next one...', port)
+            continue
```
xDaile, line 91: Do we still need to have retry decorator if we are catching the exception? I think that now it would run 100x100 times before failing.

lipoja: @xDaile thanks, nice catch.

iib/workers/tasks/opm_operations.py
```diff
@@ -92,13 +78,17 @@ def opm_serve(catalog_dir: str) -> Tuple[int, subprocess.Popen]:
...
 
 
 @retry(exceptions=AddressAlreadyInUse, tries=get_worker_config().iib_grpc_max_tries, logger=log)
```
xDaile, line 94: I think that here the retry decorator is unnecessary too.

lipoja: please check

iib/workers/tasks/opm_operations.py
lipoja: please check

tests/test_workers/test_tasks/test_opm_operations.py
lipoja: please check

xDaile (approved):

yashvardhannanavati (approved):
I like this approach. Thanks Jan!

JAVGan (approved):

PR #444, opened 2022-10-21
Fix dev environment through compose files

JAVGan:
This PR fixes the dev environment which wasn't being able to initialize properly due to some issues in the `podman-compose.yaml` and `docker-compose.yaml` files.


lipoja (approved):
works - tested with podman only

chandwanitulsi (approved):
+1.
Works with docker-compose as well

PR #448, opened 2022-10-26
Release IIB v6.7.1

yashvardhannanavati:
Refers to CLOUDDST-16213

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+## 6.7.1
+- Fix dev env compose files
+- Fix bundle_replacements bug in regenerate-bundle request
+- Upgrade OPM to 1.26.2 in dev env
+- Fix buildah bud command retries
+
 ## 6.7.0
 - Add bundle_replacements parameter to regenerate_bundle API and worker
 - Bump mako version
--- a/setup.py
+++ b/setup.py
@@ -3,7 +3,7 @@ from setuptools import setup, find_packages
 
 setup(
     name='iib',
-    version='6.7.0',
+    version='6.7.1',
     long_description=__doc__,
     packages=find_packages(exclude=['tests', 'tests.*']),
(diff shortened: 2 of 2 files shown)


CHANGELOG.md
```diff
@@ -6,6 +6,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+## 6.7.1
```
chandwanitulsi, line 9: Maybe 6.8 would be better?
The release started a bug fix but more got added

yashvardhannanavati: But it's still bug fixes and no new features are being added. So I thought it might be better to keep it 6.7.1?

JAVGan: Aside from `opm` upgrade everything is just a bugfix, so `6.7.1` makes sense IMO as we are mostly adding fixes...

I'm not sure if just a new `opm` binary counts as a "feature added" as we just upgraded a dependent component, but I'm fine with 6.8 if you believe this is a "feature" change

JAVGan: But honestly... I'm more up to `6.7.1` than `6.8` due to most of it being fixes

JAVGan (approved):

CHANGELOG.md
chandwanitulsi: yeah OPM update was making me inclined towards 6.8. I don't have strong objections to 6.7.1 as well. Happy to go with majority

chandwanitulsi (approved):

CHANGELOG.md
```diff
@@ -6,6 +6,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+- Fix dev env compose files
+- Fix bundle_replacements bug in regenerate-bundle request
+- Upgrade OPM to 1.26.2 in dev env
+- Fix buildah bud command retries
```
chandwanitulsi, line 13: I just realized we missed adding "Adding retry and port check when opm serve/opm registry serve is called" in changelog

PR #454, opened 2022-11-07
Retry decorators replaced by tenacity

xDaile:
The retry library is replaced by the tenacity library due to lack of maintenance.


iib/workers/tasks/build.py
```diff
@@ -54,11 +61,18 @@
...
+    retry=retry_if_exception_type(ExternalServiceError),
+    stop=stop_after_attempt(get_worker_config().iib_total_attempts),
+    wait=wait_chain(
+        *[
+            wait_fixed(
+                get_worker_config().iib_retry_delay + (x * get_worker_config().iib_retry_jitter)
+            )
+            for x in range(get_worker_config().iib_total_attempts)
+        ]
+    ),
```
xDaile, line 75: Here could be used something simpler.
Currently, the waiting times are 5, 10, 15, 20, ...
If we used for example wait_exponential(multiplier=4)`, waiting times would be 4, 8, 16, 16, 32.

iib/workers/tasks/build.py
```diff
@@ -126,7 +140,12 @@ def _cleanup() -> None:
     reset_docker_config()
 
 
-@retry(exceptions=IIBError, tries=get_worker_config().iib_total_attempts, logger=log)
+@retry(
```
yashvardhannanavati, line 143: Let's add backoff here as well?

iib/workers/tasks/build.py
```diff
@@ -635,7 +664,12 @@ def _update_index_image_build_state(
     update_request(request_id, payload, exc_msg)
 
 
-@retry(exceptions=IIBError, tries=get_worker_config().iib_total_attempts, logger=log)
+@retry(
```
yashvardhannanavati, line 667: How about adding back_off factor here as well?

iib/workers/tasks/build.py
```diff
@@ -664,7 +698,12 @@ def _push_image(request_id: int, arch: str) -> None:
         _skopeo_copy(destination, destination, exc_msg=exc_msg)
 
 
-@retry(exceptions=IIBError, tries=get_worker_config().iib_total_attempts, logger=log)
+@retry(
+    before_sleep=before_sleep_log(log, logging.WARNING),
+    reraise=True,
+    retry=retry_if_exception_type(IIBError),
+    stop=stop_after_attempt(get_worker_config().iib_total_attempts),
+)
```
yashvardhannanavati, line 706: Same as above

yashvardhannanavati (changes requested):

requirements.txt
```diff
@@ -378,10 +377,6 @@ psycopg2-binary==2.9.3 \
     --hash=sha256:e847774f8ffd5b398a75bc1c18fbb56564cda3d629fe68fd81971fece2d3c67e \
     --hash=sha256:ffb7a888a047696e7f8240d649b43fb3644f14f0ee229077e7f6b9f9081635bd
     # via iib (setup.py)
-py==1.11.0 \
-    --hash=sha256:51c75c4126074b472f746a24399ad32f6053d1b34b68d2fa41e558e6f4a98719 \
-    --hash=sha256:607c53218732647dff4acdfcd50cb62615cedf612e72d1724fb1a0cc6405b378
-    # via retry
```
xDaile, line 384: I know there was some recent change, which added especially this version of py, however this file I generated. Should I edit that somehow?

iib/workers/tasks/build.py
```diff
@@ -54,11 +61,18 @@
...
-    exceptions=ExternalServiceError,
-    tries=get_worker_config().iib_total_attempts,
-    delay=get_worker_config().iib_retry_delay,
-    jitter=get_worker_config().iib_retry_jitter,
-    logger=log,
+    before_sleep=before_sleep_log(log, logging.WARNING),
+    reraise=True,
+    retry=retry_if_exception_type(ExternalServiceError),
+    stop=stop_after_attempt(get_worker_config().iib_total_attempts),
+    wait=wait_chain(
```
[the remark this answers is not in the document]
xDaile: I wanted to create exactly the same functionality, where currently the waiting times are 5, 10, 15, 20, ...
If we used for example wait_exponential(multiplier=4)`, waiting times would be 4, 8, 16, 16, 32. Is that okay?

lipoja (dismissed):

iib/workers/tasks/utils.py
```diff
@@ -634,11 +641,11 @@ def set_registry_auths(registry_auths: Optional[Dict[str, Any]]) -> Generator:
...
-    exceptions=IIBError,
-    tries=get_worker_config().iib_total_attempts,
-    delay=5,
-    backoff=2,
-    logger=log,
+    before_sleep=before_sleep_log(log, logging.WARNING),
+    reraise=True,
+    retry=retry_if_exception_type(IIBError),
+    stop=stop_after_attempt(get_worker_config().iib_total_attempts),
+    wait=wait_chain(wait_exponential(multiplier=5)),
```
chandwanitulsi, line 648: How about using iib_retry_delay here instead of hard coding?

chandwanitulsi (approved):
Minor change, other than that looks good

iib/workers/tasks/utils.py
xDaile: I introduced iib_retry_multiplier into config, as delay and jitter have different meanings. Here, I forgot to use the variable, however, I already pushed the repair.

iib/workers/config.py
```diff
@@ -60,6 +60,7 @@ class Config(object):
     iib_total_attempts: int = 5
     iib_retry_delay: int = 5
     iib_retry_jitter: int = 5
+    iib_retry_multiplier: int = 5
```
yashvardhannanavati, line 63: We probably want to add this to README.md

iib/workers/tasks/build.py
```diff
@@ -54,11 +61,14 @@
...
 @retry(
-    exceptions=ExternalServiceError,
-    tries=get_worker_config().iib_total_attempts,
-    delay=get_worker_config().iib_retry_delay,
-    jitter=get_worker_config().iib_retry_jitter,
-    logger=log,
+    before_sleep=before_sleep_log(log, logging.WARNING),
+    reraise=True,
+    retry=retry_if_exception_type(ExternalServiceError),
+    stop=stop_after_attempt(get_worker_config().iib_total_attempts),
```
yashvardhannanavati, line 67: Maybe move the `get_worker_config()` call at the top of this file and use the same variable everywhere? That way, we only call the function once

yashvardhannanavati (changes requested):
Couple of minor comments

iib/workers/config.py
xDaile: Thanks, please check if the description of the constant in the README.md is okay.

lipoja (approved):

yashvardhannanavati (approved):

PR #455, opened 2022-11-22
Remove misleading log from API response

yashvardhannanavati:
CLOUDDST-15929

### DIFF
--- a/iib/workers/tasks/utils.py
+++ b/iib/workers/tasks/utils.py
@@ -756,6 +756,8 @@ def run_cmd(
     response: subprocess.CompletedProcess = subprocess.run(cmd, **params)
 
     if strict and response.returncode != 0:
+        if set(['buildah', 'manifest', 'rm']) <= set(cmd) and 'image not known' in response.stderr:
+            raise IIBError('Manifest list not found locally.')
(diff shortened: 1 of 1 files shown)


iib/workers/tasks/utils.py
```diff
@@ -756,6 +756,9 @@ def run_cmd(
     response: subprocess.CompletedProcess = subprocess.run(cmd, **params)
 
     if strict and response.returncode != 0:
+        if set(['buildah', 'manifest', 'rm']) <= set(cmd):
+            if 'image not known' in response.stderr:
```
xDaile, line 760: Could those conditions be in one if?
`if set(['buildah', 'manifest', 'rm']) <= set(cmd) and 'image not known' in response.stderr:`

xDaile (dismissed):

iib/workers/tasks/utils.py
yashvardhannanavati: yes, we could combine it into a single if

yashvardhannanavati: Done

xDaile (approved):

lipoja (approved):

PR #456, opened 2022-12-02
Increase wait time for grpc calls

yashvardhannanavati:
Refers to CLOUDDST-16641


yashvardhannanavati:
This resolves the issue where IIB times out after waiting for the index to be served. I verified this change as working.

yashvardhannanavati:

PR #458, opened 2022-12-05
Enable weekly rebuild of IIB images and manual trigger of GH action

lipoja:
Enable weekly rebuild of IIB images and manual trigger of GitHub action to build and push images

[CLOUDDST-12355]


.github/workflows/build.yml
```diff
@@ -2,27 +2,41 @@
...
 
 jobs:
-  # this job will only be triggered when release is published
+  # this job will only be triggered when on of the above trigger condition is met
   deployments:
     name: Build and Push to quay.io
     runs-on: ubuntu-latest
 
     steps:
+      - name: Get latest tag
```
yashvardhannanavati, line 21: Does this mean any new changes will be added to the last released tag?

lipoja: This will get name of the latest tag which was released on (GitHub). 
That is set by `releases-only: true` and to avoid selecting other tags it should have a prefix `v` as we are used to set for our version e.g. `v6.7.2`

yashvardhannanavati: Aha, makes sense. thank you!

yashvardhannanavati (approved):

.github/workflows/build.yml
```diff
@@ -2,27 +2,41 @@
...
     types: [published]
+  # Allow manual trigger
+  workflow_dispatch:
+  # Rebuild every Monday at 4:30 UTC
+  schedule:
+    - cron: '30 4 * * 1'
 
 jobs:
-  # this job will only be triggered when release is published
+  # this job will only be triggered when on of the above trigger condition is met
```
JAVGan, line 15: Just a spell fix

```suggestion
  # this job will only be triggered when one of the above trigger condition is met
```

.github/workflows/build.yml
```diff
@@ -2,27 +2,41 @@
...
+          repository: release-engineering/iib
+          releases-only: true
+          prefix: 'v'
+        id: iibtag
+
       - name: Checkout code
         uses: actions/checkout@v2
         with:
-          ref: ${{ github.event.release.tag_name }}
+          ref: ${{ github.event.release.tag_name || steps.iibtag.outputs.tag}}
```
JAVGan, line 32: Just to keep the consistency
```suggestion
          ref: ${{ github.event.release.tag_name || steps.iibtag.outputs.tag }}
```

JAVGan (commented):
Small things to fix, the rest LGTM

JAVGan (approved):

PR #459, opened 2022-12-07
Using ubi8 for iib messaging container image

lipoja:
This will help us to reduce security warnings on queay.io

[CLOUDDST-12355]

### DIFF
--- a/docker/message_broker/Dockerfile
+++ b/docker/message_broker/Dockerfile
@@ -1,8 +1,11 @@
 # Based from https://github.com/rmohr/docker-activemq
-FROM openjdk:8-jre
+FROM registry.access.redhat.com/ubi8/ubi:latest
 
 ENV ACTIVEMQ_VERSION 5.15.12
 ENV ACTIVEMQ_HOME /opt/activemq
+
+# keeping container image updated (security updates)
+RUN  dnf install java-1.8.0-openjdk maven -y && dnf update -y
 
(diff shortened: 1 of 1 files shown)


JAVGan:
I believe you can squash the commits, right?

I see the 013375caf54b7cfd29f4b54e51572d363ce5b6bb to run `apt update` which is later on replaced by `dnf`

lipoja:
Oh noeas, my bad. Thanks @JAVGan

docker/message_broker/Dockerfile
```diff
@@ -1,9 +1,12 @@
 # Based from https://github.com/rmohr/docker-activemq
-FROM openjdk:8-jre
+FROM registry.access.redhat.com/ubi8/ubi:latest
 
 ENV ACTIVEMQ_VERSION 5.15.12
 ENV ACTIVEMQ_HOME /opt/activemq
 
+# keeping container image updated (security updates)
+RUN  dnf install java-1.8.0-openjdk maven -y && dnf update -y
```
chandwanitulsi, line 8: Hey Jan,
Do you know if we specifically need Java 1.8?, I believe we have the latest Java 11 and Java 17 available as well.
I am thinking that maybe using latest java will give less security issues

lipoja: That is good point! I will check that.

PR #464, opened 2022-12-14
Replacing Travis tests with GitHub Actions

lipoja:
Due to recent and increasing security issues on Travis CI we should move to GitHub Actions.

### DIFF
--- a/.github/workflows/run_tox.yml
+++ b/.github/workflows/run_tox.yml
@@ -1,0 +1,42 @@
+---
+name: Run tox
+
+on:
+  pull_request:
+  workflow_dispatch:
+  push:
+    branches:
+      - "master"
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        # https://raw.githubusercontent.com/actions/python-versions/main/versions-manifest.json
+        python-version: ["3.8", "3.9"]
+    steps:
+      - uses: actions/checkout@v3
+      - name: Set up Python ${{ matrix.python-version }}
+        uses: actions/setup-python@v4
+        with:
+          python-version: ${{ matrix.python-version }}
+      - name: Install dependencies
+        run: |
+          sudo apt install libkrb5-dev -y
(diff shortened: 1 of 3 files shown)


lipoja:
Tests are failing due to `ERROR: safety: commands failed`. There is issue in sqlachemy, however it is fixed only in sqlalchemy 2.0.0b1+ (beta version). I would wait for stable release and fix it in different PR.

tox.ini
```diff
@@ -74,6 +77,13 @@ commands =
 deps = 
     -rrequirements-test.txt
 
+[testenv:unit_tests]
```
chandwanitulsi, line 80: Can you please explain why do we need a separate unit_tests env? It looks like this is doing the same as testenv:py38/39.
May I also ask if unit_tests should be mentioned above in envlist on line 4?

lipoja: Yes it is doing almost the same thing. The only difference is that `basepython` is not set for this one so it will take the one installed on the system. And for us it means it will pick up Python3.8 and Python3.9 as defined on line 17.
It is not mentioned in `envlist` so it is not automatically triggered when you run `tox` without parameters.

.github/workflows/run_tox.yml
```diff
@@ -0,0 +1,42 @@
...
+on:
+  pull_request:
+  workflow_dispatch:
+  push:
+    branches:
+      - "master"
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
```
yashvardhannanavati, line 13: Any reasons for not using fedora?

yashvardhannanavati (dismissed):

tox.ini
chandwanitulsi: Thank you for explaining. When I see the tox run for 3.8, I don't see the tests ran for python 3.8, not sure whats wrong there.
Also should we disable safety if it's failing out CI for some time?

lipoja: The run of on 3.8 was canceled because 3.9 failed. 
We can disable safety for now.

PR #465, opened 2022-12-15
Build opm cache locally to enable multi-arch builds

yashvardhannanavati:
This change is triggered by an OPM upgrade which introduced cache for FBC indexes. More specifically, OPM > 1.26.2

Refers to CLOUDDST-16852


yashvardhannanavati:

iib/workers/tasks/opm_operations.py
```diff
@@ -415,6 +420,59 @@ def opm_generate_dockerfile(
...
+
+    """
+    from iib.workers.tasks.utils import run_cmd
+
+    cmd = [
+        'opm',
+        'serve',
+        os.path.abspath(fbc_dir),
+        f'--cache-dir={local_cache_path}',
+        '--cache-only',
```
lipoja, line 461: Note: Using this tag means that this patch depends on OPM update issue since we are using v1.23.2 in our environments and this parameter was introduced later.

iib/workers/tasks/opm_operations.py
```diff
@@ -415,6 +420,59 @@ def opm_generate_dockerfile(
...
+    :param str local_cache_path: path to the locally generated cache.
+    :return: Returns path to generated cache
+    :rtype: str
+    :raises: IIBError when cache was not generated
+
+    """
+    from iib.workers.tasks.utils import run_cmd
+
+    cmd = [
+        'opm',
```
lipoja, line 457: Also I am not sure if it it still applies but I remember that @joelanford said in CLOUDDST-15194:
> modify the generated dockerfile such that the cache can be copied into the image rather than generated during the image build. We discussed that the version of opm used to generate the cache must match the version of opm in the binary image.

It means we should extract opm from the binary_image (or use opm based on from_index version)
In code we are using still the same opm version. I just wanted to point this out for discussion.

yashvardhannanavati: Thanks for the review Jan! We did talk to Joe and it looks like the latest opm version 1.26.2 will work for building both 4.11 and 4.12 indexes. In earlier indexes, the new line is not introduced (verified locally). So this change along with OPM upgrade should be enough to unblock them

iib/workers/tasks/opm_operations.py
```diff
@@ -415,6 +420,59 @@ def opm_generate_dockerfile(
...
+    :param str dockerfile_path: path to the generated Dockerfile.
+    :param str local_cache_path: path to the locally generated cache.
+    """
+    with open(dockerfile_path, 'r') as f:
+        file_data = f.read()
+
+    file_data = file_data.replace(
+        'RUN ["/bin/opm", "serve", "/configs", "--cache-dir=/tmp/cache", "--cache-only"]',
+        f'COPY {local_cache_path} /tmp/cache',
+    )
```
JAVGan, line 436: Question: Are we just replacing the local Dockerfile but not building the final image with this Dockerfile, right?

Because I believe the final image must serve from its own cache and here we're replacing the `RUN` serve command with just a `COPY` of contents from our cache to the final image cache...

My fear is that when we build the final image it will not have the correct `RUN` line...

yashvardhannanavati: good question, I don't think the final index needs to have the `RUN` line since the RUN line was actually building the cache. We are directly copying the built cache into the location OPM expects it to be. That's my understanding. 

I'll request Joe Lanford to take a look at this too just to be sure

iib/workers/tasks/opm_operations.py
yashvardhannanavati: Yes, that's right. The OPM upgrade commit and this one has to go out together for this fix to work

iib/workers/tasks/opm_operations.py
lipoja: @JAVGan We are not replacing the run of `opm serve` - it is still triggered as CMD inside the Dockerfile.
This replaces only generation of cache files (see `--cache-only` parameter).

iib/workers/tasks/opm_operations.py
chandwanitulsi: Good point Jan. We should ensure that both opm update and this change should go in the same release.

lipoja:
LGTM, but it would be nice if Joe had a quick look as well.

iib/workers/tasks/opm_operations.py
```diff
@@ -406,6 +406,11 @@ def opm_generate_dockerfile(
             log.info('Rewriting Dockerfile %s with newly generated by opm.', dockerfile_path)
         os.rename(dockerfile_path_opm_default, dockerfile_path)
 
+    local_cache_path = os.path.join(base_dir, 'cache')
+    if not os.path.exists(local_cache_path):
```
joelanford, line 410: Is there a case where the `local_cache_path` would already exist? If it does exist, how do we know if it is synced from the FBC we're about to build the image with?

joelanford: I'm pretty sure it's safe to just always run `/bin/opm serve /configs --cache-dir=/tmp/cache --cache-only`, even if the cache dir already exists. It will check cache integrity and only rebuild if hashes don't match.

iib/workers/tasks/opm_operations.py
joelanford: Yep, @yashvardhannanavati and @lipoja are correct!

iib/workers/tasks/opm_operations.py
```diff
@@ -415,6 +420,59 @@ def opm_generate_dockerfile(
...
+    :param str local_cache_path: path to the locally generated cache.
+    """
+    with open(dockerfile_path, 'r') as f:
+        file_data = f.read()
+
+    file_data = file_data.replace(
+        'RUN ["/bin/opm", "serve", "/configs", "--cache-dir=/tmp/cache", "--cache-only"]',
+        f'COPY {local_cache_path} /tmp/cache',
+    )
+
```
joelanford, line 437: Unrelated question for this replacement. Is there a way to detect if the replacement didn't replace anything? In the future, if the dockerfile generation changes again, it would be nice if the failure message is something like: "failed to replace cache building execution in dockerfile" rather than the eventual/cryptic: "exec /bin/opm: exec format error" error that would happen during the image build.

iib/workers/tasks/opm_operations.py
JAVGan: Thanks!

iib/workers/tasks/opm_operations.py
yashvardhannanavati: Added a check to confirm the `COPY` line is added to the file and throw a more comprehendible error message if the replacement fails.

iib/workers/tasks/opm_operations.py
```diff
@@ -406,15 +406,91 @@ def opm_generate_dockerfile(
...
+    local_cache_path = os.path.join(base_dir, 'cache')
+    generate_cache_locally(base_dir, fbc_dir, local_cache_path)
+    insert_cache_into_dockerfile(dockerfile_path, local_cache_path)
+
     db_path = get_worker_config()['hidden_index_db_path']
     rel_path_index_db = os.path.relpath(index_db, base_dir)
-    with open(dockerfile_path, 'a') as f:
+    with open(dockerfile_path, 'a+') as f:
         f.write(f'\nADD {rel_path_index_db} {db_path}\n')
+        verify_cache_insertion_into_dockerfile(f.readlines(), local_cache_path)
```
lipoja, line 417: I think this does not work as expected. When you are appending you move the "cursor" at the end of file. Therefore call of `f.readlines()` is incorrect and does not return what you are expecting. 

I would not write and read from the same file descriptor.

iib/workers/tasks/opm_operations.py
```diff
@@ -406,15 +406,91 @@ def opm_generate_dockerfile(
...
 
     log.info("Dockerfile was successfully generated.")
     return dockerfile_path
 
 
+def verify_cache_insertion_into_dockerfile(file_list: list, local_cache_path: str) -> None:
+    """
+    Verify local cache was copied into the dockerfile.
```
lipoja, line 425: Nitpick: We are not verifying if it was copied, we verify if the COPY command is present in Dockerfile.

tests/test_workers/test_tasks/test_opm_operations.py
```diff
@@ -166,9 +167,13 @@ def test_opm_migrate(
 
 @pytest.mark.parametrize("dockerfile", (None, 'index.Dockerfile'))
 @mock.patch('iib.workers.tasks.utils.run_cmd')
-def test_opm_generate_dockerfile(mock_run_cmd, tmpdir, dockerfile):
+@mock.patch('iib.workers.tasks.opm_operations.generate_cache_locally')
+@mock.patch('iib.workers.tasks.opm_operations.insert_cache_into_dockerfile')
+@mock.patch('iib.workers.tasks.opm_operations.verify_cache_insertion_into_dockerfile')
+def test_opm_generate_dockerfile(mock_vciid, mock_icid, mock_gcl, mock_run_cmd, tmpdir, dockerfile):
```
lipoja, line 173: I am not sure if we should test also `mock_vciid` 
The test is missing so we did not discover the incorrect reading of Dockerfile.

lipoja (changes requested):

iib/workers/tasks/opm_operations.py
yashvardhannanavati: nice catch! changed it :)

iib/workers/tasks/opm_operations.py
yashvardhannanavati: Aha, you're right. great catch

lipoja (approved):

PR #466, opened 2022-12-15
Upgrade gssapi to 1.8.2

yashvardhannanavati:


yashvardhannanavati:
Dependabot fails on gssapi. Upgrading gssapi to check if it helps with dependabot builds. @release-engineering/exd-guild-hello-operator PTAL

chandwanitulsi (approved):

lipoja:
@yashvardhannanavati where do you see the dependabot issue? 
I see that we should update `certifi`

PR #471, opened 2022-12-22
Fix user permissisons on index cache directory

yashvardhannanavati:


iib/workers/tasks/opm_operations.py
```diff
@@ -427,7 +427,7 @@ def verify_cache_insertion_edit_dockerfile(file_list: list) -> None:
     :raises: IIBError when the Dockerfile edit is unsuccessful.
     """
     copied_cache_found = False
-    match_str = f'COPY cache /tmp/cache'
+    match_str = f'COPY --chown=1001 cache /tmp/cache'
```
pbortlov, line 430: do you really need it for matching the string? would i match anything?

yashvardhannanavati: Sorry I didn't get it. The match_str is to make sure the replacement was successful. It's just verifying the correct line was added

pbortlov: this line is adding "--chown=1001" to be matched in the line below "if match_str in line". Is it a specific case to look for "chown" in dockerfile? my concern is that it won't be matched/find at all.

yashvardhannanavati: yes, it is a specific case to look for the entire line. If it is not found, an error is raised

pbortlov: thanks for the response

pbortlov (approved):

PR #477, opened 2023-01-05
Mypy fix and enablement

xDaile:
Fix for few mypy errors, and also enablement of mypy check. 
Related to: https://issues.redhat.com/browse/CLOUDDST-16919

Mypy repaired bugs:
- return inside of cycle removed
- convert Optional[str] value to str value added, - str part has to be str as the return type is imported from flask, 
- add bundle-replacements to RegenerateBundleRequestResponse

Allowed mypy check

### DIFF
--- a/iib/web/api_v1.py
+++ b/iib/web/api_v1.py
@@ -354,7 +354,7 @@ def get_builds() -> flask.Response:
 
     :rtype: flask.Response
     """
-    batch_id = flask.request.args.get('batch')
+    batch_id: Optional[str] = flask.request.args.get('batch')
     state = flask.request.args.get('state')
     verbose = str_to_bool(flask.request.args.get('verbose'))
     max_per_page = flask.current_app.config['IIB_MAX_PER_PAGE']
@@ -376,8 +376,8 @@ def get_builds() -> flask.Response:
 
     if batch_id is not None:
         query_params['batch'] = batch_id
-        batch_id = Batch.validate_batch(batch_id)
-        query = query.filter_by(batch_id=batch_id)
+        batch_id_checked: int = Batch.validate_batch(batch_id)
+        query = query.filter_by(batch_id=batch_id_checked)
 
     if request_type:
         query_params['request_type'] = request_type
@@ -481,7 +481,7 @@ def add_bundles() -> Tuple[flask.Response, int]:
     :rtype: flask.Response
     :raise ValidationError: if required parameters are not supplied
     """
-    payload: AddRequestPayload = flask.request.get_json()
+    payload: AddRequestPayload = cast(AddRequestPayload, flask.request.get_json())
     if not isinstance(payload, dict):
         raise ValidationError('The input data must be a JSON object')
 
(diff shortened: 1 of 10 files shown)


iib/workers/tasks/opm_operations.py
```diff
@@ -112,13 +114,15 @@ def opm_registry_serve(db_path: str) -> Tuple[int, subprocess.Popen]:
         try:
             cmd = ['opm', 'registry', 'serve', '-p', str(port), '-d', db_path, '-t', '/dev/null']
             cwd = os.path.dirname(db_path)
-            return (
+            result = (
```
chandwanitulsi, line 117: Minor: Can you share if this change was done to refactor the code or if it was impacting the mypy checks as well?

xDaile: All of the changes in the first commit were done, because of mypy errors. For example in this one, Mypy was expecting some return value after the "possible" end of the loop - which could not happen anyway. 

Should I add some more comments about the changes, or is it okay like this?

iib/web/api_v1.py
```diff
@@ -376,8 +376,8 @@ def get_builds() -> flask.Response:
 
     if batch_id is not None:
         query_params['batch'] = batch_id
-        batch_id = Batch.validate_batch(batch_id)
-        query = query.filter_by(batch_id=batch_id)
+        batch_id_checked: int = Batch.validate_batch(batch_id)
+        query = query.filter_by(batch_id=batch_id_checked)
```
xDaile, line 380: New variable is needed, because of the variable type change from Optional[str] to int by validate_batch function.

iib/web/errors.py
```diff
@@ -2,11 +2,12 @@
 from typing import List, NoReturn
 
 import kombu.exceptions
-from flask import current_app, jsonify, Request, Response
+from flask import current_app, jsonify, Response
 from werkzeug.exceptions import HTTPException
 
 from iib.exceptions import IIBError, ValidationError
 from iib.web import messaging, db
+from iib.web.models import Request
```
xDaile, line 10: Based on line 50: `request.add_state('failed', 'The scheduling of the request failed')`, I suppose that we want to use class Request from iib.web.models which inherit from flask.Request. (It is possible that I introduced this import when I was adding the mypy types to the iib).

iib/web/errors.py
```diff
@@ -21,9 +22,11 @@ def json_error(error: Exception) -> Response:
         if error.code == 404:
             msg = 'The requested resource was not found'
         else:
-            msg = error.description
+            msg = str(error.description)
         response = jsonify({'error': msg})
-        response.status_code = error.code
+        # None cannot be set to response.status_code
+        if error.code:
+            response.status_code = error.code
```
xDaile, line 29: Mypy error about possible setting up response.status_code to None value.

iib/web/models.py
```diff
@@ -694,14 +694,16 @@ def validate_batch(batch_id: int) -> int:
...
         error_msg = 'The batch must be a positive integer'
         if isinstance(batch_id, str):
             try:
                 rv = int(batch_id)
             except ValueError:
                 raise ValidationError(error_msg)
-        elif not isinstance(batch_id, int):
+        elif isinstance(batch_id, int):
+            rv = batch_id
+        else:
```
xDaile, line 706: Mixed usage of types (int and str). Now `rv` is always int.

iib/workers/tasks/opm_operations.py
chandwanitulsi: Oh these comments are more than enough. Thank you :)

chandwanitulsi (approved):

xDaile:
> 

Yes, I was able to create-empty-index, I can also try other API calls, if needed.

PR #484, opened 2023-01-17
fbc-operations worker implementation

chandwanitulsi:


iib/workers/tasks/opm_operations.py
```diff
@@ -499,6 +499,34 @@ def generate_cache_locally(base_dir: str, fbc_dir: str, local_cache_path: str) -
...
+    # TODO get fbc path from conf?
+    # store the fbc_fragment at /tmp/iib-**/fbc-fragment
+    fbc_fragment_path = os.path.join(temp_dir, 'fbc_fragment')
+    os.makedirs(os.path.dirname(fbc_fragment_path), exist_ok=True)
+
+    from iib.workers.tasks.build import _copy_files_from_image
+
+    # Verify from opm team if the dir path will be available in a label
+    # Verify this will this be catalog or configs?
+    _copy_files_from_image(fbc_fragment,'/configs', fbc_fragment_path)
```
lipoja, line 518: To be persistent and not use hard-coded string we might use:`fbc_dir_path = get_image_label(fbc_fragment, 'operators.operatorframework.io.index.configs.v1')`  
If it is in the config dir, if not I would vote for agreeing on setting label as we do for configs instead of hard-coding path.

iib/workers/tasks/opm_operations.py
```diff
@@ -738,3 +766,63 @@ def opm_create_empty_fbc(
...
+
+    with set_registry_token(overwrite_from_index_token, from_index, append=True):
+        index_db_path = get_hidden_index_database(from_index=from_index, base_dir=temp_dir)
+
+    _opm_registry_rm(index_db_path, [fbc_fragment_operator], temp_dir)
+
+    fbc_catalog_dir = opm_migrate(index_db=index_db_path, base_dir=temp_dir)
+
+    # copy fbc_fragment yaml to from_index catalog
+    fbc_fragment_yaml = os.path.join(fbc_fragment_path, fbc_fragment_operator, "catalog.yaml")
```
lipoja, line 812: And I think we can't expect that it will be always `.yaml` file. From what I know it can be `.json` as well.

chandwanitulsi: changed the logic to copy the file directly

iib/workers/tasks/opm_operations.py
chandwanitulsi: since we are still not sure if the path will be set in a label, I added a worker config for that

iib/workers/tasks/opm_operations.py
```diff
@@ -499,6 +499,34 @@ def generate_cache_locally(base_dir: str, fbc_dir: str, local_cache_path: str) -
         raise IIBError(error_msg)
 
 
+def extract_fbc_fragment_config(temp_dir: str, fbc_fragment: str) -> None:
```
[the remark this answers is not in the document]
chandwanitulsi: Updated

iib/workers/tasks/opm_operations.py
```diff
@@ -738,3 +751,110 @@ def opm_create_empty_fbc(
...
+    Add FBC fragment to from_index image.
+
+    This only produces the index.Dockerfile file and does not build the container image.
+
+    :param str temp_dir: the base directory to generate the database and index.Dockerfile in.
+    :param str from_index: the pull specification of the container image containing the index that
+        the index image build will be based from.
+    :param str binary_image: the pull specification of the container image where the opm binary
+        gets copied from. This should point to a digest or stable tag.
+    :param str fbc_fragment: the fbc fragment to be added in from_index.
```
lipoja, line 773: `the pull specification of  ...`

iib/workers/tasks/fbc_utils.py
```diff
@@ -69,3 +69,29 @@ def get_hidden_index_database(from_index: str, base_dir: str) -> str:
...
+
+    log.info("Extracting the fbc_fragment's catalog from  %s", fbc_fragment)
+    # store the fbc_fragment at /tmp/iib-**/fbc-fragment
+    fbc_fragment_path = os.path.join(temp_dir, 'fbc_fragment')
+    os.makedirs(os.path.dirname(fbc_fragment_path), exist_ok=True)
+
+    conf = get_worker_config()
+    _copy_files_from_image(fbc_fragment, conf['fbc_fragment_catalog_path'], fbc_fragment_path)
+
+    operator_packages = os.listdir(fbc_fragment_path)
```
lipoja, line 93: Have you tried this one? I am not sure if `_copy_files_from_image` does not copy whole directory so it means it would list only `catalog` folder. But I might be wrong here.

iib/workers/tasks/opm_operations.py
```diff
@@ -738,3 +751,110 @@ def opm_create_empty_fbc(
...
+        operator_package=fragment_operator,
+        overwrite_from_index_token=overwrite_from_index_token,
+    )
+
+    from_index_configs_dir = get_catalog_dir(from_index=from_index, base_dir=temp_dir)
+
+    if is_operator_in_db:
+        log.info('Removing %s from %s index.db ', fragment_operator, from_index)
+        _opm_registry_rm(index_db_path, [fragment_operator], temp_dir)
+        migated_catalog_dir, _ = opm_migrate(
```
lipoja, line 790: I am not sure but doesn't it rewrite the `from_index_configs_dir` data?

iib/workers/tasks/opm_operations.py
```diff
@@ -738,3 +751,110 @@ def opm_create_empty_fbc(
...
+
+    # copy fragment_operator to from_index configs
+    fragment_opr_dest_path = os.path.join(from_index_configs_dir, fragment_operator)
+    if os.path.exists(fragment_opr_dest_path):
+        shutil.rmtree(fragment_opr_dest_path)
+    shutil.copytree(os.path.join(fragment_path, fragment_operator), fragment_opr_dest_path)
+
+    # Generate cache as part of opm_generate_dockerfile()?
+    local_cache_path = os.path.join(temp_dir, 'cache')
+    generate_cache_locally(temp_dir, from_index_configs_dir, local_cache_path)
```
lipoja, line 811: This might fail if cache was already generated. Should we delete it if it exist first?

iib/workers/tasks/opm_operations.py
lipoja: I am thinking about if we do not migrate it at all? 
Because we just removed the operator? Or this is just to be sure that if anything else was changed it is copied over?

iib/workers/tasks/opm_operations.py
```diff
@@ -738,3 +751,110 @@ def opm_create_empty_fbc(
...
+
+    opm_generate_dockerfile(
+        fbc_dir=from_index_configs_dir,
```
lipoja, line 814: Are we sure data needed for build (mentioned in Dockerfile) are at this location?
I am not sure  however just from the code I assume that the index.db is not at this location.

iib/workers/tasks/fbc_utils.py
chandwanitulsi: the fbc_fragment_path looks like /tmp/iib-1mnpn1yl/fbc_fragment, this has the content of catalog in fbc_fragment,
os.listdir on fbc_fragment_path displays operator_packages like,  ['example-operator']

iib/workers/tasks/opm_operations.py
chandwanitulsi: the location of from_index_configs_dir will be like_/tmp/iib-1mnpn1yl/configs < this has the configs content of from_index
and the migrated catalog has its own dir, i.e /tmp/iib-1mnpn1yl/catalog < resulted from migrated_catalog

chandwanitulsi: yeah, even i thought we might skip the migration and one copy step , but I thought we should add it as a caution

iib/workers/tasks/opm_operations.py
chandwanitulsi: from_index_configs_dir has the configs content of from_index, and the catalog from fbc_fragment was also added here, so we have all the required data in that dir.
I am not sure how index.db location will impact here, can you please share?

iib/workers/tasks/opm_operations.py
```diff
@@ -738,3 +751,123 @@ def opm_create_empty_fbc(
...
+        operator_package=fragment_operator,
+        overwrite_from_index_token=overwrite_from_index_token,
+    )
+
+    from_index_configs_dir = get_catalog_dir(from_index=from_index, base_dir=temp_dir)
+
+    log.info("The content of from_index located at %s", from_index_configs_dir)
+
+    if is_operator_in_db:
+        log.info('Removing %s from %s index.db ', fragment_operator, from_index)
```
yashvardhannanavati, line 792: I almost feel the code under this `if` block should be a separate helper function. wdyt?

iib/workers/tasks/opm_operations.py
```diff
@@ -738,3 +751,123 @@ def opm_create_empty_fbc(
...
+            index_db_path=index_db_path, operators=[fragment_operator], base_dir=temp_dir
+        )
+        migated_catalog_dir, _ = opm_migrate(
+            index_db=index_db_path,
+            base_dir=temp_dir,
+            generate_cache=False,
+        )
+        log.info("Migated catalog after removing from db at %s", migated_catalog_dir)
+
+        for operator_package in os.listdir(migated_catalog_dir):
```
yashvardhannanavati, line 803: Could you please add a comment here explaining this copy?

iib/workers/tasks/opm_operations.py
```diff
@@ -738,3 +751,123 @@ def opm_create_empty_fbc(
...
+            )
+
+    # copy fragment_operator to from_index configs
+    fragment_opr_dest_path = os.path.join(from_index_configs_dir, fragment_operator)
+    if os.path.exists(fragment_opr_dest_path):
+        shutil.rmtree(fragment_opr_dest_path)
+    shutil.copytree(os.path.join(fragment_path, fragment_operator), fragment_opr_dest_path)
+
+    # Generate cache as part of opm_generate_dockerfile()?
+    local_cache_path = os.path.join(temp_dir, 'cache')
```
yashvardhannanavati, line 817: I think we are duplicating this piece of code in a number of places. How about converting it into a helper function instead?

yashvardhannanavati (changes requested):

iib/web/api_v1.py
```diff
@@ -1141,6 +1139,7 @@ def fbc_operations() -> Tuple[flask.Response, int]:
     try:
         handle_fbc_operation_request.apply_async(
             args=args, link_error=error_callback, argsrepr=repr(safe_args), queue=celery_queue
+        )
```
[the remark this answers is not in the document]
lipoja: All `# TODO`s should be removed

iib/workers/tasks/opm_operations.py
lipoja: > 

This brings me the question what if the image has label `operators.operatorframework.io.index.configs.v1'` set 
 as `/catalog` . This might be legitimate and since we get the path from label I wold be cautious here.

iib/workers/tasks/build.py
```diff
@@ -1099,6 +1101,103 @@ def handle_rm_request(
     )
 
 
+@app.task
+@request_logger
+def handle_fbc_operation_request(
+    request_id: int,
```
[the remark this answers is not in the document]
chandwanitulsi: I don't mind at all ordering this but I thought we add the required parameters first and optional ones later, saw the same pattern in other handle* methods

iib/workers/tasks/fbc_utils.py
```diff
@@ -69,3 +69,31 @@ def get_hidden_index_database(from_index: str, base_dir: str) -> str:
     os.makedirs(os.path.dirname(base_db_file), exist_ok=True)
     _copy_files_from_image(from_index, conf['hidden_index_db_path'], base_db_file)
     return base_db_file
+
+
+def extract_fbc_fragment(temp_dir: str, fbc_fragment: str) -> Tuple[str, str]:
+    """
+    Extract operator from the from fbc_fragment image.
+
+    :param str temp_dir: base temp directory for IIB request
```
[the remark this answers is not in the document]
chandwanitulsi: base_dir sounds confusing, but that's just my personal opinion. Happy to go with the majority

iib/workers/tasks/opm_operations.py
chandwanitulsi: I also had the same thoughts while writing this, but later realized that in the if block there was just removal and copying of operator packages, so thought its too few steps to split. Any ideas are more than welcome :)

iib/workers/tasks/opm_operations.py
chandwanitulsi: on the from_index, the label will always be set as we manage and create that index, and that's always /configs. But we never know what the future holds, so I can add change it to be safer on our end.

for the fbc_fragment, we still haven't got any response on the exact location of the catalog, hence I added a config in the worker so if it needs to change, we can change it easily. I believe the location should be same for all fbc_fragments

lipoja:
@chandwanitulsi could you rebase on master branch? Thank you!

iib/workers/tasks/opm_operations.py
chandwanitulsi: Hey @lipoja , circling back on your question 

> "This brings me the question what if the image has label operators.operatorframework.io.index.configs.v1 set
> as /catalog . This might be legitimate and since we get the path from label I wold be cautious here."

Was your question for label in from_index or fbc_fragment? For fbc_fragment we don't know what will be label, and irrespective of the location, it can be specified in configs in https://github.com/release-engineering/iib/pull/484/files#diff-ca2cc097d7757c5b9d5ef4de2c4ed7d90e1431835c57c042a50d4cb8b35994c2R77 and even if the label or "config or catalog" it will not matter as the dir is copied to /tmp/iib-**/**fbc-fragment**

I was also thinking that it is very unlikely that the label will be set as "/catalog" for from_index as if that's done even the normal add request will not work, it will fail during opm_migrate or opm_migrate will overwrite the catalog folder

iib/workers/tasks/build.py
```diff
@@ -1099,6 +1101,103 @@ def handle_rm_request(
     )
 
 
+@app.task
```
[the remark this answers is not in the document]
chandwanitulsi: Done

iib/workers/tasks/opm_operations.py
chandwanitulsi: Done

iib/workers/tasks/fbc_utils.py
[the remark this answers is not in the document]
yashvardhannanavati: Let's keep it as is :)

iib/workers/tasks/build_fbc_operations.py
```diff
@@ -0,0 +1,125 @@
...
+        request_id,
+        arches,
+        from_index,
+        overwrite_from_index,
+        overwrite_from_index_token,
+        from_index_resolved,
+        add_or_rm=True,
+    )
+    set_request_state(
+        request_id, 'complete', 'The FBC fragment were successfully added in the index image'
```
lipoja, line 124: Typo: `The FBC fragment were` -> `fragment was` or `fragments were`?

iib/workers/tasks/fbc_utils.py
```diff
@@ -87,3 +87,32 @@ def merge_catalogs_dirs(src_config: str, dest_config: str):
...
+
+    log.info("Extracting the fbc_fragment's catalog from  %s", fbc_fragment)
+    # store the fbc_fragment at /tmp/iib-**/fbc-fragment
+    conf = get_worker_config()
+    fbc_fragment_path = os.path.join(temp_dir, conf['temp_fbc_fragment_path'])
+    # Copy fbc_fragment's catalog to /tmp/iib-**/fbc-fragment
+    _copy_files_from_image(fbc_fragment, conf['fbc_fragment_catalog_path'], fbc_fragment_path)
+
+    log.info("fbc_fragment extracted at %s", fbc_fragment_path)
+    operator_packages = os.listdir(fbc_fragment_path)
```
lipoja, line 111: Is our job to validate content? Test if fragment is valid (for example: test fragment does not contain only empty directory without json/yaml files) ...

iib/workers/tasks/opm_operations.py
```diff
@@ -744,3 +751,133 @@ def opm_create_empty_fbc(
...
+
+    is_package_in_index = False
+
+    log.info("Verifying if operator package %s exists in index %s", operator_package, from_index)
+
+    # check if operater package exists in hidden index.db
+    with set_registry_token(overwrite_from_index_token, from_index, append=True):
+        index_db_path = get_hidden_index_database(from_index=from_index, base_dir=base_dir)
+
+    port, rpc_proc = opm_registry_serve(db_path=index_db_path)
```
yashvardhannanavati, line 871: Lines 871 to 876 can be replaced with a call to `https://github.com/release-engineering/iib/blob/master/iib/workers/tasks/utils.py#L1212`

yashvardhannanavati (changes requested):

iib/workers/tasks/opm_operations.py
lipoja: Thanks

iib/workers/tasks/build_fbc_operations.py
chandwanitulsi: you are right it should be 'fragment was'

iib/workers/tasks/fbc_utils.py
chandwanitulsi: hmm, good point. I am not sure whether the built fbc_component will be directly used in IIB call or if there will be any validation tests on it before calling IIB. From what I know, the catalog.json/yaml file is actually created by using veneers (https://github.com/operator-framework/example-operator-index/blob/main/README.md?plain=1#L9), so I think it will be less likely that it will be empty, but it might be possible
How about we add this additional check in the next PR, by the time we will know how the Stonesoup pipeline is using the new endpoint?

lipoja: sure - it just popped in my mind so I raised this question

iib/workers/tasks/opm_operations.py
chandwanitulsi: so if I call grpcurl_get_db_data() here, it will call opm_serve_from_index() which will end up serving data from catalog dir(as we will have from_index as fbc), but we want to specifically call opm_registry_serve() to serve from db to check the existence of operator package in db, hence I had to duplicate some code.
Please let me know if I missed something

iib/workers/tasks/fbc_utils.py
chandwanitulsi: "hmm maybe we do not need test because it will fail on cache generation?" possible

iib/workers/tasks/opm_operations.py
lipoja: The question was for label for fbc_fragment directory. Yeah lets keep it this way - I agree

lipoja:
code looks OK

iib/workers/tasks/opm_operations.py
yashvardhannanavati: So, if I check the code here: https://github.com/release-engineering/iib/blob/master/iib/workers/tasks/opm_operations.py#L49, it will indeed call opm_registry_serve for FBC indexes. Am I missing something?

iib/workers/tasks/build_fbc_operations.py
chandwanitulsi: changed

iib/workers/tasks/opm_operations.py
```diff
@@ -744,3 +751,133 @@ def opm_create_empty_fbc(
...
+                dirs_exist_ok=True,
+            )
+
+    # copy fragment_operator to from_index configs
+    fragment_opr_dest_path = os.path.join(from_index_configs_dir, fragment_operator)
+    if os.path.exists(fragment_opr_dest_path):
+        shutil.rmtree(fragment_opr_dest_path)
+    log.info(
+        "Copying content of %s to %s",
+        os.path.join(fragment_path, fragment_operator),
```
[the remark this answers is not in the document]
chandwanitulsi: Done

iib/workers/tasks/opm_operations.py
yashvardhannanavati: oh shoot! I'm sorry. you're right

yashvardhannanavati (dismissed):

chandwanitulsi:
@lipoja @yashvardhannanavati, while going through other logs I realized that adding more state changes and setting fbc_fragment_resolved would be useful in this PR, hence added the new commit

lipoja (approved):

yashvardhannanavati (approved):

PR #485, opened 2023-01-23
Update Add API endpoint to work with both not opted-in and opted-in

lipoja:


chandwanitulsi (commented):
Code looks good to me, I will test it on my local env and approve

iib/workers/tasks/fbc_utils.py
```diff
@@ -69,3 +70,23 @@ def get_hidden_index_database(from_index: str, base_dir: str) -> str:
...
+    :rtype: str
+    """
+    for conf_dir in (src_config, dest_config):
+        if not os.path.isdir(conf_dir):
+            msg = f"config directory does not exist: {conf_dir}"
+            log.error(msg)
+            raise IIBError(msg)
+
+    log.info("Merging config folders: %s to %s", src_config, dest_config)
+    shutil.copytree(src_config, dest_config)
```
yashvardhannanavati, line 91: are we sure we can use `shutil.copytree` as is? 

From the docs, it looks like this only works if the destination doesn't exist (https://docs.python.org/3/library/shutil.html#shutil.copytree). IIUC, In our case, we want to overwrite the contents of the destination so we might have to specify `dirs_exist_ok =True` since the default is false.

yashvardhannanavati (changes requested):

iib/workers/tasks/fbc_utils.py
lipoja: You are right, thank you.   Added `dirs_exist_ok`

tests/test_workers/test_tasks/test_fbc_utils.py
```diff
@@ -86,3 +88,27 @@ def test_is_image_fbc(mock_si, skopeo_output, is_fbc):
 
     mock_si.return_value = skopeo_output
     assert is_image_fbc(image) is is_fbc
+
+
+@mock.patch('shutil.copytree')
+@mock.patch('os.path.isdir')
+def test_merge_catalogs_dirs(mock_isdir, mock_cpt, tmpdir):
```
yashvardhannanavati, line 95: We could test here that copytree works as expected and the permissions on the dirs are not changed when overwritten.

iib/workers/tasks/fbc_utils.py
```diff
@@ -69,3 +70,23 @@ def get_hidden_index_database(from_index: str, base_dir: str) -> str:
...
+    """
+    for conf_dir in (src_config, dest_config):
+        if not os.path.isdir(conf_dir):
+            msg = f"config directory does not exist: {conf_dir}"
+            log.error(msg)
+            raise IIBError(msg)
+
+    log.info("Merging config folders: %s to %s", src_config, dest_config)
+    shutil.copytree(src_config, dest_config, dirs_exist_ok=True)
+    return dest_config
```
yashvardhannanavati, line 92: Should we be returning something that was passed as an argument while calling?

yashvardhannanavati: Maybe this should return nothing

yashvardhannanavati (changes requested):

iib/workers/tasks/fbc_utils.py
lipoja: I think I agree with you - return removed

yashvardhannanavati (approved):

lipoja:
I had to replace `.removeprefix()` with `.replace()` because it is not supported in py38

yashvardhannanavati (approved):

chandwanitulsi:
Hey @lipoja, I noticed this error when I was testing something for fbc changes,
`FileNotFoundError: [Errno 2] No such file or directory: '/tmp/iib-bgeqjlev/from_db'`, do you get this error on your end?
My add request was `curl -u: --negotiate -X POST -H "Content-Type: application/json" http://localhost:8080/api/v1/builds/add -d '{"bundles": ["registry-proxy.engineering.redhat.com/rh-osbs/lgallett-bundle:v1.0-9"],  "from_index": "registry-proxy.engineering.redhat.com/rh-osbs/iib-pub-pending:v4.12", "binary_image":"registry-proxy.engineering.redhat.com/rh-osbs/openshift-ose-operator-registry:v4.12"}'` on master branch
If I add `os.makedirs(os.path.join(temp_dir, 'from_db'))` or similar, this error doesn't come
Can you please check on your end?

PR #505, opened 2023-03-13
Move _cleanup function to the end of the build

xDaile:


xDaile:
Add cleanup test for general:error_callback method.

iib/workers/tasks/build.py
```diff
@@ -1055,7 +1055,6 @@ def handle_rm_request(
     :param list build_tags: List of tags which will be applied to intermediate index images.
     :raises IIBError: if the index image build fails.
     """
-    _cleanup()
```
yashvardhannanavati, line 1058: Should we keep calling the _cleanup() function at the start as well? 

Here's what I'm thinking. What if for some reason, the worker shuts down before giving the request time to complete/fail. When it comes back up and a new request is queued, the images from that request that was shut abruptly will still be in the VM. This way, we make sure the working env for a request is always clean

yashvardhannanavati: @lipoja @zanssa @xDaile thoughts?

xDaile: In my opinion, it depends on how much we are restricted by the space, for such a (hopefully) rare case. And, also, if we are willing to exchange a little bit of every request performance for a rare chance for space saving. 

Personally, I would keep cleanup just at the end, as I consider this scenario for really rare event, but correct me If I am wrong. 

With new request completed/failed it will be of course deleted anyway.

yashvardhannanavati: Let's wait for @lipoja and @zanssa 's thoughts before we merge this. :)

xDaile: I realised that it's may got an reason why cleanup was called before the request. 
Why would anyone in the first place, place cleanup before the request, not after the request? 
Then I realised, that maybe placing it before the request, was probably for some reliability reason.  
So, I am okay also with an option, where cleanup would be called also before the request.

zanssa: Out of curiosity @yashvardhannanavati, these images are built and tagged with the request id. correct? In the case mentioned above ( a worker was down for some reason), there would be a new built image with a new tag unless I am missing something. Can you explain to me the risk of building the images in that case?

lipoja: Yes, I agree we should keep the cleanup at the begging as well (as on the end of request)

zanssa: Yeah, I am okay to have it running before and after the build. Keeping it to run only before the build won't serve the purpose of the function. :)

lipoja (approved):

yashvardhannanavati (approved):

PR #507, opened 2023-03-23
Add logs for time tracking

xDaile:
During local testing, I did find that almost all the time of the request is taken, by the following line.
`rv = requests_auth_session.patch(request_url, json=payload, timeout=config.iib_api_timeout)`
However, on my local environment, the time of the update_request method was still under ~0.7s, which is far from 30s. 

As I am not able to test this, in an environment similar to what is in production/stage/QE, the best way to confirm my suspicion about the cause of timeout failing, would be probably to log the time of the patch method relative to the update_request method. 

I will fill out the issue to revert these changes later after the cause of the timeout will be found.


lipoja (approved):

chandwanitulsi (approved):
Can you please add the jira in the description if we have a jira associated with this change?

PR #509, opened 2023-04-18
Changing loading of user queue to accept new structure

lipoja:
CLOUDDST-18197

### DIFF
--- a/iib/web/api_v1.py
+++ b/iib/web/api_v1.py
@@ -1,5 +1,6 @@
 # SPDX-License-Identifier: GPL-3.0-or-later
 import copy
+import logging
 import os
 from datetime import datetime
 import time
@@ -468,11 +469,14 @@ def get_healthcheck() -> flask.Response:
     return flask.jsonify({'status': 'Health check OK'})
 
 
-def _get_user_queue(serial: Optional[bool] = False) -> Optional[str]:
+def _get_user_queue(
+    serial: Optional[bool] = False, from_index_pull_spec: Union[str, None] = None
+) -> Optional[str]:
     """
     Return the name of the celery task queue mapped to the current user.
 
     :param bool serial: whether or not the task must run serially
+    :param str from_index_pull_spec: index image pull-spec
     :return: queue name to be used or None if the default queue should be used
     :rtype: str or None
     """
@@ -486,10 +490,33 @@ def _get_user_queue(serial: Optional[bool] = False) -> Optional[str]:
     else:
         labeled_username = f'PARALLEL:{username}'
 
-    queue = flask.current_app.config['IIB_USER_TO_QUEUE'].get(labeled_username)
-    if not queue:
-        queue = flask.current_app.config['IIB_USER_TO_QUEUE'].get(username)
-    return queue
+    index_queue = flask.current_app.config['IIB_USER_TO_QUEUE'].get(labeled_username)
+    if not index_queue:
+        index_queue = flask.current_app.config['IIB_USER_TO_QUEUE'].get(username)
+
+    if index_queue is None:
+        return index_queue
+
+    if isinstance(index_queue, str):
+        # keep original behavior for IIB_USER_TO_QUEUE of type dict[str,str])
+        return index_queue
+
+    if isinstance(index_queue, dict):
+        queue_default = index_queue.get('all')
+
+        queue = index_queue.get(from_index_pull_spec)
+        if not queue:
+            logging.debug(
+                'Queue for pull spec %s is not defined. Using default queue (all): %s',
+                from_index_pull_spec,
+                queue_default,
+            )
+            return queue_default
+
+        return queue
+
+    logging.warning('Unsupported type of IIB_USER_TO_QUEUE: %s', index_queue)
+    return None
 
 
 @api_v1.route('/builds/add', methods=['POST'])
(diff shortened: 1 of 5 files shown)


iib/web/api_v1.py
```diff
@@ -513,7 +538,9 @@ def add_bundles() -> Tuple[flask.Response, int]:
     messaging.send_message_for_state_change(request, new_batch_msg=True)
 
     overwrite_from_index = payload.get('overwrite_from_index', False)
-    celery_queue = _get_user_queue(serial=overwrite_from_index)
+    celery_queue = _get_user_queue(
+        serial=overwrite_from_index, from_index_pull_spec=request.from_index.pull_specification
```
xDaile, line 542: What would happen if the request.from_index.pull_specification would not be specified?

lipoja: Nice catch, please review.

iib/web/config.py
```diff
@@ -27,7 +27,7 @@ class Config(object):
     IIB_REQUEST_LOGS_DIR: Optional[str] = None
     IIB_REQUEST_RELATED_BUNDLES_DIR: Optional[str] = None
     IIB_REQUEST_RECURSIVE_RELATED_BUNDLES_DIR: Optional[str] = None
-    IIB_USER_TO_QUEUE: Dict[str, str] = {}
+    IIB_USER_TO_QUEUE: Dict[str, Dict[str, str]] = {}
```
chandwanitulsi, line 30: can we please update the format in README as well, and have a test for queue selection?

lipoja: I am still thinking about keeping backward compatibility. For example by keeping this variable and creating new one with condition only one of them can be set.

lipoja: I changed the code so it accepts both and should not break anything. 
Unit tests for this change are on a way.

iib/web/api_v1.py
```diff
@@ -481,10 +485,43 @@ def _get_user_queue(serial: Optional[bool] = False) -> Optional[str]:
...
+        queue_default = index_queue.get('all')
+        if not queue_default:
+            if not from_index_pull_spec:
+                # we do not have defined default queue and there is no pull spect
+                # for which we should obtain queue. Returning None as celery default
+                return None
+
+            error_msg = f"default queue (all) is not defined for user {username}"
+            logging.error(error_msg)
+            raise IIBError(error_msg)
```
lipoja, line 507: Should we fail or would it be better to return `None` as celery default?

iib/web/api_v1.py
```diff
@@ -481,10 +485,43 @@ def _get_user_queue(serial: Optional[bool] = False) -> Optional[str]:
...
-    return queue
+    index_queue = flask.current_app.config['IIB_USER_TO_QUEUE'].get(labeled_username)
+    if not index_queue:
+        index_queue = flask.current_app.config['IIB_USER_TO_QUEUE'].get(username)
+
+    if index_queue is None:
+        return index_queue
+    elif isinstance(index_queue, str):
+        # keep original behavior for IIB_USER_TO_QUEUE of type dict[str,str])
+        return index_queue
```
chandwanitulsi, line 496: Minor: maybe we can club this if and elif

iib/web/api_v1.py
chandwanitulsi: failing sounds okay to me

chandwanitulsi (commented):
Minor comments, other than that change looks okay, tests are still pending

iib/web/api_v1.py
yashvardhannanavati: Can I suggest not checking the structure of the dictionary here. At this point, we should assume the config is set correctly. We should probably add validation here: https://github.com/release-engineering/iib/blob/master/iib/web/app.py#L45 

There we can check if the structure is correct and if the default `all` is present.

lipoja: I will add the validation, that is good point. However I need to know how to load the config. So if we are using one variable supporting two formats, then I need to do this check on this place. (or call a function which determines the structure)

xDaile (dismissed):

iib/web/api_v1.py
yashvardhannanavati: We can add validation for both formats in app.py, no? Or we simply make this a breaking change?

lipoja: Yes - I have changes made but forget to push it. I will have it ready for review tomorrow.

xDaile (dismissed):

iib/web/api_v1.py
lipoja: The structure changed, please review

iib/web/api_v1.py
```diff
@@ -481,10 +485,41 @@ def _get_user_queue(serial: Optional[bool] = False) -> Optional[str]:
...
+    if index_queue is None:
+        return index_queue
+
+    if isinstance(index_queue, str):
+        # keep original behavior for IIB_USER_TO_QUEUE of type dict[str,str])
+        return index_queue
+
+    if isinstance(index_queue, dict):
+        queue_default = index_queue.get('all')
+        if not queue_default:
```
yashvardhannanavati, line 501: Since you're already validating the structure in app.py, I think we can omit this part here. We can be sure the queue default will be present. Am I missing something?

lipoja: I agree, thanks.

chandwanitulsi (approved):

PR #513, opened 2023-04-27
Adding filtering builds based on `from_index`

lipoja:
CLOUDDST-18396

### DIFF
--- a/iib/web/api_v1.py
+++ b/iib/web/api_v1.py
@@ -366,7 +366,7 @@ def get_builds() -> flask.Response:
     request_type = flask.request.args.get('request_type')
     user = flask.request.args.get('user')
     index_image = flask.request.args.get('index_image')
-
+    from_index = flask.request.args.get('from_index')
     query_params = {}
 
     # Create an alias class to load the polymorphic classes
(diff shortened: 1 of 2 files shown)


iib/web/api_v1.py
```diff
@@ -396,6 +396,40 @@ def get_builds() -> flask.Response:
...
+        query = (
+            query.outerjoin(
+                request_create_empty_index_alias,
+                Request.id == request_create_empty_index_alias.id,
+            )
+            .outerjoin(request_add_alias, Request.id == request_add_alias.id)
+            .outerjoin(request_rm_alias, Request.id == request_rm_alias.id)
+            .outerjoin(request_fbc_operations_alias, Request.id == request_fbc_operations_alias.id)
+        )
+        flask.current_app.logger.error(query)
```
chandwanitulsi, line 422: Was it for intentional or for debugging? If intentional, how about adding it after query.filter()?

iib/web/api_v1.py
```diff
@@ -396,6 +396,40 @@ def get_builds() -> flask.Response:
...
+        request_add_alias = aliased(RequestAdd, flat=True)
+        request_rm_alias = aliased(RequestRm, flat=True)
+        request_fbc_operations_alias = aliased(RequestFbcOperations, flat=True)
+
+        query_params['from_index'] = from_index
+        # Get the image id of the image to be searched
+        from_index_result = Image.query.filter_by(pull_specification=from_index).first()
+        if not from_index_result:
+            # if index_image is not found in image table, then raise an error
+            raise ValidationError(f'{from_index} is not a valid index image')
```
chandwanitulsi, line 410: Minor: Although "not a valid index image" is technically correct, how about using "not a valid from_index"?

chandwanitulsi (commented):
Minor Comments, Other than that looks good

iib/web/api_v1.py
```diff
@@ -396,6 +396,40 @@ def get_builds() -> flask.Response:
         query_params['user'] = user
         query = query.join(Request.user).filter(User.username == user)
 
+    if from_index:
```
yashvardhannanavati, line 399: I think a lot of code is similar to that of index_image. Can I please ask if it would be possible to create a helper function instead? Or de-duplicate some of the code like the aliases?

iib/web/api_v1.py
lipoja: removed

iib/web/api_v1.py
lipoja: Changed

iib/web/api_v1.py
```diff
@@ -397,35 +397,57 @@ def get_builds() -> flask.Response:
...
+                # if from_index is not found in image table, then raise an error
+                raise ValidationError(f'from_index {from_index} is not a valid index image')
+
+            query = query.filter(
+                or_(
+                    request_create_empty_index_alias.from_index_id == from_index_result.id,
+                    request_add_alias.from_index_id == from_index_result.id,
+                    request_rm_alias.from_index_id == from_index_result.id,
+                    request_fbc_operations_alias.from_index_id == from_index_result.id,
                 )
```
xDaile, line 432: This can be written with in_ operator., but I am unsure if it worsens or improves readability.
`query = query.filter(
                    from_index_result.id.in_(
                        [
                        request_create_empty_index_alias.from_index_id,
                        request_add_alias.from_index_id,
                        request_rm_alias.from_index_id,
                        request_fbc_operations_alias.from_index_id,
                       ]
                  )
)`

* Parameter of the in_ operator should be probably list.

xDaile (commented):
Small comment, otherwise LGTM.

yashvardhannanavati:
LGTM! Waiting for the above comments to get resolved

docker/Dockerfile-workers
```diff
@@ -39,6 +39,11 @@ COPY docker/libpod.conf /usr/share/containers/libpod.conf
 
 COPY . .
 
+RUN mkdir -p /root/.docker
+RUN touch /root/.docker/config.json.template
+ADD docker/docker-config-copy.json /root/.docker/config.json.template
+RUN chmod 777 /root/.docker/config.json.template
```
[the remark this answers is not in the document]
lipoja: removed

iib/web/api_v1.py
lipoja: This is good idea however it is not usable for our use-case. We need to find integer in other variables. But in_ should be used to compare variable with other data. I tried your suggestion but got this error: ` AttributeError: 'int' object has no attribute 'in_'`. Therefore I would suggest to keep current query for now.

xDaile: Yes, I forgot it is intended to use with sqlalchemy object parameter.

xDaile (approved):

chandwanitulsi (approved):

PR #526, opened 2023-06-21
Adding from_index_startswith builds search filter

lipoja:
This is required for monitoring script to easily search requests for index image without specifying tag

### DIFF
--- a/iib/web/api_v1.py
+++ b/iib/web/api_v1.py
@@ -367,6 +367,7 @@ def get_builds() -> flask.Response:
     user = flask.request.args.get('user')
     index_image = flask.request.args.get('index_image')
     from_index = flask.request.args.get('from_index')
+    from_index_startswith = flask.request.args.get('from_index_startswith')
     query_params = {}
 
     # Create an alias class to load the polymorphic classes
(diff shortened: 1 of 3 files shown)


zanssa:
@lipoja Shouldn't we add test functions for testing the filter?

iib/web/api_v1.py
```diff
@@ -432,6 +433,30 @@ def get_builds() -> flask.Response:
                 )
             )
 
+        if from_index_startswith:
+            query_params['from_index_startswith'] = from_index_startswith
+            from_index_results = Image.query.filter(
```
zanssa, line 438: should this be `from_index_results` or `from_index_startswith_results `?

lipoja: yeah it would be better :)

lipoja: please review changes

xDaile (approved):

chandwanitulsi (approved):

zanssa (approved):

iib/web/api_v1.py
zanssa: Thank you!

PR #527, opened 2023-06-21
Convert IIB_OTEL_TRACING to an env variable

yashvardhannanavati:


lipoja (changes requested):

iib/common/tracing.py
```diff
@@ -87,7 +86,7 @@ def instrument_tracing(
     def decorator_instrument_tracing(func):
         @functools.wraps(func)
         def wrapper(*args, **kwargs):
-            if not Config.IIB_OTEL_TRACING:
+            if not bool(os.getenv('IIB_OTEL_TRACING', False)):
```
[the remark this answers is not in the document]
yashvardhannanavati: you're absolutely right! I should have thought of it. Fixed it

iib/web/app.py
```diff
@@ -241,7 +239,7 @@ def create_app(config_obj: Optional[str] = None) -> Flask:  # pragma: no cover
     app.register_error_handler(kombu.exceptions.KombuError, json_error)
 
     # Add Auto-instrumentation
-    if app.config['IIB_OTEL_TRACING']:
+    if bool(os.getenv('IIB_OTEL_TRACING', False)):
```
lipoja, line 242: Psst! The conversion is also here...

yashvardhannanavati: lol, nice eye! Fixed it :)

tests/test_workers/test_config.py
```diff
@@ -291,35 +292,23 @@ def test_validate_celery_config_invalid_s3_env_vars():
...
-    (
-        (
-            {'iib_otel_tracing': True},
-            (
-                '"OTEL_EXPORTER_OTLP_ENDPOINT" and "OTEL_SERVICE_NAME" environment '
-                'variables must be set to valid strings when iib_otel_tracing is set to True.'
-            ),
-        ),
-        (
-            {'iib_otel_tracing': 'random-str'},
```
lipoja, line 305: I know that nobody likes to write tests.
However it would be nice also cover the case when `IIB_OTEL_TRACING` is set incorrectly or not set at all (as it was in this test before).

`@mock.patch.dict(os.environ, {'IIB_OTEL_TRACING': 'random-str'}) `

yashvardhannanavati: Aha, that case is indeed absence of failure. I removed it because if you specify anything other than 'True', it should evaluate as false and not check otel variables. I'll add it to a test that validates successfully. Would that suffice?

yashvardhannanavati: Added. Please let me know what you think

lipoja (approved):

chandwanitulsi (approved):
LGTM

PR #529, opened 2023-06-27
Retry buildah if incorrect arch created

chandwanitulsi:
buildah bud builds an index image with incorrect arch sometimes, in order to work around that inspect the image built and rebuild the image if the arch is incorrect.
Refers to CLOUDDST-18481

### DIFF
--- a/iib/workers/tasks/build.py
+++ b/iib/workers/tasks/build.py
@@ -85,7 +85,8 @@ def _build_image(dockerfile_dir: str, dockerfile_name: str, request_id: int, arc
     :param str arch: the architecture to build this image for
     :raises IIBError: if the build fails
     """
-    destination: str = _get_local_pull_spec(request_id, arch)
+    local_destination: str = _get_local_pull_spec(request_id, arch, include_transport=True)
+    destination: str = local_destination.split('/')[1]
     log.info(
         'Building the container image with the %s dockerfile for arch %s and tagging it as %s',
         dockerfile_name,
@@ -117,6 +118,16 @@ def _build_image(dockerfile_dir: str, dockerfile_name: str, request_id: int, arc
         {'cwd': dockerfile_dir},
         exc_msg=f'Failed to build the container image on the arch {arch}',
     )
+    log.debug('Verifying that %s was built with expected arch %s', destination, arch)
+    archmap = get_worker_config()['iib_supported_archs']
+    destination_arch = get_image_label(local_destination, 'architecture')
+
+    if destination_arch not in archmap.values() or destination_arch != archmap.get(arch):
+        log.warning("Wrong arch created for %s", destination)
+        raise ExternalServiceError(
+            f'Wrong arch created, for image {destination} '
+            f'expected arch {archmap.get(arch)}, found {destination_arch}'
+        )
 
 
 def _cleanup() -> None:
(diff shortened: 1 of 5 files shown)


chandwanitulsi:

iib/workers/tasks/build.py
```diff
@@ -117,6 +118,16 @@ def _build_image(dockerfile_dir: str, dockerfile_name: str, request_id: int, arc
         {'cwd': dockerfile_dir},
         exc_msg=f'Failed to build the container image on the arch {arch}',
     )
+    log.debug('Verifying that %s was built with expected arch %s', destination, arch)
+    buildah_inspect = json.loads(run_cmd(['buildah', 'inspect', destination]))
+    archmap = {"amd64": "x86_64", "arm64": "aarch64", "s390x": "s390x", "ppc64le": "ppc64le"}
+    destination_arch = buildah_inspect['Docker']['config']['Labels']['architecture']
+
+    if destination_arch not in archmap.values() or destination_arch != archmap.get(arch, None):
```
lipoja, line 126: Nitpick: `.get(arch, None)` - `None` is redundant

iib/workers/tasks/build.py
```diff
@@ -117,6 +118,16 @@ def _build_image(dockerfile_dir: str, dockerfile_name: str, request_id: int, arc
         {'cwd': dockerfile_dir},
         exc_msg=f'Failed to build the container image on the arch {arch}',
     )
+    log.debug('Verifying that %s was built with expected arch %s', destination, arch)
+    buildah_inspect = json.loads(run_cmd(['buildah', 'inspect', destination]))
```
yashvardhannanavati, line 122: Could we please reuse the `get_image_label` function here: https://github.com/release-engineering/iib/blob/master/iib/workers/tasks/utils.py#L1091

iib/workers/tasks/build.py
```diff
@@ -117,6 +118,16 @@ def _build_image(dockerfile_dir: str, dockerfile_name: str, request_id: int, arc
...
+    archmap = {"amd64": "x86_64", "arm64": "aarch64", "s390x": "s390x", "ppc64le": "ppc64le"}
```
yashvardhannanavati, line 123: I wonder if it makes sense to move this to worker config instead. That way, it's easier to maintain and we don't forget about it if an arch specific change is made in the future.

iib/workers/tasks/build.py
```diff
@@ -117,6 +118,16 @@ def _build_image(dockerfile_dir: str, dockerfile_name: str, request_id: int, arc
...
+    destination_arch = buildah_inspect['Docker']['config']['Labels']['architecture']
+
+    if destination_arch not in archmap.values() or destination_arch != archmap.get(arch):
+        log.warning("Wrong arch created for %s", destination)
+        exp_message = f'Wrong arch created, for image {destination} expected arch {arch}, \
+            found {destination_arch}'
+        raise IIBError(exp_message)
```
yashvardhannanavati, line 130: Since the wrong arch is indeed being built by buildah, I wonder if it makes sense to raise `ExternalServiceError ` instead of `IIBError`.

yashvardhannanavati (changes requested):

iib/workers/tasks/build.py
chandwanitulsi: skopeo inspect does not work as the image is not pushed yet

iib/workers/tasks/build.py
chandwanitulsi: sounds fair

iib/workers/tasks/build.py
chandwanitulsi: hmm, I was hoping to use IIBerror for the exception message, I can try that with ExternalServiceError

iib/workers/tasks/build.py
chandwanitulsi: I found a way to use, will update in next patch

yashvardhannanavati: thank you!

yashvardhannanavati (approved):

PR #532, opened 2023-07-13
Update Sphinx documentation

xDaile:
The celery was not compatible with Sphinx >=6.2.0 because of this issue (https://github.com/celery/celery/pull/7978). As the new version of celery, compatible with Sphinx > 6.2.0 was released, we need to update the outdated documentation. 

Currently, the Sphinx library has available a newer version 7.0.1 instead of 6.2.1, however, this version is not compatible with the sphinx-rtd-theme as mentioned in  (https://github.com/readthedocs/sphinx_rtd_theme/issues/1463), so we need to stick to 6.2.1 for now.

- Documentation to new modules was added.
- docs/requeirements.txt was reordered and updated with opentelementry packages
- Added workflow job for building the docs in Github CI/CD

Additionally /web/static/api_v1.yaml was updated.


iib/common/tracing.py
```diff
@@ -7,8 +7,8 @@
 
 Usage:
     @instrument_tracing()
-    def func():
-        pass
+      def func():
+          pass
```
xDaile, line 11: This was the only way I found to write this comment, where Sphinx was not throwing errors.

docs/requirements.txt
```diff
@@ -13,7 +13,18 @@ recommonmark
 requests
 ruamel.yaml
 # https://github.com/sphinx-doc/sphinx/issues/2796
-sphinx==6.2.0
+sphinx==6.2.1
 sphinx_rtd_theme
 tenacity
 typing-extensions
+opentelemetry-api
```
yashvardhannanavati, line 20: Could you please arrange these dependencies alphabetically

iib/common/tracing.py
yashvardhannanavati: Could you please share the error you were getting without this extra tab?

xDaile: `/iib/common/tracing.py:docstring of iib.common.tracing:9:Unexpected indentation.`

yashvardhannanavati (dismissed):

docs/requirements.txt
```diff
@@ -1,19 +1,30 @@
...
 dogpile.cache
 flask
 flask-login
 flask-migrate
 flask-sqlalchemy
+opentelemetry-api
+opentelemetry-exporter-otlp
+opentelemetry-instrumentation
+opentelemetry-instrumentation-botocore
+opentelemetry-instrumentation-celery
```
chandwanitulsi, line 13: why do we have to add opentelemetry libraries for this PR?

docs/requirements.txt
```diff
@@ -1,19 +1,30 @@
-celery
+# https://github.com/sphinx-doc/sphinx/issues/2796
```
chandwanitulsi, line 1: Is this comment only valid for line 27 `sphinx==6.2.1` ?

docs/requirements.txt
yashvardhannanavati: I believe this is only for docs. They are already listed as dependencies in setup.py

xDaile: Exactly. As I added modules, which included parts of the code working with opentelemetry, into documentation, Sphinx was throwing errors about missing opentelemetry libraries.

docs/requirements.txt
xDaile: Sure! Thanks for catch.

docs/requirements.txt
```diff
@@ -1,19 +1,29 @@
...
-python-memcached
 psycopg2-binary
+python-memcached
+python-qpid-proton
 recommonmark
 requests
 ruamel.yaml
-# https://github.com/sphinx-doc/sphinx/issues/2796
-sphinx==6.2.0
+sphinx==6.2.1
```
lipoja, line 26: May I ask you a favor?
Could you test latest sphinx release - v7.0.1 I would go with pinning latest rather then older one. 
But only if everything passes.

lipoja: _# And also dependabot is suggesting to pin latest version in different PR._

lipoja: _# And also dependabot is suggesting to pin latest version in different PR._

xDaile: I already wrote the reason why this is not possible in the PR description ([sphinx-rtd-theme issue](https://github.com/readthedocs/sphinx_rtd_theme/issues/1463)).

lipoja: I should do the review better next time - starting reading description :D ... thanks!

lipoja (approved):

lipoja (dismissed):

lipoja (approved):

yashvardhannanavati (approved):

chandwanitulsi (approved):

PR #552, opened 2023-08-01
Remove Dependabot and Reconfigure Renovate bot

yashvardhannanavati:
Refers to CLOUDDST-19743

Custom config for IIB. Reference: https://docs.renovatebot.com/configuration-options/


xDaile (dismissed):

chandwanitulsi (approved):

zanssa (approved):

.github/renovate.json
```diff
@@ -0,0 +1,10 @@
+{
+  "enabled": true,
+  "enabledManagers": ["docker-compose", "dockerfile", "github-actions", "pip_requirements", "pyenv"],
```
lipoja, line 3: Are we sure that default `dockerfile` manager is matching our dockerfiles? 
https://docs.renovatebot.com/modules/manager/dockerfile/#default-config

From that regexp I think it is not. Maybe we should add `fileMatch` for `dockerfile` as well?

lipoja (changes requested):

.github/renovate.json
```diff
@@ -0,0 +1,10 @@
...
+  "branchConcurrentLimit": 5,
+  "docker-compose": {
+    "fileMatch": ["podman-compose.yml$"]
```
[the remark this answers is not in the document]
yashvardhannanavati: So, `fileMatch` appends to the existing patterns they have by default. If you check the comment here: https://github.com/release-engineering/iib/pull/535, everything we want to track is being tracked by default except for `podman-compose.yml`. Hence, I only added the missing file here.

.github/renovate.json
yashvardhannanavati: Yes, it is matching the dockerfiles : https://github.com/release-engineering/iib/pull/535

zanssa (changes requested):
Inspired by a discussion about having specific commits as it helps when looking at the commit history in the future, shouldn't this change be splitted into multiple commits as I don't see it reflecting the removal of the old method to handle the dependencies update?

yashvardhannanavati:
> Inspired by a discussion about having specific commits as it helps when looking at the commit history in the future, shouldn't this change be splitted into multiple commits as I don't see it reflecting the removal of the old method to handle the dependencies update?

Yes, you're absolutely right. I split it into two commits.

zanssa:
Giving this more thought, I am not sure how pyenv manager would work. Maybe you can help me here @yashvardhannanavati 

Looking at this MR #355, all the work needed in CLOUDDST-10861, and the tasks that were triggered by that change, do we need to keep pyenv enabled in the configuration? 

Maybe we should enable this in stages and based on that enable it for others managers. thoughts?

yashvardhannanavati:
> Giving this more thought, I am not sure how pyenv manager would work. Maybe you can help me here @yashvardhannanavati 
> 
> 
> 
> Looking at this MR #355, all the work needed in CLOUDDST-10861, and the tasks that were triggered by that change, do we need to keep pyenv enabled in the configuration? 
> 

> 
> 
> Maybe we should enable this in stages and based on that enable it for others managers. thoughts? 

So as of today, all the managers are enabled since we have the default config. The reason I didn't disable it was thinking that we can always close the PR if it ever creates one since automerge is not enabled. But I do agree that we updated the pyenv file manually when we dropped support for 3.6 and upgraded to 3.8. So I am open to dropping pyenv from that list and that will disable it. Thoughts?

xDaile (approved):

lipoja (approved):

yashvardhannanavati:
@zanssa We discussed this in the guild meeting and thought it might be better to keep the `pyenv` manager along with the others. Could you please share if you have any concerns? :)

chandwanitulsi (approved):

PR #556, opened 2023-08-07
Add related_images check in IIB

chandwanitulsi:


chandwanitulsi:

iib/web/models.py
```diff
@@ -1118,6 +1130,7 @@ def from_json(  # type: ignore[override] # noqa: F821
                 'deprecation_list',
                 'graph_update_mode',
                 'build_tags',
+                'check_related_images',
```
yashvardhannanavati, line 1133: Can I please ask why we are not adding `check_related_images` attribute to the database? Also, it's not present in `to_json` which means we won't be able to see it in the API response meaning we would probably have to check the logs to determine if the check was run.

iib/workers/tasks/build.py
```diff
@@ -778,6 +779,44 @@ def _verify_index_image(
...
+def inspect_related_images(bundles: List[str], request_id) -> None:
+    """
+    Verify if related_images in the bundle are pullable.
+
+    :param list bundles: a list of strings representing the pull specifications of the bundles to
+        add to the index image being built.
+    :param int request_id: the ID of the request this index image is for.
+    :raises IIBError: if one of the bundles does not have the pullable related_image.
+    """
+    from iib.workers.tasks.build_regenerate_bundle import _get_bundle_metadata
```
yashvardhannanavati, line 791: It looks like we are importing a private function here. We do have it in other places too so I won't block the merging of the PR on this but if it's not too much work, could you please move that function somewhere else?  If it is, please feel free to file a separate issue for it and resolve the comment

yashvardhannanavati (changes requested):

iib/web/models.py
```diff
@@ -1073,6 +1074,17 @@ def from_json(  # type: ignore[override] # noqa: F821
...
             raise ValidationError('"from_index" must be specified if no bundles are specified')
 
+        # Verify that `check_related_images` is specified when bundles are specified
+        if request_kwargs.get('check_related_images') and not request_kwargs.get('bundles'):
+            raise ValidationError(
+                '"check_related_images" must be specified only when bundles are specified'
+            )
+
+        # Verify that `check_related_images` is the correct type
+        check_related_images = request_kwargs.pop('check_related_images', False)
```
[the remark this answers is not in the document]
chandwanitulsi: yeah, I should not send it to cls._from_json() at the end. This parameter should be restricted to RequestAdd, will update in the next patch.
Thanks

iib/web/models.py
chandwanitulsi: I did not add it to the database earlier as I think this is similar to a validation check, the output index image will not be impacted. 
I agree that the parameter should be seen in response as well, will update in next patch

iib/workers/tasks/build.py
chandwanitulsi: I will add a separate PR/commit for fthat

iib/web/models.py
chandwanitulsi: Updated. Please resolve the comment if the changes are satisfactory

iib/web/models.py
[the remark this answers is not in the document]
chandwanitulsi: Updated. Please resolve the comment if the changes are satisfactory

iib/workers/tasks/build.py
chandwanitulsi: created https://github.com/release-engineering/iib/pull/560

iib/web/models.py
yashvardhannanavati: Since you're validating the attribute in the ReqestAdd `from_json` itself and this parameter is not going to be used by multiple request types, shouldn't we not pass `check_related_images ` while calling the `from_json` of the parent class?

yashvardhannanavati (changes requested):
Just a couple of minor comments. Otherwise, LGTM!

iib/web/models.py
chandwanitulsi: I tried not passing to from_json of the parent class, it errors with "The following parameters are invalid: check_related_images"

iib/web/models.py
```diff
@@ -1042,6 +1042,7 @@ class RequestAdd(Request, RequestIndexImageMixin):
     organization: Mapped[Optional[str]]
 
     omps_operator_version: Mapped[Optional[str]]
+    check_related_images: Mapped[Optional[bool]]
```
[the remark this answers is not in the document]
chandwanitulsi: updated

iib/web/models.py
yashvardhannanavati: Ah, I see. It's the generalized check for valid/invalid parameters. Thanks!

yashvardhannanavati (approved):

JAVGan (approved):

zanssa (approved):

iib/web/models.py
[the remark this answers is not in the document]
zanssa: @lipoja ping

lipoja (approved):

PR #558, opened 2023-08-08
Enable celery task deduplication

lipoja:
Before each task execution, instruct the worker to check if this task is a duplicate message.

Deduplication occurs only with tasks that have the same identifier, enabled late acknowledgment, were redelivered by the message broker and their state is SUCCESS in the result backend.

[CLOUDDST-19499]

### DIFF
--- a/iib/workers/config.py
+++ b/iib/workers/config.py
@@ -95,6 +95,11 @@ class Config(object):
     temp_fbc_fragment_path = 'fbc-fragment'
     # For now, only allow a single process so that all tasks are processed serially
     worker_concurrency: int = 1
+    # Before each task execution, instruct the worker to check if this task is a duplicate message.
+    # Deduplication occurs only with tasks that have the same identifier,
+    # enabled late acknowledgment, were redelivered by the message broker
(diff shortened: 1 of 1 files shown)


iib/workers/config.py
```diff
@@ -95,6 +95,11 @@ class Config(object):
     temp_fbc_fragment_path = 'fbc-fragment'
     # For now, only allow a single process so that all tasks are processed serially
     worker_concurrency: int = 1
+    # Before each task execution, instruct the worker to check if this task is a duplicate message.
+    # Deduplication occurs only with tasks that have the same identifier,
+    # enabled late acknowledgment, were redelivered by the message broker
+    # and their state is SUCCESS in the result backend.
+    worker_deduplicate_successful_tasks = True
```
yashvardhannanavati, line 102: Could you please add the missing type?

 `worker_deduplicate_successful_tasks : bool = True`

yashvardhannanavati (changes requested):
Could you please add the missing type?

`worker_deduplicate_successful_tasks: bool = True`

PR #561, opened 2023-08-16
Log a warning message when "architecture" label is missing under "Labels"

yashvardhannanavati:
When using an Openshift binary as the binary_image, the "architecture" label is added under the "Labels" section on an index image. This is done for the OCP cluster to recognize the arch of the image. This is an OCP specific thing and it's not necessary that the "architecture" label will be added under "Labels" if a user specifies any other binary_image. Hence, in-case the label is missing from the index, do not fail the request.

Refers to CLOUDDST-19860

### DIFF
--- a/iib/workers/tasks/build.py
+++ b/iib/workers/tasks/build.py
@@ -119,8 +119,15 @@ def _build_image(dockerfile_dir: str, dockerfile_name: str, request_id: int, arc
         exc_msg=f'Failed to build the container image on the arch {arch}',
     )
     log.debug('Verifying that %s was built with expected arch %s', destination, arch)
-    archmap = get_worker_config()['iib_supported_archs']
+    archmap = worker_config['iib_supported_archs']
     destination_arch = get_image_label(local_destination, 'architecture')
+
+    if not destination_arch:
+        log.warn(
+            'The "architecture" label was not found under "Labels".'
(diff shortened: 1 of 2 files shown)


iib/workers/tasks/build.py
```diff
@@ -119,9 +119,16 @@ def _build_image(dockerfile_dir: str, dockerfile_name: str, request_id: int, arc
...
     )
     log.debug('Verifying that %s was built with expected arch %s', destination, arch)
-    archmap = get_worker_config()['iib_supported_archs']
+    archmap = worker_config['iib_supported_archs']
     destination_arch = get_image_label(local_destination, 'architecture')
 
+    if not destination_arch:
+        log.warn(
+            '"architecture" label not found under "Labels".'
+            'Skipping the check that confirms if the architecture label was set correctly.'
```
xDaile, line 128: Can the warning be rewritten, so the sentence would start with a capital letter? I find logs more readable that way (looking for capital letters, when looking for another log message).

chandwanitulsi (approved):

iib/workers/tasks/build.py
yashvardhannanavati: Done. :)

xDaile (approved):

PR #566, opened 2023-09-18
Flush request to database 

lipoja:
Flush request to database to prevent UniqueViolation due to parallel requests processing.

[CLOUDDST-20061]

### DIFF
--- a/iib/web/models.py
+++ b/iib/web/models.py
@@ -257,6 +257,13 @@ class Image(db.Model):
         if not image:
             image = Image(pull_specification=pull_specification)
             db.session.add(image)
+            try:
+                db.session.commit()
+            except sqlalchemy.exc.IntegrityError:
+                current_app.logger.info(
+                    'Image pull specification is already in database. "%s"', pull_specification
+                )
+            image = cls.query.filter_by(pull_specification=pull_specification).first()
 
         return image
 
@@ -284,6 +291,11 @@ class Operator(db.Model):
         if not operator:
             operator = Operator(name=name)
             db.session.add(operator)
+            try:
+                db.session.commit()
+            except sqlalchemy.exc.IntegrityError:
+                current_app.logger.info('Operators is already in database. "%s"', name)
+            operator = cls.query.filter_by(name=name).first()
 
         return operator
 
@@ -1696,6 +1708,11 @@ class User(db.Model, UserMixin):
         if not user:
             user = User(username=username)
             db.session.add(user)
+            try:
+                db.session.commit()
+            except sqlalchemy.exc.IntegrityError:
+                current_app.logger.info('User is already in database. "%s"', username)
+            user = cls.query.filter_by(username=username).first()
 
         return user
 
--- a/tests/test_web/test_api_v1.py
+++ b/tests/test_web/test_api_v1.py
@@ -1234,7 +1234,8 @@ def test_patch_request_add_success(
     }
 
     for bundle in bundles:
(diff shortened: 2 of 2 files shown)


iib/web/models.py
```diff
@@ -257,6 +257,7 @@ def get_or_create(cls, pull_specification: str) -> Image:
         if not image:
             image = Image(pull_specification=pull_specification)
             db.session.add(image)
+            db.session.flush()
```
lipoja, line 260: Question for discussion might be if it is OK to flush db.session at this place?
Does anybody see potential failure doing it?

xDaile (approved):

iib/web/models.py
xDaile: Why we are not doing that also after adding a User?

yashvardhannanavati:
> Flush request to database to prevent UniqueViolation due to parallel requests processing.
I wonder how parallel requests processing is causing the UniqueViolation. Although the pullspec is same, the primary key i.e. ID for the two parallel requests would be different, no?

lipoja:
> > Flush request to database to prevent UniqueViolation due to parallel requests processing.
> 
> I wonder how parallel requests processing is causing the UniqueViolation. Although the pullspec is same, the primary key i.e. ID for the two parallel requests would be different, no?

Bu there is unique on pullspec. 
see  https://github.com/release-engineering/iib/blob/2b5052c65faf0dc0ad6506fede7b6f5101549efe/iib/web/models.py#L233

Imaging running two requests on two workers almost at the same time. 
First worker added this pullspec to db.session.
Second worker checked if pullspec is in DB., it was not there since it was not flushed yet.
First db.sessoin.flush()
Second worker hit the flush few moments later when there was the insert of pullspec again and it failed.

This is my hypotheses, I will sum it up in RCA tomorrow.

The question is if by adding this flush we do not flush some items to DB too early.

iib/web/models.py
lipoja: even though users are added hardly ever you are right. I added flush there as well.

JAVGan: > Question for discussion might be if it is OK to flush db.session at this place?
> Does anybody see potential failure doing it?

I don't see a problem with this. Looking at the [docs](https://docs.sqlalchemy.org/en/20/orm/session_basics.html#flushing) it would be eventually flushed anyway whenever a query would be executed, so doing this beforehand to avoid the aforementioned unique violation seems ok. It should also be an atomic operation as we're dealing with a postgreSQL in the backend so it should lock the operation before allowing a secondary "parallel flush" to happen.

It may impact a bit in performance, though, but I think it's something worthy in the tradeoff.

JAVGan (dismissed):

iib/web/models.py
chandwanitulsi: It looked like adding db.session.flush() could fix the image uniqueviolation error first, but when digging deeper into this, I noticed we already create the session object with autoflush=True, shouldn't that take care of flushing the image object already?

chandwanitulsi (approved):

iib/web/models.py
lipoja: Thanks for the comment and I guess you are right in some way. From what I read in docu flush() will not help us here. Hence I change it to commit(). Flush will move the SQL queries to transaction buffer in DB. And it will stay there until commit is called.

xDaile (approved):

xDaile (approved):

yashvardhannanavati (approved):

PR #578, opened 2023-10-17
Adding graph update mode for merge index image API endpoint

lipoja:
[CLOUDDST-20163]


iib/web/iib_static_types.py
```diff
@@ -388,6 +389,7 @@ class MergeIndexImageRequestResponse(APIPartImageBuildRequestResponse):
     build_tags: List[str]
     deprecation_list: List[str]
     distribution_scope: str
+    graph_update_mode: str
```
xDaile, line 392: Type can be Literal["replaces", "semver", "semver-skippatch"]

iib/workers/tasks/build_merge_index_image.py
```diff
@@ -179,6 +184,7 @@ def handle_merge_request(
     distribution_scope: Optional[str] = None,
     binary_image_config: Optional[str] = None,
     build_tags: Optional[List[str]] = None,
+    graph_update_mode: Optional[str] = None,
```
xDaile, line 187: I suppose Mypy could have a problem with this line (using Literal and str].

xDaile (dismissed):

iib/web/iib_static_types.py
```diff
@@ -390,7 +390,7 @@ class MergeIndexImageRequestResponse(APIPartImageBuildRequestResponse):
     build_tags: List[str]
     deprecation_list: List[str]
     distribution_scope: str
-    graph_update_mode: Literal['replaces', 'semver', 'semver-skippatch']
+    graph_update_mode: GRAPH_MODE_LITERAL
```
JAVGan, line 393: Is it possible to do this in the previous commit instead of this one?

iib/web/iib_static_types.py
```diff
@@ -162,7 +162,7 @@ class MergeIndexImagesPayload(TypedDict):
     build_tags: NotRequired[List[str]]
     deprecation_list: NotRequired[List[str]]
     distribution_scope: NotRequired[str]
-    graph_update_mode: NotRequired[str]
+    graph_update_mode: NotRequired[GRAPH_MODE_LITERAL]
```
JAVGan, line 165: Is it possible to do this in the previous commit instead of this one?

JAVGan (approved):
Besides the fix for the first commit in the second one everything else LGTM

iib/web/iib_static_types.py
lipoja: The other way around. The `graph_update_mode` should not be present in the first commit.

iib/web/iib_static_types.py
lipoja: The other way around. The graph_update_mode should not be present in the first commit.

JAVGan (approved):

xDaile (approved):

chandwanitulsi (approved):

PR #588, opened 2023-11-10
Allow skopeo inspect to check for images for any arch

chandwanitulsi:
When '--raw' option was not given in inspect, skopeo tries to look for manifest of amd64 arch, but some images are built for other arches only, hence skopeo inspect was failing for those images

### DIFF
--- a/iib/workers/tasks/build.py
+++ b/iib/workers/tasks/build.py
@@ -814,7 +814,7 @@ def inspect_related_images(bundles: List[str], request_id) -> None:
             for related_image in bundle_metadata['found_pullspecs']:
                 related_image_pull_spec = related_image.to_str()
                 try:
-                    skopeo_inspect(f"docker://{related_image_pull_spec}")
+                    skopeo_inspect(f'docker://{related_image_pull_spec}', '--raw')
                 except IIBError as e:
                     log.error(e)
                     invalid_related_images.append(related_image_pull_spec)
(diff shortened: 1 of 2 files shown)


chandwanitulsi:

iib/workers/tasks/build.py
```diff
@@ -814,7 +814,7 @@ def inspect_related_images(bundles: List[str], request_id) -> None:
             for related_image in bundle_metadata['found_pullspecs']:
                 related_image_pull_spec = related_image.to_str()
                 try:
-                    skopeo_inspect(f"docker://{related_image_pull_spec}")
+                    skopeo_inspect(f'docker://{related_image_pull_spec}', '--raw')
```
lipoja, line 817: I am not sure if you have any test data, but does the command fail when there is invalid/missing arch of an image?

chandwanitulsi: yeah, it fails. Basically if I run skopeo inspect, skopeo by default tries to look for manifest for the arch same as the system where the command is run.  (it was amd64 in our case)
Here is an example, `skopeo --command-timeout 300s inspect docker://registry.redhat.io/amq7/amq-broker-openj9-11-rhel8@sha256:d4010af53429502f41d722c39d69a7bb147bb06d19e3d4a69de70dbe88e3990e
FATA[0001] Error parsing manifest for image: choosing image instance: no image found in manifest list for architecture amd64, variant "", OS linux 
`
images for amq7/amq-broker-openj9-11-rhel8 are built only for ppc64le and s390x arches, so when I run inspect in amd64 system, it fails, but when adding "--raw" the command doesn't fail. 
Another option was could use --override-os / --override-arch, but I think that would be an overkill as then we will have to find the arch of image, and the intent here is just to check if the image is available in the registry

lipoja: Thank you!

lipoja (approved):

yashvardhannanavati (approved):

PR #589, opened 2023-11-10
Adding config variable to allow adding bundles without ocp version label

lipoja:
- added iib_no_ocp_label_allow_list to worker config file
- allow bundles without "com.redhat.openshift.versions" label set to be added for index images set in config
- added unit tests for is_bundle_version_valid

[CLOUDDST-20868]


iib/workers/tasks/build_merge_index_image.py
```diff
@@ -415,6 +424,12 @@ def is_bundle_version_valid(bundle_path: str, valid_ocp_version: str) -> bool:
         raise IIBError(f'Invalid OCP version, "{valid_ocp_version}", specified in Index Image')
     try:
         bundle_version_label = get_image_label(bundle_path, 'com.redhat.openshift.versions')
+        if allow_no_ocp_version and not bundle_version_label:
+            log.debug(
+                'Marking bundle %s without label `com.redhat.openshift.versions` as valid.',
+                bundle_path,
+            )
```
JAVGan, line 431: What about using it as `logging.warning` instead? Then we may see in the logs when actually processing these images without need to fetch the current config.

Also if the config changes we may preserve which images were processed without the versions.

lipoja: I am open to change it to info. But not warning. It is not an error state. It is expected and I just wanted to have any information in logfile about processing bundles without the label.

JAVGan: Sure, makes sense, `info` looks good to me

lipoja: changed

JAVGan: Thank you!

JAVGan (approved):

iib/workers/config.py
```diff
@@ -40,6 +40,8 @@ class Config(object):
     iib_index_image_output_registry: Optional[str] = None
     iib_log_level: str = 'INFO'
     iib_max_recursive_related_bundles = 15
+    # list of index images to which we can add bundles without "com.redhat.openshift.versions" label
```
chandwanitulsi, line 43: same, please specify the label can be missing in 'bundle images'

chandwanitulsi (commented):
Minotr comments, mostly looks good

iib/workers/config.py
chandwanitulsi: nvm, it does look okay

chandwanitulsi (approved):

PR #592, opened 2023-11-16
Add guardrails for UniqueViolation caused by parallel requests

yashvardhannanavati:
Sometime, when multiple requests are submitted with the same parameters, it can cause a UniqueViolation in the database. Example, when requests A and B started, a value wasn't present so they both decide to add a value X. A's addition is successful while when B tries to add it, a UniqueViolation is triggered since A already added it.

Refers to CLOUDDST-20863

### DIFF
--- a/iib/web/models.py
+++ b/iib/web/models.py
@@ -253,12 +253,24 @@ class Image(db.Model):
                 f'Image {pull_specification} should have a tag or a digest specified.'
             )
 
-        image = cls.query.filter_by(pull_specification=pull_specification).first()
+        # cls.query triggers an auto-flush of the session by default. So if there are
+        # multiple requests with same parameters submitted to IIB, call to query pre-maturely
+        # flushes the contents of the session not allowing our handlers to resolve conflicts.
+        # https://docs.sqlalchemy.org/en/20/orm/session_api.html#sqlalchemy.orm.Session.params.autoflush
+        with db.session.no_autoflush:
+            image = cls.query.filter_by(pull_specification=pull_specification).first()
+
         if not image:
             image = Image(pull_specification=pull_specification)
-            db.session.add(image)
             try:
-                db.session.commit()
+                # This is a SAVEPOINT so that the rest of the session is not rolled back when
+                # adding the image conflicts with an already existing row added by another request
+                # with similar pullspecs is submitted at the same time. When the context manager
+                # completes, the objects local to it are committed. If an error is raised, it
+                # rolls back objects local to it while keeping the parent session unaffected.
+                # https://docs.sqlalchemy.org/en/20/orm/session_transaction.html#using-savepoint
+                with db.session.begin_nested():
+                    db.session.add(image)
             except sqlalchemy.exc.IntegrityError:
                 current_app.logger.info(
                     'Image pull specification is already in database. "%s"', pull_specification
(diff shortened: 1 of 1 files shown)


iib/web/models.py
```diff
@@ -287,12 +299,23 @@ def get_or_create(cls, name: str) -> Operator:
...
+                # This is a SAVEPOINT so that the rest of the session is not rolled back when
+                # adding the image conflicts with an already existing row added by another request
+                # with similar pullspecs is submitted at the same time. When the context manager
+                # completes, the objects local to it are committed. If an error is raised, it
+                # rolls back objects local to it while keeping the parent session unaffected.
+                # https://docs.sqlalchemy.org/en/20/orm/session_transaction.html#using-savepoint
+                with db.session.begin_nested():
+                    db.session.add(operator)
             except sqlalchemy.exc.IntegrityError:
                 current_app.logger.info('Operators is already in database. "%s"', name)
```
xDaile, line 320: Do we need this catch, if the begin_nested will not fail if the operator is already added?

xDaile (approved):

lipoja (approved):

iib/web/models.py
```diff
@@ -253,12 +253,24 @@ def get_or_create(cls, pull_specification: str) -> Image:
...
             try:
-                db.session.commit()
+                # This is a SAVEPOINT so that the rest of the session is not rolled back when
+                # adding the image conflicts with an already existing row added by another request
+                # with similar pullspecs is submitted at the same time. When the context manager
+                # completes, the objects local to it are committed. If an error is raised, it
+                # rolls back objects local to it while keeping the parent session unaffected.
+                # https://docs.sqlalchemy.org/en/20/orm/session_transaction.html#using-savepoint
+                with db.session.begin_nested():
+                    db.session.add(image)
```
chandwanitulsi, line 273: Can you share why do we need to catch the exception now? db.session.add() will not commit anything to db, so what can trigger sqlalchemy.exc.IntegrityError?

yashvardhannanavati: `db.session.add()` will get committed to the database when the context manager completes on line 274. So there will be a chance that the exception will be raised.

chandwanitulsi: AFAIK, db.session.add just adds the object in the session, but does not commit to the database. I verified this with IIB dev-env as well, when all the db.session.add() ran, IIB db did not have any image/batch/request table updated, but all these tables were updated only after db.session.commit() was run in api_v1.py file

yashvardhannanavati: Under normal circumstances, it does not commit. But this is a Savepoint transaction. 

If I could point you to the link I have put in the code comment: https://docs.sqlalchemy.org/en/20/orm/session_transaction.html#using-savepoint 

Quoting the third paragraph from it.
`When the context manager yielded by [Session.begin_nested()](https://docs.sqlalchemy.org/en/20/orm/session_api.html#sqlalchemy.orm.Session.begin_nested) completes, it “commits” the savepoint, which includes the usual behavior of flushing all pending state. When an error is raised, the savepoint is rolled back and the state of the [Session](https://docs.sqlalchemy.org/en/20/orm/session_api.html#sqlalchemy.orm.Session) local to the objects that were changed is expired.`

So in this specific case, db.session.add() will mark the end of the context manager (` with db.session.begin_nested():`) which means the changes local to the context manager will be committed.

chandwanitulsi: "database changes only when commit() is called explicitly."  exactly, even at line 274 image will not be committed to db. FWIW, I am not against this approach, but sqlalchemy.exc.IntegrityError exception is raised when we commit to db, right? So this exception will never be raised

yashvardhannanavati: It will be raised even on `flush`, I think. It is a documented behavior for PostgreSQL too. 

Check out the third code snippet in the link I posted above for savepoint. It is doing exactly what we are trying to do.

IIUC, when the changes are flushed, they will be in the transaction buffer of the db. Yes, they wouldn't be written to disk yet but if you query the database using this session, it will return the changed values you have flushed.

chandwanitulsi: "It will be raised even on flush, I think. It is a documented behavior for PostgreSQL too." Okay, I checked the documentation, and this sounds true.
and +1 to "when the changes are flushed, they will be in the transaction buffer of the db. Yes, they wouldn't be written to disk yet but if you query the database using this session, it will return the changed values you have flushed." as well.
No more blocking questions from me :). Thanks for the patience.

yashvardhannanavati: Thank you for asking the questions. It helps clear any doubts :D

PR #593, opened 2023-11-22
Adding parameter ignore_bundle_ocp_version to merge-index API

lipoja:
If ignore_bundle_ocp_version parameter is set to true while calling merge-index API and target_index image is listed in iib_no_ocp_label_allow_list in config file then any bundle without "com.redhat.openshift.versions" set will be added to target_index.

[CLOUDDST-20869]

### DIFF
--- a/iib/web/static/api_v1.yaml
+++ b/iib/web/static/api_v1.yaml
@@ -1143,6 +1143,7 @@ components:
               type: string
               example: rm
     RegenerateBundleRequest:
+      $ref: '#/components/schemas/RegenerateBundleResponseVerbose'
       type: object
       properties:
         from_bundle_image:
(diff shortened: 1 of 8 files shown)


iib/web/static/api_v1.yaml
```diff
@@ -1346,6 +1347,14 @@ components:
                 image pull specs configured in IIB_GRAPH_MODE_INDEX_ALLOW_LIST in the IIB API Config. If not
                 specified, "--mode" will not be added to OPM commands to add the bundle(s) to the index.
               default: None
+            ignore_bundle_ocp_version:
+              type: string
```
zanssa, line 1351: shouldn't this be boolean? Also, I think it is only added in the response schema. As a user, if I am gonna submit a merge index request, I would check the properties/request schema rather than the response to see what parameters I can use when submitting a request. Let me know if I am missing something here!

lipoja: Yes, this should be boolean and I missed the request part. Thank you for catching it - fixed.

xDaile (approved):

JAVGan (approved):

zanssa (approved):

PR #597, opened 2023-12-07
Add Pydantic models to IIB

xDaile:


iib/web/models.py
```diff
@@ -390,6 +399,64 @@ class Request(db.Model):
...
+
+        request = self(**request_kwargs)
+
+        if build_tags_allowed:
+            for bt in payload.build_tags:
+                request.add_build_tag(bt)
+
+        request.add_state('in_progress', 'The request was initiated')
+        return request
+
```
xDaile, line 459: All code in from_json() methods could be reduced to this one function. 

Line 451 is giving me a warning that the method cannot call itself. I intended to allow other Request classes to inherit this 'from_json_replacement_even_shorter' method via the Request parent class.

iib/web/models.py
```diff
@@ -2163,3 +2181,61 @@ def get_mutable_keys(self) -> Set[str]:
...
+        request_kwargs['batch'] = batch
+
+    request = Request(**request_kwargs)
+
+    if build_tags_allowed:
+        for bt in payload.build_tags:
+            request.add_build_tag(bt)
+
+    request.add_state('in_progress', 'The request was initiated')
+    return request
```
xDaile, line 2241: All code in from_json() methods could be reduced to this one function.

PR #599, opened 2023-12-19
Check if source_from_index is in iib_no_ocp_label_allow_list

lipoja:
I wrongly assumed that ISV will use overwrite index token and set the check if image is on allow list only for the target (like we do this for bootstraping). However now I know it is not true and this use case slipped my hands. We have to also tests the source index image because the target can be newly created empty index image in our IIB repo. 

[CLOUDDST-20869]

### DIFF
--- a/iib/workers/tasks/build_merge_index_image.py
+++ b/iib/workers/tasks/build_merge_index_image.py
@@ -128,13 +128,16 @@ def _add_bundles_missing_in_source(
             missing_bundles.append(bundle)
             missing_bundle_paths.append(bundle['bundlePath'])
 
-    if ignore_bundle_ocp_version and target_index is not None:
+    if ignore_bundle_ocp_version:
+        target_index_tmp = '' if target_index is None else target_index
         allow_no_ocp_version = any(
-            target_index.startswith(index)
+            target_index_tmp.startswith(index) or source_from_index.startswith(index)
             for index in get_worker_config()['iib_no_ocp_label_allow_list']
         )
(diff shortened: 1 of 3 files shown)


yashvardhannanavati (approved):

iib/workers/tasks/build_merge_index_image.py
```diff
@@ -128,9 +128,10 @@ def _add_bundles_missing_in_source(
             missing_bundles.append(bundle)
             missing_bundle_paths.append(bundle['bundlePath'])
 
-    if ignore_bundle_ocp_version and target_index is not None:
+    if ignore_bundle_ocp_version:
+        target_index_tmp = '' if target_index is None else target_index
         allow_no_ocp_version = any(
-            target_index.startswith(index)
+            target_index_tmp.startswith(index) or source_from_index.startswith(index)
             for index in get_worker_config()['iib_no_ocp_label_allow_list']
```
xDaile, line 135: Does this change take into account that "random_text".startswith('') will result in True?

xDaile (approved):

iib/workers/tasks/build_merge_index_image.py
lipoja: well, it did not . However I added config validation to ensure nobody sets empty string in the allow list.
Please re-review

xDaile: My bad, when I was thinking about this, I switched the parameters (I thought the target index was checked in `startswith()` method). However, this check is useful in my opinion.

xDaile (approved):

PR #600, opened 2023-12-20
Use container-tool podman for _opm_index_add commands

yashvardhannanavati:
Refers to CLOUDDST-20876

### DIFF
--- a/iib/workers/tasks/build.py
+++ b/iib/workers/tasks/build.py
@@ -540,6 +540,7 @@ def _opm_index_rm(
     binary_image: str,
     from_index: str,
     overwrite_from_index_token: Optional[str] = None,
+    container_tool: Optional[str] = None,
 ) -> None:
     """
     Remove the input operators from the operator index.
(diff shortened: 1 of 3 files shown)


iib/workers/tasks/build.py
```diff
@@ -1220,6 +1227,7 @@ def handle_rm_request(
                 binary_image=prebuild_info['binary_image'],
                 from_index=from_index_resolved,
                 overwrite_from_index_token=overwrite_from_index_token,
+                container_tool='podman',
```
JAVGan, line 1230: I wonder whether do we want to have this hardcoded as it is or configurable so it can be turned on/off without need to rewrite the code.

Did someone discuss about this?

yashvardhannanavati: The reason I did not make it configurable is because it is not used for every request. It is only used for non-fbc builds which are all EOL as of today and this piece of code will only cease to exist in the months to come. Having said that, I am open to making it configurable if guild members think otherwise.

JAVGan: Got it, thanks!

JAVGan (approved):

PR #628, opened 2024-02-01
Fixing logging in tracing.py

lipoja:
[CLOUDDST-21024]


xDaile (approved):

iib/common/tracing.py
```diff
@@ -81,37 +81,37 @@ def __new__(cls):
...
+            else:
+                context = propagator.extract(carrier={})
+
+            log.info('Context inside %s: %s', span_name, context)
             if kwargs.get('traceparent'):
                 log.debug('traceparent is %s' % str(kwargs.get('traceparent')))
                 carrier = {'traceparent': kwargs.get('traceparent')}
                 trace_context = propagator.extract(carrier)
-                log.debug('Context is %s', trace_context)
+                log.info('Context is %s', trace_context)
```
chandwanitulsi, line 114: Hey @lipoja , should these log message be visible in build logs?

lipoja: I am sorry - what build log are you referring to?

lipoja: I was debugging it and `kwargs` is always empty :( ... so we do not hit this part of code

xDaile (approved):

iib/common/tracing.py
```diff
@@ -35,6 +35,7 @@ def func():
 
 
 log = logging.getLogger(__name__)
+log.setLevel(logging.DEBUG)
```
xDaile, line 38: For some reason, this line sets logging to the debug level for the whole IIB.

xDaile (approved):

iib/common/tracing.py
chandwanitulsi: Thanks, Jan, I checked more and got this bit.
Should this be fixed as well? and will it be part of this PR?

lipoja: it will not be part of this PR - I think it needs more investigation / understanding.

yashvardhannanavati (approved):

PR #629, opened 2024-02-05
Check request state, before its update

xDaile:
[Related JIRA issue - CLOUDDST-21515](https://issues.redhat.com/browse/CLOUDDST-21515)

When we mark a request that is in a queue, as failed, the IIB worker still picks it up, as it is still in the queue (`apply_async` method). IIB worker sets the state to `in_progress` at the beginning of each handle method with a call to the `update_request` method under the hood, no matter what, the actual state of the request is.

With the proposed changes, the method `update_request` now checks if the request is already in the final state. If so, it will raise an exception, and continue with other requests in the queue. To be able to not retry this 'update_request' method (retry decorator), a new exception was needed (different from IIBError). 

Also, an edit in the `failed_request_callback` was needed, as this method also updates the state of the request.

### DIFF
--- a/iib/workers/api_utils.py
+++ b/iib/workers/api_utils.py
@@ -14,7 +14,7 @@ from tenacity import (
     wait_exponential,
 )
 
-from iib.exceptions import IIBError
+from iib.exceptions import IIBError, FinalStateAlreadyReached
 from iib.workers.config import get_worker_config
 from iib.workers.tasks.iib_static_types import UpdateRequestPayload
 import time
@@ -168,6 +168,14 @@ def update_request(
             rv.status_code,
             rv.text,
         )
+        rv_error = rv.json().get('error', None)
+        if rv_error in [
+            "A failed request cannot change states",
+            "A complete request cannot change states",
+        ]:
+            raise FinalStateAlreadyReached(
+                f"Request cannot be updated, as the request is already in the final state"
+            )
         if exc_msg:
             _exc_msg = exc_msg.format(**payload, request_id=request_id)
         else:
--- a/iib/exceptions.py
+++ b/iib/exceptions.py
@@ -13,6 +13,10 @@ class IIBError(BaseException):
     """An error was encountered in IIB."""
 
 
+class FinalStateAlreadyReached(BaseException):
+    """Request is already in the final state."""
+
+
 class ValidationError(BaseException):
     """Denote invalid input."""
 
--- a/iib/workers/tasks/general.py
+++ b/iib/workers/tasks/general.py
@@ -4,7 +4,7 @@ from typing import Any
 
 import celery.app.task
 
-from iib.exceptions import IIBError
+from iib.exceptions import IIBError, FinalStateAlreadyReached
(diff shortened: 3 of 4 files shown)


iib/workers/api_utils.py
```diff
@@ -150,6 +150,11 @@ def update_request(
     # Prevent a circular import
     start_time = time.time()
     request_url = f'{config.iib_api_url.rstrip("/")}/builds/{request_id}'
+
+    log.info(f'Checking the current state of request {request_id}')
+    if requests_auth_session.get(request_url).json()['state'] in ['failed', 'complete']:
```
xDaile, line 155: I was also thinkin about importing `RequestStateMapping.get_final_states()` from `models.py`, however I did not found single example with import from `iib.web` in the `iib.workers` part. (This could possibly create circular import I believe).

lipoja (approved):

yashvardhannanavati:
Are we sure it's not happening because the request execution time is reaching its timeout and the broker is assuming the task has to be requed? Do we really need to make this change?

iib/workers/api_utils.py
```diff
@@ -168,6 +168,14 @@ def update_request(
             rv.status_code,
             rv.text,
         )
+        rv_error = rv.json().get('error', None)
+        if rv_error in [
+            "A failed request cannot change states",
+            "A complete request cannot change states",
+        ]:
+            raise FinalStateAlreadyReached(
```
yashvardhannanavati, line 176: Please correct me if I am mistaken, but it looks like the change is gracefully handling the second execution of the request and not preventing it from being executed again in the first place?

yashvardhannanavati (changes requested):

xDaile:
> Are we sure it's not happening because the request execution time is reaching its timeout and the broker is assuming the task has to be requed? Do we really need to make this change?

The issue I tried to address with this change, was the case, when someone submits the request, then the user marks this request as failed (while in the queue), and the worker picks the requests as if it was not marked as failed.

iib/workers/api_utils.py
chandwanitulsi: While this PR will not solve the initially thought bug, but isn't it better to have this improvement as we already have it now?
With this change the time spent on retries for modifying the failed requests would be saved. 
IMO it will surely make sense to close the PR if its adding too much complication, but if its not and the benefits outweigh the complication, it could be worth keeping it.

xDaile:
Currently, proposed changes will not be merged, because, in the mentioned scenario, the RabbitMQ queue should have been purged, which would result in the requests dropping out of the queue. When we do not purge the queue (either on purpose or not) and mark some requests as failed, time consumption improvement is not worth adding complexity to the IIB project.

xDaile:
Currently, proposed changes will not be merged, because, in the mentioned scenario, the RabbitMQ queue should have been purged, which would result in the requests dropping out of the queue. When we do not purge the queue (either on purpose or not) and mark some requests as failed, time consumption improvement is not worth adding complexity to the IIB project.

PR #631, opened 2024-02-22
Adding attributes to span

lipoja:
[CLOUDDST-21023]

### DIFF
--- a/iib/common/common_utils.py
+++ b/iib/common/common_utils.py
@@ -1,0 +1,27 @@
+# SPDX-License-Identifier: GPL-3.0-or-later
+from typing import Dict
+
+
+def get_binary_versions() -> Dict:
+    """
+    Return string containing version of binary files used by IIB.
+
+    :return: Dictionary with all binary used and their version
+    :rtype: dict
+    """
+    from iib.workers.tasks.utils import run_cmd
+
+    opm_version_cmd = ['opm', 'version']
+    podman_version_cmd = ['podman', '-v']
+    buildah_version_cmd = ['buildah', '-v']
+
+    try:
+        return {
+            'opm': run_cmd(opm_version_cmd, exc_msg='Failed to get opm version.').strip(),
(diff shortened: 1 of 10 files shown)


chandwanitulsi (approved):

iib/common/common.py
```diff
@@ -0,0 +1,38 @@
...
+            'podman': run_cmd(podman_version_cmd, exc_msg='Failed to get podman version.').strip(),
+            'buildah': run_cmd(
+                buildah_version_cmd, exc_msg='Failed to get buildah version.'
+            ).strip(),
+        }
+    except FileNotFoundError:
+        return {'opm': '', 'podman': '', 'buildah': ''}
+
+
+def get_binary_versions_text() -> str:
```
yashvardhannanavati, line 30: It feels like we can get rid of this function as it's only formatting text, maybe add it to the above function?

iib/common/common.py
```diff
@@ -0,0 +1,38 @@
+# SPDX-License-Identifier: GPL-3.0-or-later
```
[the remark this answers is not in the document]
lipoja: renamed

iib/common/common.py
lipoja: removed

iib/common/common_utils.py
```diff
@@ -0,0 +1,27 @@
...
+    try:
+        return {
+            'opm': run_cmd(opm_version_cmd, exc_msg='Failed to get opm version.').strip(),
+            'podman': run_cmd(podman_version_cmd, exc_msg='Failed to get podman version.').strip(),
+            'buildah': run_cmd(
+                buildah_version_cmd, exc_msg='Failed to get buildah version.'
+            ).strip(),
+        }
+    except FileNotFoundError:
+        return {'opm': '', 'podman': '', 'buildah': ''}
```
yashvardhannanavati, line 27: nitpick: maybe log the error too?

yashvardhannanavati (approved):

PR #635, opened 2024-03-04
Backup index.db to prevent opm bug consequence

xDaile:
Related JIRA issue: [CLOUDDST-20957](https://issues.redhat.com/browse/CLOUDDST-20957)
Related OPM bug: [OCPBUGS-30214](https://issues.redhat.com/browse/OCPBUGS-30214)

Method `opm_registry_add_fbc` calls `_opm_registry_add.
  
The first time method `opm_registry_add_fbc` calls  `_opm_registry_add`,  `_opm_registry_add` fails. This failure is the correct expected behavior, however, the OPM command modifies the database file (should not be modified on an error). 

As this  `_opm_registry_add` method has the retry decorator, it is triggered again, now failing with a different error (the error is a consequence of the database modification in the previous run attempt). Only the second error (wrong one) is propagated to the parent `opm_registry_add_fbc`  method. (As we have the retry decorator with the `opm_registry_add_fbc`  method, this whole process is repeated). 

As a result, the error, resulting from running the OPM command on the modified database file, is propagated into the state_reason. 

In the solution, I propose a workaround, which creates a backup of the database file, then runs the OPM command and uses backup on command failure.

### DIFF
--- a/iib/workers/tasks/opm_operations.py
+++ b/iib/workers/tasks/opm_operations.py
@@ -574,7 +574,18 @@ def _opm_registry_add(
         log.info('Using force to add bundle(s) to index')
         cmd.extend(['--overwrite-latest'])
 
-    run_cmd(cmd, {'cwd': base_dir}, exc_msg='Failed to add the bundles to the index image')
+    # after commit 643fc9499222107f52b7ba3f9f3969fa36812940 index.db backup was added
+    # due to the opm bug https://issues.redhat.com/browse/OCPBUGS-30214
+    index_db_backup = index_db + ".backup"
+    shutil.copyfile(index_db, index_db_backup)
+
+    try:
+        run_cmd(cmd, {'cwd': base_dir}, exc_msg='Failed to add the bundles to the index image')
+    except Exception as e:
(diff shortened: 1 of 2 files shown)


iib/workers/tasks/opm_operations.py
```diff
@@ -512,12 +512,7 @@ def generate_cache_locally(base_dir: str, fbc_dir: str, local_cache_path: str) -
         raise IIBError(error_msg)
 
 
-@retry(
-    before_sleep=before_sleep_log(log, logging.WARNING),
-    reraise=True,
-    retry=retry_if_exception_type(IIBError),
-    stop=stop_after_attempt(2),
-)
+# Retry decorator was previously used here, but removed due to OPM bug
```
yashvardhannanavati, line 515: might be good to post the PR number in the comments so we know what to revert.

lipoja (approved):

xDaile:
Logs from worker with implemented workaround (local run)
[workaround_logs.txt](https://github.com/release-engineering/iib/files/14515450/workaround_logs.txt)

PR #637, opened 2024-03-04
Use multiple OPM versions in IIB

chandwanitulsi:
IIB will have multiple versions of opm installed.
A specific opm version determined by index image version will be used for the IIB operation.
Refers to CLOUDDST-20065


iib/workers/tasks/opm_operations.py
```diff
@@ -907,3 +907,224 @@ def verify_operator_exists(
...
+    :param str index_image: from_index_image for the request
+    :return str opm_path: opm_path to be used for the entire request
+
+    """
+    global opm_path
+    log.info("Determining wthe OPM version to use")
+    worker_config = get_worker_config()
+    opm_paths_config = worker_config.get("iib_ocp_opm_mapping")
+    if not opm_paths_config:
+            log.error("OPM config is not set")
```
lipoja, line 926: Is it really an error? I think this could be backwards compatible. Therefore I would just logged info message that iib_ocp_opm_mapping is not set. Unless we want to make it mandatory.

lipoja: Should net be the message something like "iib_ocp_opm_mapping configuration is not set"

iib/workers/tasks/opm_operations.py
```diff
@@ -907,3 +907,224 @@ def verify_operator_exists(
...
+    if index_image:
+        index_version =_find_index_version(index_image)
+        for ocp_version, opm_version in opm_paths_config.items():
+            if ocp_version == index_version:
+                opm_path = opm_version
+                break
```
lipoja, line 932: I am wondering if simple `opm_path = opm_paths_config.get(index_version, opm_paths_config['default'])` would not be enough? Instead of this for loop.

iib/workers/tasks/opm_operations.py
lipoja: When `opm_paths_config` is not set I think we should call `return` and do not continue.

iib/workers/config.py
```diff
@@ -218,6 +218,19 @@ class DevelopmentConfig(Config):
         str
     ] = '/var/lib/requests/recursive_related_bundles'
     iib_dogpile_backend: str = 'dogpile.cache.memcached'
+    iib_ocp_opm_mapping: dict = {
```
lipoja, line 221: I am missing some config validation in which I would like  to check if all these omp version really exists in file system. If not raise an error.

iib/workers/tasks/opm_operations.py
```diff
@@ -907,3 +907,224 @@ def verify_operator_exists(
...
+def set_opm_verison(index_image: str) -> str:
+    """
+    This function will verify if opm config is set in worker
+    if the config entry exists for the index image version
+    and return the opm version path to be used in the entire IIB request.
+    :param str index_image: from_index_image for the request
+    :return str opm_path: opm_path to be used for the entire request
+
+    """
+    global opm_path
```
lipoja, line 921: Maybe I missed that but I do not see this global variable defined anywhere. 

Somewhere at the top of this file should be `opm_path = "omp"`

iib/workers/tasks/opm_operations.py
```diff
@@ -89,7 +89,7 @@ def opm_serve(catalog_dir: str) -> Tuple[int, subprocess.Popen]:
 
     for port in _gen_port_for_grpc():
         try:
-            cmd = ['opm', 'serve', catalog_dir, '-p', str(port), '-t', '/dev/null']
+            cmd = [opm_path, 'serve', catalog_dir, '-p', str(port), '-t', '/dev/null']
```
lipoja, line 92: I am missing `global opm_path` in these functions (here and below)

iib/workers/tasks/opm_operations.py
```diff
@@ -907,3 +907,224 @@ def verify_operator_exists(
...
     return is_package_in_index, index_db_path
+
+
+def set_opm_verison(index_image: str) -> str:
+    """
+    This function will verify if opm config is set in worker
+    if the config entry exists for the index image version
+    and return the opm version path to be used in the entire IIB request.
+    :param str index_image: from_index_image for the request
+    :return str opm_path: opm_path to be used for the entire request
```
lipoja, line 918: I do not see any returns. This function just sets the global variable but does not return anything (which is correct).

lipoja (changes requested):
I think this is good way how to implement it this change. We just need to polish it and test it little bit. Thank you for working on it!

iib/workers/tasks/utils.py
```diff
@@ -798,7 +756,7 @@ def run_cmd(
         log.error('The command "%s" failed with: %s', ' '.join(cmd), response.stderr)
         regex: str
         match: Optional[re.Match]
-        if cmd[0] == 'opm':
+        if cmd[0].endswith('opm'):
```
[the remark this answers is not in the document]
chandwanitulsi: yeah I don't remember why I updated this, will definitely change

iib/workers/tasks/utils.py
```diff
@@ -842,15 +800,15 @@ def _get_binary_versions():
     :return: String with all binary used and their version
     :rtype: str
     """
-    opm_version_cmd = ['opm', 'version']
+    # opm_version_cmd = [opm_path, 'version']
```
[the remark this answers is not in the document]
chandwanitulsi: The opm path is not set when this function is called at the start of the request, so I need to find a better placement

iib/workers/tasks/opm_operations.py
chandwanitulsi: I will try this :+1:

iib/workers/tasks/utils.py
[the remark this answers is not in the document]
lipoja: We can use decorator for setting the opm_path as well and place it before this one.

iib/workers/tasks/opm_operations.py
chandwanitulsi: Even without the` global opm_path` I am able to use the opm_path set in set_opm_verison()

iib/workers/tasks/opm_operations.py
chandwanitulsi: I added it earlier at the top of file as you have suggested, but I don't see any changes with adding that or in the current implementation, could you please share how declaring at the top as well can help here?

iib/workers/tasks/opm_operations.py
chandwanitulsi: I am not sure whether it should be mandatory but we definitely need to support multiple opm in IIB soon

iib/workers/tasks/opm_operations.py
lipoja: Note: use global keyword is necessary inside a function when you want to modify the variable. For accessing it is not necessary.
----------------------------
I leave it up to you.

iib/workers/tasks/utils.py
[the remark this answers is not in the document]
chandwanitulsi: Updated

iib/workers/tasks/utils.py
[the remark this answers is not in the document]
chandwanitulsi: As discussed, setting the decorator only wont be very helpful as logs would indicate 2 different versions of opm at times, so will take the route of displaying all available opms here

iib/workers/tasks/opm_operations.py
chandwanitulsi: updated already

iib/workers/tasks/opm_operations.py
chandwanitulsi: Updated

iib/workers/config.py
chandwanitulsi: Updated

iib/workers/config.py
```diff
@@ -347,6 +361,21 @@ def validate_celery_config(conf: app.utils.Settings, **kwargs) -> None:
...
+    Validate iib_ocp_opm_mapping config variable
+    :param dict iib_ocp_opm_mapping: the value of iib_ocp_opm_mapping variable
+    :raises iib.exceptions.ConfigError: if the configuration is invalid
+    """
+    if iib_ocp_opm_mapping is not None:
+        if not isinstance(iib_ocp_opm_mapping, dict):
+            raise ConfigError('iib_ocp_opm_mapping must be a dictionary')
+        opms_defined = set(iib_ocp_opm_mapping.values())
+        for opm_version in opms_defined:
+            if not os.path.exists(f'/usr/bin/{opm_version}'):
```
lipoja, line 375: I would go with this kind of check:
https://stackoverflow.com/questions/377017/test-if-executable-exists-in-python

Personally, first thing what come to my mind was: https://docs.python.org/3.12/library/shutil.html#shutil.which

I know we are installing opm in our playbooks to /usr/bin but since this is opensource and opm could be put anywhere in $PATH then I would go with `which` command.

iib/workers/tasks/opm_operations.py
```diff
@@ -907,3 +911,238 @@ def verify_operator_exists(
...
+    ]
+    if container_tool:
+        cmd.append('--container-tool')
+        cmd.append(container_tool)
+    with set_registry_token(overwrite_target_index_token, from_index):
+        run_cmd(cmd, {'cwd': base_dir}, exc_msg='Failed to deprecate the bundles')
+
+
+def _find_index_version(index_image: str) -> str:
+    from iib.workers.tasks.utils import get_image_label
```
lipoja, line 1096: docstring

iib/workers/tasks/opm_operations.py
```diff
@@ -907,3 +911,238 @@ def verify_operator_exists(
...
+    opm_paths_config = worker_config.get("iib_ocp_opm_mapping")
+    if opm_paths_config is None or from_index is None:
+        log.warning("Either iib_ocp_opm_mapping config is not set or from_index is not set, using the default opm")
+        return
+    index_version =_find_index_version(from_index)
+    opm_path = opm_paths_config.get(index_version, opm_paths_config["default"])
+    log.info("opm_path set to %s", opm_path)
+
+
+def set_opm(func):
```
lipoja, line 1125: docstring

And I guess this is just for testing. There are too many `.warning()` for debug.

iib/workers/tasks/utils.py
```diff
@@ -842,15 +800,20 @@ def _get_binary_versions():
...
     podman_version_cmd = ['podman', '-v']
     buildah_version_cmd = ['buildah', '-v']
 
-    opm_version = run_cmd(opm_version_cmd, exc_msg='Failed to get opm version.').strip()
     podman_version = run_cmd(podman_version_cmd, exc_msg='Failed to get podman version.').strip()
     buildah_version = run_cmd(buildah_version_cmd, exc_msg='Failed to get buildah version.').strip()
-
-    return f"opm {opm_version}\n{podman_version}\n{buildah_version}"
+    binary_versions = f"{podman_version}\n{buildah_version}"
+    opms_present = [f for f in os.listdir('/usr/bin') if re.match(r'opm', f)]
```
lipoja, line 810: I think we can leverage `iib_ocp_opm_mapping`. 
Something like: 
```python
opms_present = set(iib_ocp_opm_mapping.values()) if iib_ocp_opm_mapping is not None else ['opm']
```

lipoja (changes requested):

iib/workers/tasks/utils.py
chandwanitulsi: updated

iib/workers/tasks/opm_operations.py
chandwanitulsi: Yeah I missed removing those before committing, updated the patch

iib/workers/tasks/opm_operations.py
chandwanitulsi: Added

iib/workers/config.py
chandwanitulsi: Updated, this is simpler, Thank you :)

iib/workers/tasks/opm_operations.py
```diff
@@ -918,3 +922,254 @@ def verify_operator_exists(
...
+
+    opm version is based on from_index/source_from_index.
+
+    :param str from_index: from_index_image for the request
+    """
+    global opm_path
+    opm_path = 'opm'
+    log.info("Determining the OPM version to use")
+    worker_config = get_worker_config()
+    opm_paths_config = worker_config.get("iib_ocp_opm_mapping")
```
yashvardhannanavati, line 1139: Could you please add this variable to the docs?

iib/workers/tasks/opm_operations.py
```diff
@@ -918,3 +922,254 @@ def verify_operator_exists(
...
+
+def set_opm_verison(from_index: Optional[str] = None):
+    """
+    Set the opm version to be used for the entire IIB operation.
+
+    opm version is based on from_index/source_from_index.
+
+    :param str from_index: from_index_image for the request
+    """
+    global opm_path
```
yashvardhannanavati, line 1135: Do we have to set a global variable? Can we set a default in the base `Config` object that gets overwritten if a value is set in iib-playbooks?

iib/workers/tasks/opm_operations.py
```diff
@@ -918,3 +922,254 @@ def verify_operator_exists(
...
+
+    index_image_version = index_image.split(":")[1]
+    if index_image_version.startswith("v4."):
+        return index_image_version
+    else:
+        index_version = get_image_label(index_image, 'com.redhat.index.delivery.version')
+        return index_version
+
+
+def set_opm_verison(from_index: Optional[str] = None):
```
yashvardhannanavati, line 1127: Maybe define this function inside `set_opm` for clarity ?

yashvardhannanavati (changes requested):

iib/workers/tasks/opm_operations.py
chandwanitulsi: Thanks for catching, will add in next patch

iib/workers/tasks/opm_operations.py
chandwanitulsi: I was thinking about that as well, but a separate function was easily testable and IMO separate function gave more readability.

iib/workers/tasks/opm_operations.py
chandwanitulsi: I don't think in this case we need to update anything at config level, but can you give an example how that could help?

iib/workers/tasks/opm_operations.py
```diff
@@ -918,3 +922,254 @@ def verify_operator_exists(
...
+        run_cmd(cmd, {'cwd': base_dir}, exc_msg='Failed to add the bundles to the index image')
+
+
+@retry(
+    before_sleep=before_sleep_log(log, logging.WARNING),
+    reraise=True,
+    retry=retry_if_exception_type(IIBError),
+    stop=stop_after_attempt(2),
+)
+def opm_index_rm(
```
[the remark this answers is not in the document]
yashvardhannanavati: Agreed that most of the tests would get affected. But wouldn't that just be a simple find and replace to fix?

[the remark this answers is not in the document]
yashvardhannanavati: I just want to make sure we are able to revert specific changes

iib/workers/tasks/opm_operations.py
yashvardhannanavati: well, that had an opposite effect on me and I had to find the decorator function to make sense of this one. Maybe we can move it after the decorator function in the file if you prefer to keep them separate

iib/workers/tasks/opm_operations.py
yashvardhannanavati: The approach I suggested was to set `opm_path='opm'` in the `Config` class as default. Then, you change that variable on line 1146 if the user has specified `opm_paths_config`. That should eliminate the need of having a global variable.

iib/workers/tasks/utils.py
```diff
@@ -799,7 +757,7 @@ def run_cmd(
         log.error('The command "%s" failed with: %s', ' '.join(cmd), response.stderr)
         regex: str
         match: Optional[re.Match]
-        if cmd[0] == 'opm':
+        if cmd[0].startswith('opm'):
```
yashvardhannanavati, line 760: Are we assuming the opm binary will be in a directory that's in the PATH? Is something stopping me from specifying `/usr/yashn/opm-1.26.33` in the config? If not, should this be changed to regex that checks for the `opm` string in the value?

lipoja: My assumption was that opm is in $PATH. But you have a good point. 
We should check both - $PATH and if the path set in config is executable opm.

iib/workers/tasks/opm_operations.py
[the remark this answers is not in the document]
chandwanitulsi: Updated the MR to contain 2 commits

iib/workers/tasks/opm_operations.py
chandwanitulsi: Sure, updated

iib/workers/tasks/opm_operations.py
chandwanitulsi: Its still unclear how that can benefit instead of using a global variable, I meant the example of same in the earlier comment.
Also, config variables generally remain same for the entire service and for all request types, so changing a config variable in every request seemed a bit odd

iib/workers/tasks/utils.py
chandwanitulsi: sounds valid, I updated in next patch

iib/workers/tasks/opm_operations.py
yashvardhannanavati: Apologies my previous comment wasn't clear enough.

Right, I am not suggesting to change the config variable during every request. 

In opm_operations.py, `opm_path='opm'` is being initialized in 3 places. Let's say if we have to change the way the opm command is used in the future, like `/usr/bin/opm`, we would have to replace it in 3 places. Having a config variable makes it easier to change, can avoid potential bugs in the future and make the code easier to understand. My suggestion is to use `opm_path=worker_config.opm_path` and then modify `opm_path`

Secondly, using global variable generally makes the code difficult to understand and maintain especially when it's being updated inside the functions. Maybe we can think about creating a class with `__init__`, `get_opm_path` and `set_opm_path` functions? 

@chandwanitulsi @lipoja @xDaile thoughts on the class?

yashvardhannanavati (changes requested):

iib/workers/config.py
```diff
@@ -348,6 +363,22 @@ def validate_celery_config(conf: app.utils.Settings, **kwargs) -> None:
             )
 
 
+def _validate_multiple_opm_mapping(iib_ocp_opm_mapping: Dict[str, str]) -> None:
```
xDaile, line 366: I would add a check, that the default value is present here.

iib/workers/tasks/opm_operations.py
```diff
@@ -918,3 +922,259 @@ def verify_operator_exists(
...
+    def wrapper(*args, **kwargs):
+        from iib.workers.tasks.utils import _get_function_arg_value
+
+        if 'from_index' in inspect.getfullargspec(func).args:
+            from_index = _get_function_arg_value('from_index', func, args, kwargs)
+        # if target_index is present instead of from_index
+        if 'target_index' in inspect.getfullargspec(func).args:
+            from_index = _get_function_arg_value('target_index', func, args, kwargs)
+        try:
+            set_opm_verison(from_index)
```
xDaile, line 1132: typo verision

iib/workers/tasks/opm_operations.py
```diff
@@ -918,3 +922,259 @@ def verify_operator_exists(
...
+
+    opm version is based on from_index/source_from_index.
+
+    :param str from_index: from_index_image for the request
+    """
+    global opm_path
+    opm_path = 'opm'
+    log.info("Determining the OPM version to use")
+    worker_config = get_worker_config()
+    opm_paths_config = worker_config.get("iib_ocp_opm_mapping")
```
xDaile, line 1153: Would there be an error from black/flake8 when it would in the following shape?
`get_worker_config().get("iib_ocp_opm_mapping")`

iib/workers/tasks/opm_operations.py
```diff
@@ -918,3 +922,259 @@ def verify_operator_exists(
...
+            from_index = _get_function_arg_value('from_index', func, args, kwargs)
+        # if target_index is present instead of from_index
+        if 'target_index' in inspect.getfullargspec(func).args:
+            from_index = _get_function_arg_value('target_index', func, args, kwargs)
+        try:
+            set_opm_verison(from_index)
+            func(*args, **kwargs)
+        finally:
+            global opm_path
+            opm_path = 'opm'
```
xDaile, line 1136: I am not sure if I understand, but will this line overwrite the previously set opm path in the set_opm_version()?

iib/workers/tasks/opm_operations.py
```diff
@@ -918,3 +922,259 @@ def verify_operator_exists(
...
+        The format of the token must be in the format "user:password".
+    :param bool overwrite_csv: a boolean determining if a bundle will be replaced if the CSV
+        already exists.
+    :param str container_tool: the container tool to be used to operate on the index image
+    :raises IIBError: if the ``opm index add`` command fails.
+    """
+    # The bundles are not resolved since these are stable tags, and references
+    # to a bundle image using a digest fails when using the opm command.
+
+    from iib.workers.tasks.utils import run_cmd, set_registry_token
```
xDaile, line 968: Is there a reason for doing the same import multiple times at this place (opm_index_rm, derpecate_bundles too)?

iib/workers/tasks/opm_operations.py
xDaile: I am in favour of using the class object here, as it would clear some uncertainties related to the usage of global variables.

lipoja: I agree, we should have one variable with default "opm" path. And it might be part of worker config.
And yes, there are better ways then using global variable.  So the ideal to have it as class attribute sounds good.

iib/workers/tasks/utils.py
```diff
@@ -799,7 +757,7 @@ def run_cmd(
         log.error('The command "%s" failed with: %s', ' '.join(cmd), response.stderr)
         regex: str
         match: Optional[re.Match]
-        if cmd[0] == 'opm':
+        if 'opm' in cmd[0]:
```
lipoja, line 760: What about something more foolproof
```python
from pathlib import Path

if Path(cmd[0]).stem.startswith('opm'):
```

iib/common/common_utils.py
```diff
@@ -11,13 +12,19 @@ def get_binary_versions() -> Dict:
     """
     from iib.workers.tasks.utils import run_cmd
 
-    opm_version_cmd = ['opm', 'version']
     podman_version_cmd = ['podman', '-v']
     buildah_version_cmd = ['buildah', '-v']
 
+    worker_config = get_worker_config()
+    iib_ocp_opm_mapping = worker_config.get("iib_ocp_opm_mapping")
+    opms_present = set(iib_ocp_opm_mapping.values()) if iib_ocp_opm_mapping is not None else ['opm']
```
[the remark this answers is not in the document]
chandwanitulsi: It is logged in splunk, its in opm_operations.py set_opm_version(). 
This function is also run before the opm version to be used in the request is determined. Hence, in this function all opm versions are logged and then set_opm_version()., the opm version that will be used in request is logged

iib/workers/tasks/opm_operations.py
chandwanitulsi: Shouldn't be an issue, will check

iib/workers/tasks/opm_operations.py
chandwanitulsi: Yes, we want the during the entire handle_add/rm request a specific opm should be used, and then after the function completes, it should be set back to default

iib/workers/tasks/opm_operations.py
chandwanitulsi: I have mostly moved these functions but I also believe these were imported withi the function to avoid circular dependencies

iib/workers/config.py
chandwanitulsi: Hmm, in that case I think it will make sense to remove the default value from the mapping if we are setting the default opm in config

iib/workers/tasks/opm_operations.py
chandwanitulsi: So I read a lot online regarding the use of global variables and class attr, it looks like there are strong opinions outside for both approaches. 
To go with majority I am exploring the class attribute way. I also read that get() and set() methods are not ideal way as well as we can effectively use class.attr or obj.attr directly without needing to have getters and setters (using property rather than getter and setter sounded like a preferred approach), therefore I am avoiding the complete get/set way but using the set(), also with this approach, we will have to access the attr everywhere by class.attr

For the other approach Michal mentioned about putting opm_*() functions inside the class as well, we can definitely do that, but then in that case, some functions like opm_index_add, opm_index_rm, and deprecate_bundles will become instance methods, but there will still be several other functions in the file, which does not need to be instance methods and IMO it would be better to keep those outside the class, but that could make the file structure a bit confusing

README.md
```diff
@@ -435,7 +435,9 @@ The custom configuration options for the Celery workers are listed below:
 * `iib_supported_archs` - the architectures supported by IIB. IIB can build index images for these
   architectures. The dictionary has mapping of arch aliases with formal names like
   `{"arm64": "aarch64"}`
-
+* `iib_ocp_opm_mapping` - the mapping for which opm version should be used for
+    a particular ocp version like
+    `{"v4.15": "opm-v1.28.0"}`
```
JAVGan, line 440: Shouldn't `iib_default_opm` be added here as well?

iib/workers/config.py
JAVGan: What about validating that whenever `iib_ocp_opm_mapping` is None the default should be set?

iib/workers/tasks/opm_operations.py
```diff
@@ -918,3 +930,258 @@ def verify_operator_exists(
...
+
+        :param str from_index: from_index_image for the request
+        """
+        log.info("Determining the OPM version to use")
+        opm_versions_config = get_worker_config().get('iib_ocp_opm_mapping')
+        if opm_versions_config is None or from_index is None:
+            log.warning(
+                "Either iib_ocp_opm_mapping config or from_index is not set, using the default opm"
+            )
+            return
```
JAVGan, line 1160: With this I understand we should enforce the definition of the `iib_default_opm`, no?

Because the `Opm.opm_version` uses a `get()` operation to retrieve the value from config which might be `None`, so we must be certain that it isn't.

JAVGan: Also I believe it would be nice to print the default `opm` version at least in this warning.

iib/workers/config.py
chandwanitulsi: Updated the validation such that default to be set, irrespective of iib_ocp_opm_mapping is there or not

iib/workers/tasks/opm_operations.py
chandwanitulsi: The default opm version would be printed in common_utils.py get_binary_versions()

iib/workers/config.py
chandwanitulsi: Thanks for the review :)

iib/workers/tasks/opm_operations.py
chandwanitulsi: we should really clean up these imports and fix circular dependencies, but I believe that should be in scope of another PR

iib/workers/tasks/opm_operations.py
chandwanitulsi: updated the config for validating the default setting as well

iib/workers/tasks/opm_operations.py
xDaile: Sorry, my terminology was not good. What I meant with "class object" was the creation of the OPM class instance with incoming requests, where, for example, the OPM version could be set during the __init__() method. I mentioned moving the methods under this class, as it would allow us to use the instance opm version automatically.

lipoja (dismissed):

xDaile (approved):

JAVGan (dismissed):

iib/workers/tasks/opm_operations.py
```diff
@@ -1,10 +1,12 @@
+import inspect
```
yashvardhannanavati, line 1: nit: could you please arrange these in the alphabetical order?

iib/workers/tasks/build_merge_index_image.py
```diff
@@ -196,6 +197,7 @@ def _add_bundles_missing_in_source(
 @instrument_tracing(
     span_name="workers.tasks.build.handle_merge_request", attributes=get_binary_versions()
 )
+@Opm.set_opm
```
yashvardhannanavati, line 200: I am not convinced by the idea of having a decorator inside the class. May I ask what's stopping us from having a single @classmethod inside the class like `set_opm_version` which will set the class variable if `None` otherwise return the class variable as is?

yashvardhannanavati (changes requested):

docker/Dockerfile-workers
```diff
@@ -24,7 +24,11 @@ RUN dnf -y install \
     && dnf update -y \
     && dnf clean all
 
-ADD https://github.com/operator-framework/operator-registry/releases/download/v1.26.4/linux-amd64-opm /usr/bin/opm
+ADD https://github.com/operator-framework/operator-registry/releases/download/v1.26.4/linux-amd64-opm /usr/bin/opm-v1.26.4
+RUN chmod +x /usr/bin/opm-v1.26.4
+ADD https://github.com/operator-framework/operator-registry/releases/download/v1.28.0/linux-amd64-opm /usr/bin/opm-v1.28.0
+RUN chmod +x /usr/bin/opm-v1.28.0
+RUN ln -s /usr/bin/opm-v1.28.0 /usr/bin/opm
```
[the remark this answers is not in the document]
chandwanitulsi: Only one of the them is not moved to path location, the link is created to ensure we have a default opm as well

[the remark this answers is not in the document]
yashvardhannanavati: ack, makes sense. Could you please mention that in a comment?

iib/workers/config.py
```diff
@@ -285,13 +300,17 @@ def validate_celery_config(conf: app.utils.Settings, **kwargs) -> None:
     if not conf.get('iib_api_url'):
         raise ConfigError('iib_api_url must be set')
 
+    if not conf.get('iib_default_opm'):
```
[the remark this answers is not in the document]
chandwanitulsi: so that means we don't need to ensure its set in playbooks as well?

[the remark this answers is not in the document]
lipoja: If we set `iib_default_opm` in `Config` class then it is optional and does not have to be set playbook. 
For me it is preferred since it will keep setting of IIB backwards compatible.

iib/workers/tasks/opm_operations.py
```diff
@@ -1104,3 +1116,72 @@ def deprecate_bundles(
...
+    Find the OCP version of the given index image.
+
+    :param str index_image: index_image pull spec
+    :return: version of index image
+    :rtype: str
+
+    """
+    from iib.workers.tasks.utils import get_image_label
+
+    index_image_version = index_image.split(":")[1]
```
yashvardhannanavati, line 1130: I don't think we should rely on the `tag` to get the OCP version of the index image. Only the label should be considered unless there's a use case for using the tag. Even if there is a use case, the first choice should be the label, no?

yashvardhannanavati (commented):
Just a single comment. Apologies for missing that earlier. Looks great! 

Thanks for being patient with me on this. If you could merge the last two commits into one, we can merge the PR post that :)

README.md
```diff
@@ -506,6 +510,7 @@ Greenwave to check if your bundle image builds have passed the tests in the Gree
 have defined. The IIB request submitted to that queue will succeed only if the policy is satisfied.
 
 # Documentation
+
```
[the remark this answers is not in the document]
chandwanitulsi: my linting kept adding this on save, I will remove

iib/workers/tasks/opm_operations.py
chandwanitulsi: The reason I did not take label as first choice was that from the tag we just need to fetch the version, whereas from the label, we inspect the image and then get the label so just pure operational steps wise, tag was first preference

chandwanitulsi:
> Just a single comment. Apologies for missing that earlier. Looks great!
> 
> Thanks for being patient with me on this. If you could merge the last two commits into one, we can merge the PR post that :)

The second commit can be seen as a complete improvement on the first one, could yo share how merge the last two commits into one would help?

iib/workers/tasks/opm_operations.py
yashvardhannanavati: right, but using the tag for getting the index version is unreliable and error prone. Nothing stops a user from specifying the tag as `v4.<random-string>` for a v4.17 index and then IIB would use the default OPM (which might not be the correct one)

iib/workers/config.py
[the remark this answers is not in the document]
yashvardhannanavati: I do not have strong opinions on this one so I am good either way. But if we start doing validations for one default config value, we are setting a precedent for all. I'll leave this call on Tulsi

yashvardhannanavati:
> > Just a single comment. Apologies for missing that earlier. Looks great!
> > Thanks for being patient with me on this. If you could merge the last two commits into one, we can merge the PR post that :)
> 
> The second commit can be seen as a complete improvement on the first one, could yo share how merge the last two commits into one would help?

Yes, in the git history, it would seem like we added the decorator change to the code base (which we never did as it wasn't the preferred approach) and then we changed it to using the classmethod. While it is an improvement on the first commit in the context of this PR, in the context of the repository it should be a single addition of using the classmethod

iib/workers/config.py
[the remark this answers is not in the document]
lipoja: @chandwanitulsi I guess I did not read it after myself. What I meant is that we do not need validation if the value is present since it will be there by default.

iib/workers/tasks/opm_operations.py
lipoja: I am here with Yash. We should use label as the reference for index image version. Do not worry about performance. It will be fast since there is memcache in place and we get this info on multiple places so we are fetching the data only once.

chandwanitulsi: Yup, already removed the function

iib/workers/config.py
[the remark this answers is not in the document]
chandwanitulsi: Thank you :)

lipoja (approved):

xDaile (approved):

PR #654, opened 2024-03-26
Fix comma bundle version validation

xDaile:
This PR resolves the difference between Pyxis and IIB bundle version validations.

Method `is_bundle_version_valid()` compares comma style list with ocp_version. The problem here was, that the result could be different, based on the order of the items in the list. On the other hand, code in [Pyxis ocpversion.py](https://gitlab.cee.redhat.com/metadata/pyxis/pyxis/-/blob/master/pyxis/graphql/custom/operator_index/ocpversion.py) approaches this check differently - they sort the list before the comparison. 

Related issue: [CLOUDDST-22327](https://issues.redhat.com/browse/CLOUDDST-22327)

### DIFF
--- a/tests/test_workers/test_tasks/test_build_merge_index_image.py
+++ b/tests/test_workers/test_tasks/test_build_merge_index_image.py
@@ -631,7 +631,7 @@ def test_add_bundles_missing_in_source_none_missing(
             'csvName': 'bundle2-2.0',
         },
     ]
-    mock_gil.side_effect = ['v=4.5', 'v4.7,v4.6', 'v4.5-v4.8', 'v4.5,v4.6,v4.7']
+    mock_gil.side_effect = ['v=4.5', 'v4.8,v4.7', 'v4.5-v4.8', 'v4.5,v4.6,v4.7']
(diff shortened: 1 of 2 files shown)


tests/test_workers/test_tasks/test_build_merge_index_image.py
```diff
@@ -631,7 +631,7 @@ def test_add_bundles_missing_in_source_none_missing(
             'csvName': 'bundle2-2.0',
         },
     ]
-    mock_gil.side_effect = ['v=4.5', 'v4.7,v4.6', 'v4.5-v4.8', 'v4.5,v4.6,v4.7']
+    mock_gil.side_effect = ['v=4.5', 'v4.8,v4.7', 'v4.5-v4.8', 'v4.5,v4.6,v4.7']
```
xDaile, line 634: This one case is an exact example of what was wrong. IIB should have been comparing 4.6 with the ocp_version parameter `4.6`, but the list's first value  `4.7` was used.

JAVGan (approved):

yashvardhannanavati (approved):

PR #655, opened 2024-04-04
Fix bug with FBC YAML Operators

JAVGan:
When an operator has catalog in yaml insted of json, some requests (like add) fails on "duplicate package" as `merge_catalogs_dirs` function generates FBC from hidden index.db created as json files and then copy over its content to final location where are all yaml files.

To prevent this issue this simple fix will walk through the config directory and convert any YAML files into JSON to ensure that everything uses the same format in the index image.

Refers to CLOUDDST-21432

### DIFF
--- a/iib/workers/tasks/fbc_utils.py
+++ b/iib/workers/tasks/fbc_utils.py
@@ -3,13 +3,18 @@
 import os
 import logging
 import shutil
+import json
+from pathlib import Path
 
 from typing import Tuple, List
+import ruamel.yaml
+
 from iib.exceptions import IIBError
 from iib.workers.config import get_worker_config
 from iib.common.tracing import instrument_tracing
 
 log = logging.getLogger(__name__)
+yaml = ruamel.yaml.YAML()
 
 
 def is_image_fbc(image: str) -> bool:
@@ -86,6 +91,7 @@ def merge_catalogs_dirs(src_config: str, dest_config: str):
             msg = f"config directory does not exist: {conf_dir}"
             log.error(msg)
             raise IIBError(msg)
+        enforce_json_config_dir(conf_dir)
 
     log.info("Merging config folders: %s to %s", src_config, dest_config)
     shutil.copytree(src_config, dest_config, dirs_exist_ok=True)
@@ -116,3 +122,24 @@ def extract_fbc_fragment(temp_dir: str, fbc_fragment: str) -> Tuple[str, List[st
         raise IIBError("No operator packages in fbc_fragment %s", fbc_fragment)
 
     return fbc_fragment_path, operator_packages
+
+
+def enforce_json_config_dir(config_dir: str) -> None:
+    """
+    Ensure the files from config dir are in JSON format.
+
+    It will walk recursively and convert any YAML files to the JSON format.
+
+    :param str config_dir: The config dir to walk recursively converting any YAML to JSON.
+    """
+    log.info("Enforcing JSON content on config_dir: %s", config_dir)
+    for dirpath, _, filenames in os.walk(config_dir):
+        for file in filenames:
+            in_file = os.path.join(dirpath, file)
+            if in_file.lower().endswith(".yaml"):
+                out_file = os.path.join(dirpath, f"{Path(in_file).stem}.json")
+                log.debug(f"Converting {in_file} to {out_file}.")
+                with open(in_file, 'r') as yaml_in, open(out_file, 'w') as json_out:
+                    data = yaml.load(yaml_in)
+                    json.dump(data, json_out)
+                os.remove(in_file)
--- a/iib/workers/tasks/fbc_utils.py
+++ b/iib/workers/tasks/fbc_utils.py
@@ -5,8 +5,8 @@ import logging
 import shutil
 import json
 from pathlib import Path
-
 from typing import Tuple, List
+
 import ruamel.yaml
 
 from iib.exceptions import IIBError
@@ -86,12 +86,15 @@ def merge_catalogs_dirs(src_config: str, dest_config: str):
     :param str src_config: source config directory
     :param str dest_config: destination config directory
     """
+    from iib.workers.tasks.opm_operations import opm_validate
+
     for conf_dir in (src_config, dest_config):
         if not os.path.isdir(conf_dir):
             msg = f"config directory does not exist: {conf_dir}"
             log.error(msg)
             raise IIBError(msg)
         enforce_json_config_dir(conf_dir)
+        opm_validate(conf_dir)
 
     log.info("Merging config folders: %s to %s", src_config, dest_config)
     shutil.copytree(src_config, dest_config, dirs_exist_ok=True)
--- a/iib/workers/tasks/fbc_utils.py
+++ b/iib/workers/tasks/fbc_utils.py
@@ -93,11 +93,11 @@ def merge_catalogs_dirs(src_config: str, dest_config: str):
             msg = f"config directory does not exist: {conf_dir}"
             log.error(msg)
             raise IIBError(msg)
-        enforce_json_config_dir(conf_dir)
-        opm_validate(conf_dir)
 
     log.info("Merging config folders: %s to %s", src_config, dest_config)
     shutil.copytree(src_config, dest_config, dirs_exist_ok=True)
+    enforce_json_config_dir(conf_dir)
+    opm_validate(conf_dir)
 
 
 def extract_fbc_fragment(temp_dir: str, fbc_fragment: str) -> Tuple[str, List[str]]:
--- a/iib/workers/tasks/opm_operations.py
+++ b/iib/workers/tasks/opm_operations.py
@@ -366,6 +366,7 @@ def opm_migrate(
 
     run_cmd(cmd, {'cwd': base_dir}, exc_msg='Failed to migrate index.db to file-based catalog')
     log.info("Migration to file-based catalog was completed.")
+    opm_validate(fbc_dir_path)
 
     if generate_cache:
         # Remove outdated cache before generating new one
@@ -1116,6 +1117,20 @@ def deprecate_bundles(
         cmd.append(container_tool)
     with set_registry_token(overwrite_target_index_token, from_index):
         run_cmd(cmd, {'cwd': base_dir}, exc_msg='Failed to deprecate the bundles')
+
+
+def opm_validate(config_dir: str) -> None:
+    """
+    Validate the declarative config files in a given directory.
+
+    :param str config_dir: directory containing the declarative config files.
+    :raises IIBError: if the validation fails
+    """
+    from iib.workers.tasks.utils import run_cmd
+
+    log.info("Validating files under %s", config_dir)
+    cmd = [Opm.opm_version, 'validate', config_dir]
+    run_cmd(cmd, exc_msg=f'Failed to validate the content from config_dir {config_dir}')
 
 
 class Opm:
--- a/tests/test_workers/test_tasks/test_opm_operations.py
+++ b/tests/test_workers/test_tasks/test_opm_operations.py
@@ -152,6 +152,7 @@ def test_serve_cmd_at_port_delayed_initialize(
     assert mock_run_cmd.call_count == 7
 
 
+@mock.patch('iib.workers.tasks.opm_operations.opm_validate')
 @mock.patch('iib.workers.tasks.opm_operations.shutil.rmtree')
 @mock.patch('iib.workers.tasks.opm_operations.generate_cache_locally')
 @mock.patch('iib.workers.tasks.utils.run_cmd')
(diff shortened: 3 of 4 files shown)


JAVGan:
@release-engineering/exd-guild-hello-operator PTAL.

I took the liberty to convert all files in the `config` dir to JSON as it's the default one from `opm`, having the solution "supporting" YAML in input but always outputting in JSON. Please let me know whether it's OK to go with this.

lipoja:
@JAVGan I have to think about this solution - tomorrow.
My assumption was that the error was caused by `opm migrate` where we we generated `json` but copied it over to `/config` dir so it ended up with two files (yaml, json) for one operator and then everything crashed when we were generating cache. 

My idea was to detect what format is on the input as a source_image.  Then use `opm migrate -o yaml` or `json`.
This would work for our use-case - bootstrapping - with empty target_image.

JAVGan:
> My idea was to detect what format is on the input as a source_image. Then use opm migrate -o yaml or json.
This would work for our use-case - bootstrapping - with empty target_image.

@lipoja I was thinking about this, but what would happen if we receive multiple operators (to add, for example) being one of them, say, JSON and the rest YAML? Or vice-versa? That's why I thought on streamlining everything, but sure, we can discuss this tomorrow.

iib/workers/tasks/fbc_utils.py
```diff
@@ -3,13 +3,17 @@
 import os
 import logging
 import shutil
+import json
+from pathlib import Path
 
+import ruamel.yaml
```
yashvardhannanavati, line 9: nit: while you're at it, could you please add a newline after `from typing import Tuple` to correctly segregate the imports?

yashvardhannanavati (dismissed):

yashvardhannanavati:
> > My idea was to detect what format is on the input as a source_image. Then use opm migrate -o yaml or json.
> > This would work for our use-case - bootstrapping - with empty target_image.
> 
> @lipoja I was thinking about this, but what would happen if we receive multiple operators (to add, for example) being one of them, say, JSON and the rest YAML? Or vice-versa? That's why I thought on streamlining everything, but sure, we can discuss this tomorrow.

Yeah, I think agree with Jonathan's reasoning here since we are supporting adding multiple operators in a single fragment.

lipoja:
Yeah I know what you mean. I thought we could go with opm generating the output. The reason behind that is that opm is "consuming" those files so it would not generate anything that is faulty. 
If we go this way I think we should validate the output -to be sure everything is correct and `opm` can load it. 
`opm validate` might help us - in my opinion we should use this before every build.

iib/workers/tasks/fbc_utils.py
```diff
@@ -118,3 +123,23 @@ def extract_fbc_fragment(temp_dir: str, fbc_fragment: str) -> Tuple[str, str]:
         raise IIBError("More than 1 package is present in fbc_fragment %s", fbc_fragment)
 
     return fbc_fragment_path, operator_packages[0]
+
+
+def enforce_json_config_dir(config_dir: str) -> None:
```
lipoja, line 128: I would like to have info message saying that we are converting yaml to json.

iib/workers/tasks/fbc_utils.py
```diff
@@ -118,3 +123,23 @@ def extract_fbc_fragment(temp_dir: str, fbc_fragment: str) -> Tuple[str, str]:
...
+    """
+    for dirpath, _, filenames in os.walk(config_dir):
+        for file in filenames:
+            in_file = os.path.join(dirpath, file)
+            if in_file.lower().endswith(".yaml"):
+                out_file = os.path.join(dirpath, f"{Path(in_file).stem}.json")
+                log.debug(f"Converting {in_file} to {out_file}.")
+                with open(in_file, 'r') as yaml_in, open(out_file, 'w') as json_out:
+                    data = yaml.load(yaml_in)
+                    json.dump(data, json_out)
```
lipoja, line 144: I would love to see some validation that converted output is correct.

iib/workers/tasks/fbc_utils.py
```diff
@@ -86,6 +90,7 @@ def merge_catalogs_dirs(src_config: str, dest_config: str):
             msg = f"config directory does not exist: {conf_dir}"
             log.error(msg)
             raise IIBError(msg)
+        enforce_json_config_dir(conf_dir)
```
lipoja, line 93: I am on the edge if we should enforce it even when both inputs are in yaml. 
Do you think that there might be situation when yaml format is preferred? 

Just to discuss every corner-case:
I mean team maintaining opted-in operators might be surprised when the catalog is converted to json.
I am not sure if they use the resulted image for some tests - is not here any team that is extracting the data put inside the index and expecting yaml?

lipoja (changes requested):

iib/workers/tasks/fbc_utils.py
JAVGan: That's something I want to know as well: Does it make a huge difference in the operator PoV to have YAML or JSON files? Because if this matters then enforcing one format all over the place is really not good, but I'm not sure how OLM deals with this.

What could be done is to, like, verify the format of the first operator and enforce that format, I'm just not sure whether this would be overengineering or not...

Either way, if the format type for OLM operators really matters, instead of just the content, I may update the `opm migrate -o {FORMAT}` in IIB but I agree discussing this with more people would be ideal

iib/workers/tasks/fbc_utils.py
JAVGan: There's a validation in `tests/test_workers/test_tasks/test_fbc_utils.py:L175` already, do you mean adding an extra validation layer within the `fbc_utils.py` as well? Like, reopening the written data and ensuring it has the same data and if not raise IIBError?

JAVGan: Oh, I get it now, `opm validate`, I'll implement it

JAVGan: Added in abbc6e6452762c5810c40dfc5d2d783abc2e8b10

iib/workers/tasks/fbc_utils.py
JAVGan: Added

iib/workers/tasks/opm_operations.py
```diff
@@ -918,3 +918,17 @@ def verify_operator_exists(
             is_package_in_index = True
             log.info("operator package %s found in index_db %s", operator_package, index_db_path)
     return is_package_in_index, index_db_path
+
+
+def opm_validate(config_dir: str) -> None:
+    """
+    Validate the declarative config JSON files in a given directory.
```
lipoja, line 925: nitpick: I think that `opm validate` does not do difference between YAML or JSON. So we can remove the word `JSON` from here. And below (docstring and log message)

tests/test_workers/test_tasks/test_opm_operations.py
```diff
@@ -754,3 +754,12 @@ def test_verify_operator_exists(
     )
     mock_ors.assert_has_calls([mock.call(db_path=index_db_path)])
     assert package_exists == opr_exists
+
+
+@mock.patch('iib.workers.tasks.utils.run_cmd')
+def test_opm_validate(mock_rc, tmpdir):
```
lipoja, line 760: I am not sure if this test is meaningful. 
It is just testing if `run_cmd` really run the command. 

I would suggest to move this test to `test_merge_catalogs_dirs` where it makes sense to test if `opm validate` was called.

But we can keep this one as well. But it would be nice to test it in the `test_merge_catalogs_dirs`

lipoja (changes requested):

iib/workers/tasks/fbc_utils.py
lipoja: @xDaile @chandwanitulsi @yashvardhannanavati thoughts?

chandwanitulsi: Very valid question. 
AFAIK product teams use OLM to test/use their operators and OLM will support both json and yaml format without any differences to user so we should be good here. But its possible to have outliers in product teams, I have reached out to Jordan to verify whether we are expected to have any specific format of files in index image based on inputs, let's see what he suggests

iib/workers/tasks/opm_operations.py
JAVGan: I've removed the _JSON_ word from docstrings

tests/test_workers/test_tasks/test_opm_operations.py
JAVGan: Ok, I've moved this check to the `test_merge_catalog_dirs`, I also agree it makes more sense

JAVGan:
Rebased

iib/workers/tasks/opm_operations.py
```diff
@@ -1144,3 +1144,17 @@ def set_opm_version(cls, from_index: Optional[str] = None):
...
+    """
+    Validate the declarative config files in a given directory.
+
+    :param str config_dir: directory containing the declarative config files.
+    :raises IIBError: if the validation fails
+    """
+    from iib.workers.tasks.utils import run_cmd
+
+    log.debug("Validating JSON files under %s", config_dir)
+    cmd = ['opm', 'validate', config_dir]
```
yashvardhannanavati, line 1159: Should this be `Opm.opm_version` since we merged the change for multiple OCP versions

JAVGan: Nice catch!

I updated it accordingly.

JAVGan: PTAL again

iib/workers/tasks/fbc_utils.py
chandwanitulsi: Based on the suggestion, I don't think product teams would check the content and expect a certain format of catalog file in the index, so we can go ahead with this

iib/workers/tasks/opm_operations.py
```diff
@@ -1116,6 +1116,20 @@ def deprecate_bundles(
...
+def opm_validate(config_dir: str) -> None:
+    """
+    Validate the declarative config files in a given directory.
+
+    :param str config_dir: directory containing the declarative config files.
+    :raises IIBError: if the validation fails
+    """
+    from iib.workers.tasks.utils import run_cmd
+
+    log.debug("Validating JSON files under %s", config_dir)
```
chandwanitulsi, line 1128:
```suggestion
    log.debug("Validating files under %s", config_dir)
```

iib/workers/tasks/fbc_utils.py
```diff
@@ -81,11 +86,15 @@ def merge_catalogs_dirs(src_config: str, dest_config: str):
...
     """
+    from iib.workers.tasks.opm_operations import opm_validate
+
     for conf_dir in (src_config, dest_config):
         if not os.path.isdir(conf_dir):
             msg = f"config directory does not exist: {conf_dir}"
             log.error(msg)
             raise IIBError(msg)
+        enforce_json_config_dir(conf_dir)
+        opm_validate(conf_dir)
```
chandwanitulsi, line 97: If we are adding opm validate, how about we add validate after opm migrate every time before we generate the dockerfile or build the index image? this will add more robustness to our built index image

xDaile (commented):
I agree to changes, however I IMO do not have enough expertise to approve this.

iib/workers/tasks/fbc_utils.py
chandwanitulsi: @lipoja Ah yes, I read your comment thoroughly this time :)

JAVGan: I've added a call to `opm_validate` after the migrate command under `opm_migrate`, please let me know whether it's what you're expecting.

iib/workers/tasks/opm_operations.py
JAVGan: Fixed, PTAL again

iib/workers/tasks/fbc_utils.py
lipoja: OK, my assumption was that it can validate SQLite as well but it can't. 
Then I guess it should be enough to have it in this function. 
What about to move it at the bottom of this function? So we do not call it twice but validate the actual result of merge?

JAVGan: That's a great idea!

I'll adjust it

JAVGan: @lipoja added on 7c0b5189b6f07208ed6cdb7cccd33e0760234731

yashvardhannanavati (dismissed):

chandwanitulsi (dismissed):

JAVGan:
@lipoja @chandwanitulsi @yashvardhannanavati I've rebased and fixed the conflicts, please give a final review so I can merge it

PR #661, opened 2024-04-08
Allow non tagged index image in IIB_GRAPH_MODE_INDEX_ALLOW_LIST

xDaile:
Related isssue: CLOUDDST-22149

This change allows us to specify index images without a tag to use with the graph_update_mode parameter.


tests/test_web/test_api_v1.py
```diff
@@ -884,6 +911,7 @@ def test_add_bundle_success(
 
     rv = client.post('/api/v1/builds/add', json=data, environ_base=auth_env)
     rv_json = rv.json
+    print(rv_json)
```
JAVGan, line 914: Do we need this print in the test? I guess it was some leftover, right? xD

xDaile: Nice catch! It somehow slipped through.

iib/web/models.py
```diff
@@ -815,7 +815,10 @@ def validate_graph_mode(graph_update_mode: Optional[str], index_image: Optional[
                 f'"graph_update_mode" must be set to one of these: {graph_mode_options}'
             )
         allowed_from_indexes: List[str] = current_app.config['IIB_GRAPH_MODE_INDEX_ALLOW_LIST']
```
chandwanitulsi, line 817: Is there a plan to update IIB_GRAPH_MODE_INDEX_ALLOW_LIST config? 
lets say we want to use graph_update_mode on registry.stage.redhat.io/redhat/community-operator-index:v4.17, with this approach IIB will raise an error for 4.17 index, I will still have to make sure that the 4.17 index is added to the config list, which we are trying to avoid

iib/web/models.py
```diff
@@ -815,7 +815,10 @@ def validate_graph_mode(graph_update_mode: Optional[str], index_image: Optional[
...
-        if index_image not in allowed_from_indexes:
+        if index_image is None or (
```
yashvardhannanavati, line 818: should this be 

```
if index_image not in allowed_from_indexes:
    if index_image is None or index_image.split(":")[0] not in allowed_from_indexes:
        raise Forbidden...
```

xDaile: Yes, I like this approach. Updated.

iib/web/models.py
xDaile: Please check now. Old rows from the config should work as well as the potential new ones (without tags).

iib/web/models.py
```diff
@@ -816,10 +816,11 @@ def validate_graph_mode(graph_update_mode: Optional[str], index_image: Optional[
             )
         allowed_from_indexes: List[str] = current_app.config['IIB_GRAPH_MODE_INDEX_ALLOW_LIST']
         if index_image not in allowed_from_indexes:
-            raise Forbidden(
-                '"graph_update_mode" can only be used on the'
-                f' following index image: {allowed_from_indexes}'
-            )
+            if index_image is None or index_image.split(":")[0] not in allowed_from_indexes:
```
lipoja, line 819: Is not this creating backward incompatibility? 
I mean right now "users" are used to specify pull spec for index image and when they deploy this change all index images will work. That might be unexpected right?

tests/test_web/test_api_v1.py
```diff
@@ -578,12 +579,35 @@ def test_add_bundles_graph_update_mode_not_allowed(
...
     error_msg = (
         '"graph_update_mode" can only be used on the'
-        ' following index image: [\'some-unique-index\']'
+        f' following index image: {graph_update_mode_list}'
     )
     assert error_msg == rv.json['error']
     mock_smfsc.assert_not_called()
 
 
+@pytest.mark.parametrize('from_index', ('some-common-index:v4.15', 'another-common-index:v4.15'))
```
lipoja, line 588: please add also:  'another-common-index:v4.14', but which test should fail in current settings.

iib/web/models.py
xDaile: How do you mean all index images will work?  Can you give an example?
Old pull-spec (image:tag) will work as the condition on line 818 will result in True.

lipoja: Apologies, I overlooked the first condition. Makes sense, thank you!

tests/test_web/test_api_v1.py
xDaile: If I am correct, this is already achieved in the test 'test_add_bundles_graph_update_mode_not_allowed' with 'some-common-index:v4.15' as equivalent to  'another-common-index:v4.14'. (Graph_update mode list contains 'some-unique-index' and  'some-common-index:v1.14'.

chandwanitulsi (approved):
lgtm, Jan's comment should be addressed before merge and we should also ensure that in the release, IIB_GRAPH_MODE_INDEX_ALLOW_LIST is updated in playbooks

xDaile:
> lgtm, Jan's comment should be addressed before merge and we should also ensure that in the release, IIB_GRAPH_MODE_INDEX_ALLOW_LIST is updated in playbooks

We do not need to update IIB_GRAPH_MODE_INDEX_ALLOW_LIST, as the solution will also work with the current config.

chandwanitulsi:
> > lgtm, Jan's comment should be addressed before merge and we should also ensure that in the release, IIB_GRAPH_MODE_INDEX_ALLOW_LIST is updated in playbooks
> 
> We do not need to update IIB_GRAPH_MODE_INDEX_ALLOW_LIST exactly before merge, as the solution will also work with the current config.

Not before merge, before the release or create another issue to update the config in playbook, otherwise we might forget to update

xDaile:
> > > lgtm, Jan's comment should be addressed before merge and we should also ensure that in the release, IIB_GRAPH_MODE_INDEX_ALLOW_LIST is updated in playbooks
> > 
> > 
> > We do not need to update IIB_GRAPH_MODE_INDEX_ALLOW_LIST exactly before merge, as the solution will also work with the current config.
> 
> Not before merge, before the release or create another issue to update the config in playbook, otherwise we might forget to update

[Opened Gitlab MR](https://gitlab.cee.redhat.com/ansible-playbooks/iib-playbooks/-/merge_requests/595) with updated config (If we merge Gitlab MR first or together with this PR, we do not need to think of this during the next release).

tests/test_web/test_api_v1.py
lipoja: I was blind. Thanks.

lipoja (approved):

PR #669, opened 2024-04-30
Add atomic port locking mechanism

xDaile:
Related issue: CLOUDDST-22624

As seen in RCA: gitops 1.10.3 not added to prod 4.13 index and also in CLOUDDST-22350, there was an issue where two workers picked the same port at the same VM. 

This decorator with atomic operation `os.open(lock_file, os.O_CREAT | os.O_EXCL)` should resolve this issue. 

Few notes:
- I did not use `continue` within the loop, as it was applied within the try-except block. Instead I used this `port_found` flag.
- os.open(lock_file, os.O_CREAT | os.O_EXCL) is atomic according to [man pages](https://linux.die.net/man/3/open)

### DIFF
--- a/iib/workers/config.py
+++ b/iib/workers/config.py
@@ -3,7 +3,7 @@ import os
 import logging
 import shutil
 import types
-from typing import Any, Dict, List, Optional, Type, Union
+from typing import Any, Dict, List, Optional, Tuple, Type, Union
 
 from kombu import Queue
 from celery import Celery, app
@@ -34,9 +34,13 @@ class Config(object):
     )
     iib_greenwave_url: Optional[str] = None
     iib_grpc_init_wait_time: int = 100
-    iib_grpc_max_port_tries: int = 100
     iib_grpc_max_tries: int = 5
-    iib_grpc_start_port: int = 50051
+    # size of both ranges, needs to be the same, ranges neeeds to be exclusive
+    iib_opm_port_ranges: Dict[str, Tuple[int, int]] = {
+        "opm_port": (50051, 50151),
+        "opm_pprof_port": (50151, 50251),
+    }
+    iib_opm_pprof_lock_required_min_version = "1.29.0"
     iib_image_push_template: str = '{registry}/iib-build:{request_id}'
     iib_index_image_output_registry: Optional[str] = None
     iib_log_level: str = 'INFO'
--- a/iib/workers/tasks/opm_operations.py
+++ b/iib/workers/tasks/opm_operations.py
@@ -1,10 +1,14 @@
+from functools import wraps
 import logging
 import os
 import random
 import shutil
+import socket
 import subprocess
+import tempfile
 import time
-from typing import List, Optional, Tuple, Generator, Union, Set
+from typing import Callable, Generator, List, Optional, Set, Tuple, Union
+from packaging.version import Version
 
 from tenacity import (
     before_sleep_log,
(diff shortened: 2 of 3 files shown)


iib/workers/config.py
```diff
@@ -36,6 +36,7 @@ class Config(object):
     iib_grpc_init_wait_time: int = 100
     iib_grpc_max_port_tries: int = 100
     iib_grpc_max_tries: int = 5
+    iib_grpc_port_lock_dir = '/tmp/'
```
lipoja, line 39: Can we get this by calling `tempfile.gettempdir()` 
Just need to be sure it returns "global" temp and not just the one related for user.

xDaile: Updated to `lock_dir = tempfile.gettempdir()`.

JAVGan:
@xDaile I prefer the parametrized decorator as we can easily configure the ranges whenever required. You may proceed with this approach

iib/workers/tasks/opm_operations.py
```diff
@@ -26,29 +28,56 @@
...
+            lock_dir = tempfile.gettempdir()
+
+            # Attempt to acquire the lock for each port in the range (shuffled order)
+            port_found = False
+            for port in port_stack:
+                lock_file = os.path.join(lock_dir, f'iib_port_{port}.lock')
+
+                # Create the lock file atomically
+                try:
+                    os.open(lock_file, os.O_CREAT | os.O_EXCL)
```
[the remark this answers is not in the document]
lipoja: I thought we might use something already implemented: https://pypi.org/project/filelock/

xDaile:
> We also have to think about a use case when a lock file is created and then we restart the worker. The file will stay there locked. Is there some way how to prevent that?

I was thinking about cronjob, which would delete the lock files older than some time (one hour should be more than enough I think). 
Do you have some other solution in mind?

version_with_class.py
```diff
@@ -0,0 +1,191 @@
...
+from typing import Dict, Tuple, List, Generator, Callable, List
+from iib.exceptions import IIBError, AddressAlreadyInUse
+
+
+class Config:
+    iib_opm_port_ranges: Dict[str, Tuple[int, int]] = {
+        "opm_port": (50051, 50151),
+        "opm_ppof_port": (50151, 50251),
+    }
+    iib_opm_pprof_lock_required_versions = ["1.40.0"]
```
lipoja, line 16: Let's not keep a list of versions with pprof. I believe that it will be present there from now on. And it looks like it is there since opm version [v1.29.0](https://github.com/operator-framework/operator-registry/releases/tag/v1.29.0)

version_with_class.py
```diff
@@ -0,0 +1,191 @@
...
+    # The port stack is empty - we tried out all ports from allowed range
+    # therefore we will raise and IIB error
+    err_msg = f'No free port has been found after {num_of_attempts} attempts.'
+    print(err_msg)
+    raise Exception(err_msg)
+
+def get_port_stacks() -> Tuple[List[List[int]], List[str]]:
+    conf = get_worker_config()
+    ports_description = list(conf.iib_opm_port_ranges)
+    if Opm.opm_version in conf.iib_opm_pprof_lock_required_versions:
```
lipoja, line 79: This might be simple `if Opm.opm_version >= 1.29.0` (however it would be great to have the version in config)

xDaile: @lipoja please check

iib/workers/tasks/opm_operations.py
```diff
@@ -26,29 +31,194 @@
...
+
+        if self.locked:
+            log.debug(f"Error: Port {self.port} is already locked")
+            raise IIBError("Port lock attempt on already locked port")
+
+        # check if the port is free, opm service is doing the check too,
+        # however this way, we do not have to rely on their error message format
+        s = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+        try:
+            s.bind(("localhost", self.port))
```
lipoja, line 65: You have to also close the port before calling opm.

iib/workers/tasks/opm_operations.py
```diff
@@ -26,29 +31,194 @@
 log = logging.getLogger(__name__)
 
 
-def _gen_port_for_grpc() -> Generator[int, None, None]:
+class PortFileLock:
+    """A class representing file-lock used during OPM operations."""
+
+    def __init__(self, purpose: str, port: int, lock_on_init: bool = False):
+        """Initialize the PortFileLock object."""
```
lipoja, line 38: Once we agree on this solution - please fill in docsctrig.

iib/workers/tasks/opm_operations.py
```diff
@@ -26,29 +31,194 @@
...
+        random.shuffle(port_stacks)
+
+        # Attempt to acquire the lock for each port in the range (shuffled order)
+        lock_success = False
+
+        # Initialize the generator
+        gen = port_file_locks_generator(port_stacks=port_stacks, port_purposes=port_purposes)
+
+        # Use the function to retrieve values from the generator
+        while lock_success is not True:
```
lipoja, line 188: `while lock_success is not True`  
`while lock_success is False` ?
`while not lock_success` ?

iib/workers/tasks/opm_operations.py
```diff
@@ -74,69 +244,88 @@ def opm_serve_from_index(base_dir: str, from_index: str) -> Tuple[int, subproces
...
+    pprof_port: Optional[int] = None,
+) -> Tuple[int, subprocess.Popen]:
     """
     Locally start OPM registry service, which can be communicated with using gRPC queries.
 
     Due to IIB's paralellism, the service can run multiple times, which could lead to port
     binding conflicts. Resolution of port conflicts is handled in this function as well.
 
+    :param int opm_port: OPM port number obtained from create_port_filelock decorator
+    :param int opm_port: Pprof opm port number obtained from create_port_filelock decorator
```
lipoja, line 302: opm_port -> pprof_port

iib/workers/tasks/opm_operations.py
```diff
@@ -74,69 +244,88 @@ def opm_serve_from_index(base_dir: str, from_index: str) -> Tuple[int, subproces
...
+    pprof_port: Optional[int] = None,
+) -> Tuple[int, subprocess.Popen]:
     """
     Locally start OPM service, which can be communicated with using gRPC queries.
 
     Due to IIB's paralellism, the service can run multiple times, which could lead to port
     binding conflicts. Resolution of port conflicts is handled in this function as well.
 
+    :param int opm_port: OPM port number obtained from create_port_filelock decorator
+    :param int opm_port: Pprof opm port number obtained from create_port_filelock decorator
```
lipoja, line 260: opm_port -> pprof_port

lipoja (changes requested):

iib/workers/tasks/opm_operations.py
lipoja: What about this one? Should we close the port?

iib/workers/tasks/opm_operations.py
lipoja: Lets add the doc string and tests if we want to release it tomorrow.

iib/workers/tasks/opm_operations.py
xDaile: Fixed, please check

iib/workers/tasks/opm_operations.py
```diff
@@ -26,31 +30,205 @@
...
+    random.shuffle(ports_list)
+
+    return ports_list, port_purposes
+
+
+def create_port_filelocks(func: Callable) -> Callable:
+    """
+    Create a file-lock on random port from the configured range.
+
+    :param function func: the function to be decorated    :return: the decorated function
```
lipoja, line 181: `:return: the decorated function` - on new line

lipoja (changes requested):

iib/workers/tasks/opm_operations.py
```diff
@@ -26,31 +30,206 @@
...
+    port_stacks: List[List[int]],
+    port_purposes: List[str],
+) -> Generator[List[PortFileLock], None, None]:
+    """
+    Generate PortFileLock from port_stacks and port_purposes.
+
+    Each item in the port_stacks list represents set of ports, from which list of PortFileLocks
+    will be generated.
+
+    :param List[List[int]] port_stacks: List with List of port numbers for one generated item
```
lipoja, line 112: nitpick: should we stick to this type of typing or should we use the one we use everywhere?
`:param list(list(int)) port_stacks: `

iib/workers/tasks/opm_operations.py
```diff
@@ -26,31 +30,206 @@
...
+            s.bind(("localhost", self.port))
+        except socket.error:
+            err_msg = f"Port {self.port} is already in use."
+            log.debug(err_msg)
+            raise AddressAlreadyInUse(err_msg)
+        finally:
+            s.close()
+
+        # create file-lock
+        try:
```
lipoja, line 77: I think it would be better to combine the locking and testing port in one block of code:
```python
         try:
            s.bind(("localhost", self.port))
            f = os.open(self.filename, os.O_CREAT | os.O_EXCL | os.O_RDWR)
            self.locked = True
            log.debug(f"Port {self.port} used as {self.purpose} was locked.")
        except socket.error:
            err_msg = f"Port {self.port} is already in use."
            log.debug(err_msg)
            raise AddressAlreadyInUse(err_msg)
        except FileExistsError: 
            s.close()
            err_msg = f"Port {self.port} is already locked by other IIB worker."
            log.debug(err_msg)
            raise AddressAlreadyInUse(err_msg)
        finally:
            s.close()
            f.close()
```

lipoja (changes requested):

iib/workers/tasks/opm_operations.py
xDaile: There seem to be two issues (solved in the last commit) with this solution. 
- socket.error is probably a subclass of FileExistsError, as in this order every time it got caught by socket.error
- we need to check if we opened the file to be able  to close it like that

Please check the last commit, to see the difference. 
I am okay with either solution, which one would you prefer?

iib/workers/tasks/opm_operations.py
lipoja: Well, we do not have it standardized - but everywhere else is used just standard python notation

iib/workers/tasks/opm_operations.py
lipoja: Pssst, just hit [ENTER] and push it :)

iib/workers/tasks/opm_operations.py
```diff
@@ -26,31 +30,213 @@
...
+        return f"PortFileLock(port: {self.port}, purpose: {self.purpose}, locked: {self.locked})"
 
-    log.debug('Get random ports from range [%d, %d)', port_start, port_end)
+    def lock_acquire(self):
+        """
+        Create a file representing port lock.
+
+        Before trying to create a a port-lock file, we try to check if the port is free.
+        """
+        log.debug(f"Attempt to lock port {self.port}.")
```
lipoja, line 66: Could you, please, use `%s` notation. 
https://docs.python.org/3/howto/logging.html#optimization

JAVGan (dismissed):

lipoja (approved):

PR #673, opened 2024-05-09
Add OPM v1.40.0 to dev env

yashvardhannanavati:
We are now using OPM v1.40.0 for OCP v4.16 index images. This commit adds the OPM version to the dev env to facilitate local testing.

resolves:CLOUDDST-22689

### DIFF
--- a/iib/workers/config.py
+++ b/iib/workers/config.py
@@ -234,7 +234,7 @@ class DevelopmentConfig(Config):
         "v4.12": "opm-v1.26.4",
         "v4.13": "opm-v1.26.4",
         "v4.14": "opm-v1.26.4",
-        "v4.15": "opm-v1.28.0",
+        "v4.15": "opm-v1.26.4",
         "v4.16": "opm-v1.40.0",
     }
 
--- a/iib/workers/config.py
+++ b/iib/workers/config.py
@@ -235,7 +235,7 @@ class DevelopmentConfig(Config):
         "v4.13": "opm-v1.26.4",
         "v4.14": "opm-v1.26.4",
         "v4.15": "opm-v1.28.0",
-        "v4.16": "opm-v1.28.0",
+        "v4.16": "opm-v1.40.0",
     }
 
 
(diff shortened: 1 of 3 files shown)


chandwanitulsi:
Could you please update the mapping in config as well?

yashvardhannanavati:
> Could you please update the mapping in config as well?

nice catch! Added

lipoja (dismissed):

iib/workers/config.py
```diff
@@ -231,7 +231,7 @@ class DevelopmentConfig(Config):
         "v4.13": "opm-v1.26.4",
         "v4.14": "opm-v1.26.4",
         "v4.15": "opm-v1.28.0",
-        "v4.16": "opm-v1.28.0",
+        "v4.16": "opm-v1.40.0",
```
chandwanitulsi, line 234: Earlier 1.28 was added for 4.15 because of very early discussions, but now we know the right mapping we intend to use in our application.
Could you please remove v1.28.0 and keep only 1.26.4 and 1.40.0 to keep everything consistent

lipoja: We are blocked - opm-v1.40.0 binds to port 6060 so we can't run multiple opms at one machine

xDaile: Resolved by [Add atomic port locking mechanism](https://github.com/release-engineering/iib/pull/669)

xDaile (approved):

iib/workers/config.py
xDaile: @yashvardhannanavati could you address the @chandwanitulsi comment about removing 1.28.0, so we can release this in 8.7.5? I want to confirm that the [add attomic port locking mechanism PR](https://github.com/release-engineering/iib/pull/669) is working correctly, so I would like to release those two PRs at once.

yashvardhannanavati: Done :)

chandwanitulsi (approved):

xDaile (approved):

lipoja (approved):

PR #680, opened 2024-05-30
Release s3 resources in a timely manner

yashvardhannanavati:
This commit ensures the entire log file is not loaded in-memory and is directly streamed from the s3 bucket

### DIFF
--- a/iib/web/s3_utils.py
+++ b/iib/web/s3_utils.py
@@ -24,10 +24,12 @@ def get_object_from_s3_bucket(
     """
     file_name = f'{s3_key_prefix}/{s3_file_name}'
     log.info('getting file from s3 : %s', file_name)
-    s3 = boto3.resource(service_name='s3')
     try:
-        response = s3.meta.client.get_object(Bucket=bucket_name, Key=file_name)
+        s3_client = boto3.client('s3')
+        response = s3_client.get_object(Bucket=bucket_name, Key=file_name)
+        return response['Body'].read()
     except Exception as error:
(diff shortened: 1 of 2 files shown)


iib/web/s3_utils.py
```diff
@@ -24,10 +24,12 @@ def get_object_from_s3_bucket(
     """
     file_name = f'{s3_key_prefix}/{s3_file_name}'
     log.info('getting file from s3 : %s', file_name)
-    s3 = boto3.resource(service_name='s3')
     try:
-        response = s3.meta.client.get_object(Bucket=bucket_name, Key=file_name)
+        s3_client = boto3.client('s3')
+        response = s3_client.get_object(Bucket=bucket_name, Key=file_name)
+        return response['Body'].read()
     except Exception as error:
```
zanssa, line 31: would not it be better to log specific s3 error? I see that you are using botocore.exceptions.ClientError in the test

yashvardhannanavati: I don't exactly remember the reason behind using the generic exception when the code was added but if I were to guess, I believe we can encounter more than just ClientError when talking to S3 and probably wanted to propagate the error from S3 as is to IIB.

chandwanitulsi (approved):

lipoja (approved):

iib/web/s3_utils.py
lipoja: Are we ready to resolve this conversation and merge this PR?

yashvardhannanavati: yes, had a conversation with Zainab offline

PR #685, opened 2024-06-06
verify_operators_exists returns only unique packages

lipoja:
It appears that are multiple `packageName` with same name present in the result of `'api.Registry/ListBundles'`.
This caused to add every occurrence in the list of packages in index.db. This list was then passed as argument to `opm registry rm` command which caused failure because same package wanted to be removed more then once.

This is fixed by returning only unique packages names.

[CLOUDDST-23175]

### DIFF
--- a/iib/workers/tasks/opm_operations.py
+++ b/iib/workers/tasks/opm_operations.py
@@ -4,7 +4,7 @@ import random
 import shutil
 import subprocess
 import time
-from typing import List, Optional, Tuple, Generator, Union
+from typing import List, Optional, Tuple, Generator, Union, Set
 
 from tenacity import (
     before_sleep_log,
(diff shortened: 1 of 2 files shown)


iib/workers/tasks/opm_operations.py
```diff
@@ -900,14 +900,14 @@ def verify_operators_exists(
...
-    :rtype: (list, str)
+    :rtype: (set, str)
     """
     from iib.workers.tasks.build import terminate_process, get_bundle_json
     from iib.workers.tasks.iib_static_types import BundleImage
     from iib.workers.tasks.utils import run_cmd
     from iib.workers.tasks.utils import set_registry_token
 
-    packages_in_index: List[str] = []
+    packages_in_index: Set[str] = set()
```
yashvardhannanavati, line 910: it would be nice to add a unit test for this

lipoja: unit tests were added, please check :) thank you!

yashvardhannanavati (approved):

PR #689, opened 2024-06-13
Fix bug introduced by e1393a2 commit

xDaile:
By merging [Add  atomic port locking mechanism](https://github.com/release-engineering/iib/commit/e1393a259769e54d2348f3dba5e7ce39f6943575) we introduced a few bugs. 

- incorrect close of the file
- incorrect checking of the OPM version
- inconsistency between config variable and opm_registry_serve and opm_serve parameter opm_pprof_port/pprof_port.

This PR aims to fix the mentioned bugs.

### DIFF
--- a/iib/workers/tasks/opm_operations.py
+++ b/iib/workers/tasks/opm_operations.py
@@ -2,6 +2,7 @@ from functools import wraps
 import logging
 import os
 import random
+import re
 import shutil
 import socket
 import subprocess
@@ -78,7 +79,7 @@ class PortFileLock:
             # test if port is free
             s.bind(("localhost", self.port))
             # create file-lock
-            f = os.open(self.filename, os.O_CREAT | os.O_EXCL | os.O_RDWR)
+            f = os.open(self.filename, os.O_CREAT | os.O_EXCL)
             self.locked = True
             log.debug("Port %s used as %s was locked.", self.port, self.purpose)
         except FileExistsError:
@@ -92,7 +93,7 @@ class PortFileLock:
         finally:
             s.close()
             if f:
-                f.close()
+                os.close(f)
 
     def unlock(self):
         """Delete file representing port lock."""
@@ -143,7 +144,7 @@ def port_file_locks_generator(
     raise IIBError(err_msg)
 
 
-def get_opm_port_stacks() -> Tuple[List[List[int]], List[str]]:
+def get_opm_port_stacks(port_purposes: List[str]) -> Tuple[List[List[int]], List[str]]:
     """
     Get stack with port numbers and list of their intended purposes.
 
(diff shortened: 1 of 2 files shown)


lipoja (dismissed):

JAVGan (dismissed):

iib/workers/tasks/opm_operations.py
```diff
@@ -1369,3 +1370,22 @@ def set_opm_version(cls, from_index: Optional[str] = None):
         if index_version in opm_versions_config:
             Opm.opm_version = opm_versions_config.get(index_version)
         log.info("OPM version set to %s", Opm.opm_version)
+
+    @classmethod
+    def get_opm_version(cls):
```
chandwanitulsi, line 1375: so basically this function gives the actual version of the opm like 1.26.4 or 1.40, and Opm.opm_version is directly giving the opm bainry name, like opm-1.26.4

we validate the opm at _validate_multiple_opm_mapping() and at get_binary_versions(), so that should ensure opm-1.26.4 would result in valid opm.
I am thinking if running opm version again could create confusions in the log, and if getting the version value with a string operation on Opm.opm_version and then using Version() on that would be useful?

p.s Opm.opm_version and get_opm_version are sounding same with different values, can we use better names or atleast add a comment to note the difference

xDaile: I was thinking the same, running the `re.search()` on `OPM.opm_version`, however then I saw the value `opm` somewhere in the tests at this class attribute, which made me think running the command would be safer.

xDaile: I just renamed the function to get_opm_version_number

xDaile: I was also checking the mentioned methods, and I do not see, any checks that would ensure that the opm_version would not be something like mentioned `opm`.

chandwanitulsi: okay makes sense, one can actually install opm v1.26.4 under filename  opmv1264 and both the validations will pass.

iib/workers/tasks/opm_operations.py
```diff
@@ -1369,3 +1371,22 @@ def set_opm_version(cls, from_index: Optional[str] = None):
...
+
+    @classmethod
+    def get_opm_version_number(cls):
+        """
+        Get the opm version to be used for the entire IIB operation.
+
+        :return: currently set-up Opm version
+        :rtype: str
+        """
+        log.info("Determining the OPM version to use")
```
chandwanitulsi, line 1383:
```suggestion
        log.info("Determining the OPM version number")
```

xDaile: fixed, please check

iib/workers/tasks/opm_operations.py
```diff
@@ -184,60 +187,78 @@ def get_opm_port_stacks() -> Tuple[List[List[int]], List[str]]:
...
+        @wraps(func)
+        def inner(*args, **kwargs):
 
-        # Use the function to retrieve values from the generator
-        while not lock_success:
-            new_locks = next(gen)
-            currently_active_locks = []
+            # If we do not have any ports to lock
+            if len(port_purposes) == 0:
+                return func(*args, **kwargs)
```
xDaile, line 213: This part is new. The rest of the create_port_filelocks is just shifted by one Tab.

xDaile:
Example build:
[Add with pprof_wort](https://iib.qe.engineering.redhat.com/api/v1/builds/49330/logs)

iib/workers/tasks/opm_operations.py
xDaile: The reason for the introduction of this part is the `generate_cache_locally` method, which needs only `opm_pprof_port`, but only with opm version 1.40.0, so in case of opm version 1.26.4, we need to be able to bypass the `port_file_locks_generator` (which would throw an error, as it would get empty port_stack).

JAVGan (approved):

chandwanitulsi:
@xDaile , I am getting this error in add request when using registry-proxy.engineering.redhat.com/rh-osbs/iib-pub-pending:v4.16 as from_index
`The command "grpcurl -plaintext localhost:50062 api.Registry/ListBundles" failed with: ERROR:
  Code: Unknown
  Message: failed to open file for package "openshift-gitops-operator", channel "gitops-1.8", key "openshift-gitops-operator.v1.7.0": open /tmp/opm-serve-cache-4111982304/cache/openshift-gitops-operator_gitops-1.8_openshift-gitops-operator.v1.7.0.json: too many open files
`

JAVGan:
I did a test in my local and the PR is working fine

chandwanitulsi:
> @xDaile , I am getting this error in add request when using registry-proxy.engineering.redhat.com/rh-osbs/iib-pub-pending:v4.16 as from_index `The command "grpcurl -plaintext localhost:50062 api.Registry/ListBundles" failed with: ERROR: Code: Unknown Message: failed to open file for package "openshift-gitops-operator", channel "gitops-1.8", key "openshift-gitops-operator.v1.7.0": open /tmp/opm-serve-cache-4111982304/cache/openshift-gitops-operator_gitops-1.8_openshift-gitops-operator.v1.7.0.json: too many open files `

@xDaile since the PR is tested on local again, no objections from my end to merge it :) Thanks for your patience

PR #693, opened 2024-06-21
replace related images registry for stage bundles

chandwanitulsi:
Related to CLOUDDST-22108


iib/workers/tasks/build.py
```diff
@@ -687,6 +696,11 @@ def inspect_related_images(bundles: List[str], request_id) -> None:
             bundle_metadata = get_bundle_metadata(operator_manifest, False)
             for related_image in bundle_metadata['found_pullspecs']:
                 related_image_pull_spec = related_image.to_str()
+                if replace_related_image_registry:
+                    related_image_regsitry = related_image_pull_spec.split('/')[0]
```
JAVGan, line 700: Maybe it's a good idea to log that you're replacing the registry here?

iib/workers/config.py
```diff
@@ -237,6 +237,9 @@ class DevelopmentConfig(Config):
         "v4.15": "opm-v1.26.4",
         "v4.16": "opm-v1.40.0",
     }
+    iib_related_image_registry_replacement: dict = {
+        "registry.stage.redhat.io": "registry.stage.redhat.io"
```
yashvardhannanavati, line 241: Why do we need this mapping here? The default should be set to an empty dict. We can add values in the playbooks.

iib/workers/tasks/build.py
```diff
@@ -671,10 +671,19 @@ def inspect_related_images(bundles: List[str], request_id) -> None:
     :raises IIBError: if one of the bundles does not have the pullable related_image.
     """
     invalid_related_images = []
+    replace_related_image_registry = None
```
yashvardhannanavati, line 674: Is there a reason behind declaring it first here?

iib/workers/tasks/build.py
```diff
@@ -671,10 +671,19 @@ def inspect_related_images(bundles: List[str], request_id) -> None:
...
     for bundle in bundles:
         manifest_location = get_image_label(
             bundle, "operators.operatorframework.io.bundle.manifests.v1"
         )
+        bundle_registry = bundle.split('/')[0]
+        if (
```
yashvardhannanavati, line 680: This entire blob can be replaced with  

`replace_related_image_registry = worker_config.iib_related_image_registry_replacement.get(bundle_registry)`

yashvardhannanavati: In fact, it could be a single line like

`replace_related_image_registry = worker_config.iib_related_image_registry_replacement.get(bundle.split('/')[0])`

getting rid of line 679 too, no? I am open to keeping 679 for clarity, though

iib/workers/config.py
yashvardhannanavati: Also, shouldn't the mapping be `"registry.redhat.io": "registry.stage.redhat.io"`?

yashvardhannanavati (changes requested):

iib/workers/tasks/build.py
chandwanitulsi: Sure, I don't mind adding another log message.
The image pull spec is already mentioned in the skopeo inspect  log, so I thought it wouldn't be necessary

iib/workers/config.py
chandwanitulsi: I added it only in the DevelopmentConfig, so that if anyone needs to test with a stage bundle, they will have the config set in dev-env.

The mapping is IF a bundle is coming from "registry.stage.redhat.io", then use the "registry.stage.redhat.io" as the registry for related image pull spec as well. It is NOT just replace  registry.redhat.io with registry.stage.redhat.io.

iib/workers/tasks/build.py
chandwanitulsi: Yeah I created `bundle_registry` only for ease of read

If I directly use `replace_related_image_registry = worker_config.iib_related_image_registry_replacement.get(bundle_registry)` and if iib_related_image_registry_replacement config is not set, it fails with keyError, so we cannot have the config unset, it will always need a default.

iib/workers/config.py
```diff
@@ -237,6 +237,9 @@ class DevelopmentConfig(Config):
         "v4.15": "opm-v1.26.4",
         "v4.16": "opm-v1.40.0",
     }
+    iib_related_image_registry_replacement: dict = {
```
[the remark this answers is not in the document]
chandwanitulsi: Thanks for catching, I forgot to add the README file in the commit

iib/workers/tasks/build.py
chandwanitulsi: I recall it was failing on one case, but let me check again

iib/workers/tasks/build.py
yashvardhannanavati: yeah, and the default can be an empty dict, right?

iib/workers/config.py
yashvardhannanavati: > I added it only in the DevelopmentConfig, so that if anyone needs to test with a stage bundle, they will have the config set in dev-env.

I would rather prefer having a default empty dict in `Config` and then anyone running the dev-env can set it to whatever they want.

> The mapping is IF a bundle is coming from "registry.stage.redhat.io", then use the "registry.stage.redhat.io" as the registry for related image pull spec as well. It is NOT just replace registry.redhat.io with registry.stage.redhat.io.

I'm not sure this is the best way to do this. What happens if I want the related images to be checked but don't want the registries replaced? Maybe the config option could be {`user`:`registry.stage.redhat.io`}? Or something else?

chandwanitulsi: >I would rather prefer having a default empty dict in Config and then anyone running the dev-env can set it to whatever they want
okay, I will update.

>What happens if I want the related images to be checked but don't want the registries replaced?
Can you give an example? The case in jira was specific for stage pushes, so stage pushes always have bundle in registry.stage.redhat.io and for stage pushes we should only check for related image availability in registry.stage.redhat.io.
 Theoretically we can use the use config as `{"iib-stage@REDHAT.COM":"registry.stage.redhat.io}` but then instead of saying check for images in the same registry as bundle, we would explicitly say check for images in this registry for this specific user and conceptually the first option sounded better to me. It would also make a meaningful test in iib-api-test as there everything comes from iib-tester

iib/workers/tasks/build.py
chandwanitulsi: I went with default option of config not set, but I can change it to dict based on thread on config file

iib/workers/config.py
chandwanitulsi: I added the default dict in the Config but then I also wanted to understand when we chose to add the config only in Config and when do we add it in DevelopmentConfig as well.
Turns out for some configs, we need to have the default in Config but its better to set the in DevelopmentConfig as well like for iib_organization_customizations, iib_request_logs_dir, etc, so I went with same approach here. For now the changes are added without the change in dictionary mapping.

iib/workers/tasks/build.py
chandwanitulsi: Added the log message

iib/workers/tasks/build.py
chandwanitulsi: Updated

iib/workers/tasks/build.py
chandwanitulsi: Updated

iib/workers/config.py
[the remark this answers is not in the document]
chandwanitulsi: Updated

iib/workers/tasks/build.py
yashvardhannanavati: A log message might be more helpful for debugging purposes

iib/workers/config.py
yashvardhannanavati: > Can you give an example? The case in jira was specific for stage pushes, so stage pushes always have bundle in registry.stage.redhat.io and for stage pushes we should only check for related image availability in registry.stage.redhat.io.

Given the open source nature of IIB, we cannot just be thinking of use cases specific to business logic. Keeping it like this means if I am a different user trying to have the related images check without the registry replacement, that use case wouldn't be valid. 

> Theoretically we can use the use config as {user":"registry.stage.redhat.io} but then instead of saying check for images in the same registry as bundle, we would explicitly say check for images in this registry for this specific user and conceptually the first option sounded better to me. It would also make a meaningful test in iib-api-test as there everything comes from iib-tester

I'm not saying we use {user":"registry.stage.redhat.io}. It could be something else, maybe {"user": {"registry1": "registry2"}} or something completely different. I just want the use case I mentioned above to be covered since imposing some config change for everyone just because a particular user wants it sounds wrong. 

Also, if we say "check for images in the same registry as bundle" that becomes very specific and constrained. I think there's room to make the feature more generic. Like the bundle could be "registry.foo.io" while its relatedImages could be "registry.bar.io" and "quay.io". If we go with the nested dict approach, we can allow for multiple registry replacements. 


>only in Config and when do we add it in DevelopmentConfig as well.
Turns out for some configs, we need to have the default in Config but its better to set the in DevelopmentConfig as well like for iib_organization_customizations, iib_request_logs_dir, etc, so I went with same approach here. For now the changes are added without the change in dictionary mapping.

That's a good question. I don't remember the reason behind `iib_organization_customizations` being added to Dev Config. The variables in `Config` get inherited to all other classes, so setting a variable value there would mean the user doesn't have to provide a custom value if they're happy with using the default or don't have any use for that config variable. I'm ok with setting the variable in `Development Config` but as long as it doesn't have a value specific to RH.

iib/workers/config.py
```diff
@@ -71,6 +71,7 @@ class Config(object):
         "ppc64le": "ppc64le",
     }
     iib_default_opm: str = 'opm'
+    iib_related_image_registry_replacement: Optional[dict] = {}
```
xDaile, line 74: Why do we not use `Optional[Dict[str, str]]` instead of the `Optional[dict]`?

xDaile (approved):

JAVGan (approved):

PR #695, opened 2024-06-25
Fix OPM port locking mechanism bug

xDaile:
Related issue: CLOUDDST-23409

Several fixed bugs with the port locking mechanism:
- `Port_file_locks_generator` was changed to class. The reason is, that this `port_file_locks_generator` was throwing an exception (in the right way), however it was very difficult to debug, how and when it was called.
- Added log messages to better understand the flow of the process
- in the `create_port_filelocks`, we create a copy of `port_purposes` as previously we were working with data-object created when we used `\@create_port_filelocks` decorator, however, this became a problem once we removed an item from the given array. The next call of this decorated function used the already edited value, which caused problems.
- we check the `create_port_filelocks` in the ` create_port_filelocks` even after `get_opm_port_stacks` call, as there could be a change, that the `port_purposes_updated` would be empty, so the `PortFileLockGenerator` would throw an error like "No free port has been found after 0 attempts"

### DIFF
--- a/iib/workers/tasks/opm_operations.py
+++ b/iib/workers/tasks/opm_operations.py
@@ -1,4 +1,5 @@
 from functools import wraps
+from copy import deepcopy
 import logging
 import os
 import random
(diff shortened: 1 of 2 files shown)


JAVGan (approved):

xDaile:
[Related API test run](https://master-jenkins-csb-distribution.apps.ocp-c1.prod.psi.redhat.com/job/iib-api-test/24/console)

PR #699, opened 2024-07-04
Update rm endpoint to remove fbc opted in operators

chandwanitulsi:
Earlier the code will not even reach to the point of removing packages from /configs because opm rm was failing on opted in operators, change rm endpoint such that both opted-in and non-opted-in operators can be removed


iib/workers/tasks/build.py
```diff
@@ -1049,35 +1051,46 @@ def handle_rm_request(
...
+            # if operator is not opted in, remove from db
+            operators_in_db, index_db_path = verify_operators_exists(
+                from_index=from_index,
+                base_dir=temp_dir,
+                operator_packages=operators,
+                overwrite_from_index_token=overwrite_from_index_token,
+            )
+
+            if operators_in_db:
+                fbc_dir, _ = opm_registry_rm_fbc(
```
chandwanitulsi, line 1079: I have assumed here migrate is only needed if we are making any changes in the index.db, but I am not fully sure about it. 
If there are any cases where we need to have migrate, would be happy to update this

iib/workers/tasks/build.py
```diff
@@ -1049,35 +1051,46 @@ def handle_rm_request(
...
                 operator_path = os.path.join(catalog_from_index, operator)
                 if os.path.exists(operator_path):
                     log.debug('Removing operator from from_index FBC %s', operator_path)
                     shutil.rmtree(operator_path)
-            # overwrite data in `catalog_from_index` by data from `catalog_from_db`
-            # this adds changes on not opted in operators to final
-            merge_catalogs_dirs(catalog_from_db, catalog_from_index)
+
+            # if operator is not opted in, remove from db
+            operators_in_db, index_db_path = verify_operators_exists(
```
lipoja, line 1071: Would not be easier to just add this check (`verify_operators_exists` and `if`) before `fbc_dir, _ = opm_registry_rm_fbc` line 1052 (before change). Then the rest is same, right?
We just need to skip removal from SQLite db when it is not present. 

Or maybe we can just add this `verify_operators_exists` inside `opm_registry_rm_fbc` ? 

What do you see as benefit of this way? I am not saying it is wrong. I was just thinking if we can avoid changing so many lines (since they are "tested" by production use).

chandwanitulsi: > Would not be easier to just add this check (verify_operators_exists and if) before fbc_dir, _ = opm_registry_rm_fbc line 1052 (before change). Then the rest is same, right?
I think I got stuck a merge_catalog step and we also need to generate dockerfile, so if operator is not db and opm_registry_rm_fbc is not run, we won't have dockerfile

> Or maybe we can just add this verify_operators_exists inside opm_registry_rm_fbc
I had 2 thoughts about it
1. I was trying to keep just opm operations in opm_rm function
2. In create_empty_index, we get list of operators by opm serve and then give the list to opm rm, so if we put verify in operators in opm_rm function, we would be duplicating steps, (for create we would have to give list of db operators only in opm_rm and then manually remove fbc-opted-in operators from config)

> avoid changing so many lines (since they are "tested" by production use).
Yup, I agree to this.

lipoja (approved):

xDaile (approved):

PR #707, opened 2024-07-26
Add add-deprecations API endpoint

yashvardhannanavati:
Refers to CLOUDDST-23446

### DIFF
--- a/iib/web/api_v1.py
+++ b/iib/web/api_v1.py
@@ -25,6 +25,7 @@ from iib.web.models import (
     Operator,
     Request,
     RequestAdd,
+    RequestAddDeprecations,
     RequestFbcOperations,
     RequestMergeIndexImage,
     RequestRecursiveRelatedBundles,
@@ -53,6 +54,7 @@ from iib.workers.tasks.build_merge_index_image import handle_merge_request
 from iib.workers.tasks.build_create_empty_index import handle_create_empty_index_request
 from iib.workers.tasks.general import failed_request_callback
 from iib.web.iib_static_types import (
+    AddDeprecationRequestPayload,
     AddRequestPayload,
     AddRmBatchPayload,
     CreateEmptyIndexPayload,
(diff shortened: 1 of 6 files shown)


yashvardhannanavati:
I haven't added the API documentation on purpose so we do not market the endpoint before it's implemented completely.

iib/web/migrations/versions/49d13af4b328_add_add_deprecations_api_endpoint.py
```diff
@@ -0,0 +1,162 @@
...
+
+def upgrade():
+    op.create_table(
+        'deprecation_schema',
+        sa.Column('id', sa.Integer(), nullable=False),
+        sa.Column('schema', sa.Text(), nullable=False),
+        sa.PrimaryKeyConstraint('id'),
+    )
+    with op.batch_alter_table('deprecation_schema', schema=None) as batch_op:
+        batch_op.create_index(batch_op.f('ix_deprecation_schema_schema'), ['schema'], unique=True)
```
xDaile, line 27: Is there a reason to name the index `ix_deprecation_schema_schema` instead of `ix_deprecation_schema`?

iib/web/migrations/versions/49d13af4b328_add_add_deprecations_api_endpoint.py
```diff
@@ -0,0 +1,162 @@
...
+        sa.Column('operator_id', sa.Integer(), autoincrement=False, nullable=False),
+        sa.ForeignKeyConstraint(
+            ['operator_id'],
+            ['operator.id'],
+        ),
+        sa.ForeignKeyConstraint(
+            ['request_add_deprecations_id'],
+            ['request_add_deprecations.id'],
+        ),
+        sa.PrimaryKeyConstraint('request_add_deprecations_id', 'operator_id'),
```
xDaile, line 122: Do we need this `PrimaryKeyConstraint` if we set ``

iib/web/api_v1.py
```diff
@@ -1321,3 +1323,27 @@ def fbc_operations() -> Tuple[flask.Response, int]:
...
+    )
+    if not isinstance(payload, dict):
+        raise ValidationError('The input data must be a JSON object')
+
+    request = RequestAddDeprecations.from_json(payload)
+    db.session.add(request)
+    db.session.commit()
+
+    flask.current_app.logger.debug('Successfully validated request %d', request.id)
+    return flask.jsonify({'msg': 'This API endpoint hasn not been implemented yet'}), 201
```
JAVGan, line 1349: Shouldn't we return 501 instead of 201 while the endpoint is not functional?

iib/web/models.py
```diff
@@ -2200,3 +2214,159 @@ def get_mutable_keys(self) -> Set[str]:
...
+    )
+
+    __mapper_args__ = {
+        'polymorphic_identity': RequestTypeMapping.__members__['add_deprecations'].value
+    }
+
+    @classmethod
+    def from_json(  # type: ignore[override] # noqa: F821
+        cls,
+        kwargs: AddDeprecationRequestPayload,
```
JAVGan, line 2320: IMO using `kwargs` as an argument instead of a list of kw arguments `**kwargs` is a bit weird... What about naming it `payload` or something simlilar?

iib/web/api_v1.py
yashvardhannanavati: yes! good eye..changed it and added a TODO in the test which @chandwanitulsi can fix once she implements the worker functionality

iib/web/models.py
JAVGan: Ok, agreed!

Marking it as resolved

iib/web/api_v1.py
JAVGan: Tks!

JAVGan (approved):

xDaile (approved):

lipoja (approved):

PR #711, opened 2024-08-01
create-emty-index - remove FBC opted in operators.

lipoja:
FBC opted in operator caused opm to fail when removing it because operator was not present in database. If this occurs we try to remove operators once again but we enable permissive mode which silents errors when operator is not present.

[CLOUDDST-23683]

### DIFF
--- a/iib/workers/tasks/opm_operations.py
+++ b/iib/workers/tasks/opm_operations.py
@@ -932,13 +932,17 @@ def opm_registry_add_fbc(
     )
 
 
-def _opm_registry_rm(index_db_path: str, operators: List[str], base_dir: str) -> None:
+def _opm_registry_rm(
+    index_db_path: str, operators: List[str], base_dir: str, permissive: bool = False
+) -> None:
     """
     Generate and run the opm command to remove operator package from index db provided.
 
     :param str index_db_path: path where the input index image is temporarily copied
     :param list operators: list of operator packages to be removed
     :param base_dir: the base directory to generate the database and index.Dockerfile in.
+    :param permissive: enables permissive mode for opm registry rm
+        WARNING: Do not enable permissive mode outside create-emtpy-index API.
     """
     from iib.workers.tasks.utils import run_cmd
 
@@ -951,6 +955,9 @@ def _opm_registry_rm(index_db_path: str, operators: List[str], base_dir: str) ->
         '--packages',
         ','.join(operators),
     ]
+    if permissive:
+        cmd.append('--permissive')
+
     run_cmd(cmd, {'cwd': base_dir}, exc_msg='Failed to remove operators from the index image')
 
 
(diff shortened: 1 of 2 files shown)


iib/workers/tasks/opm_operations.py
```diff
@@ -1034,7 +1040,16 @@ def opm_create_empty_fbc(
 
     # Remove all the operators from the index
     set_request_state(request_id, 'in_progress', 'Removing operators from index image')
-    _opm_registry_rm(index_db_path=index_db_path, operators=operators, base_dir=temp_dir)
+    try:
+        _opm_registry_rm(index_db_path=index_db_path, operators=operators, base_dir=temp_dir)
+    except IIBError as e:
+        if 'Failed to remove operators from the index image' in str(e):
```
yashvardhannanavati, line 1046: This will try with permissive mode enabled for almost any failure. Is there a way to check the error message slightly more specific to package not being present?

chandwanitulsi:
@lipoja, how about we only remove operators from db? Like instead of doing serve from /configs which gives all operators, we do serve from index.db, that will give us non-opted-in operators and we just pass those to opm rm command? Once we migrate, we can ensure /configs is empty and build the index image?

iib/workers/tasks/opm_operations.py
lipoja: I've added check for more specific message. 

Do you both (@chandwanitulsi , @yashvardhannanavati) think it is sufficient or should we go down the road that Tulsi was suggesting? Unfortunately it would meant a lot of changes in our code (it is not that easy as this solution).

iib/workers/tasks/opm_operations.py
```diff
@@ -1034,7 +1040,16 @@ def opm_create_empty_fbc(
...
+    try:
+        _opm_registry_rm(index_db_path=index_db_path, operators=operators, base_dir=temp_dir)
+    except IIBError as e:
+        if 'Error deleting packages from database' in str(e):
+            log.info('Enable permissive mode for opm registry rm', from_index)
+            _opm_registry_rm(
+                index_db_path=index_db_path, operators=operators, base_dir=temp_dir, permissive=True
+            )
+        else:
+            raise e
```
yashvardhannanavati, line 1052: This should probably be raise IIBError?

iib/workers/tasks/opm_operations.py
yashvardhannanavati: @lipoja I think that depends on what happens when we add `--permissive-mode`. Does that command also remove the operators in index.db i.e non-opted in operators? If yes, then I think we should enable permissive mode by default and that would be sufficient. 

If not, then what happens if the `rm` command fails and the operators list contains few opted-in and few non-opted in operators? If this is the case, it might be more appropriate to implement what Tulsi suggested.

iib/workers/tasks/opm_operations.py
lipoja: It is IIBError. It just propagates the exception which was received.

yashvardhannanavati: ah, the callback..makes sense!

iib/workers/tasks/opm_operations.py
lipoja: I've added the comment (warning) as agreed, please check if this is sufficient.

yashvardhannanavati (approved):

PR #713, opened 2024-08-27
add-deprecation worker implementation

chandwanitulsi:


iib/workers/tasks/build_add_deprecations.py
```diff
@@ -0,0 +1,198 @@
...
+    span_name="workers.tasks.build.handle_add_deprecations_request",
+    attributes=get_binary_versions(),
+)
+def handle_add_deprecations_request(
+    request_id: int,
+    operator_package: str,
+    deprecation_schema: str,
+    from_index: Optional[str] = None,
+    binary_image: Optional[str] = None,
+    overwrite_from_index: bool = False,
```
yashvardhannanavati, line 54: Should this be `overwrite_from_index: Optional[bool] = False`? If not, this should probably be moved up before all the Optional params begin

iib/workers/tasks/build_add_deprecations.py
```diff
@@ -0,0 +1,198 @@
...
+            temp_dir,
+            [operator_package],
+            overwrite_from_index_token,
+        )
+        if not operators_in_db:
+            log.debug(
+                '%s operator package is either not present in index or it is opted in fbc',
+                operator_package,
+            )
+            err_msg = f'Cannot add deprecations for {operator_package}'
```
yashvardhannanavati, line 111: Probably pass the same message as the debug log on Line 107 here? It's more descriptive

iib/workers/tasks/build_add_deprecations.py
```diff
@@ -0,0 +1,198 @@
...
+            log.error(err_msg)
```
yashvardhannanavati, line 112: Also, we probably don't need both `log.debug` and `log.error`. Simply logging the error should be good, no?

iib/workers/config.py
```diff
@@ -83,13 +83,16 @@ class Config(object):
...
+        'iib.workers.tasks.build_add_deprecations',
         'iib.workers.tasks.general',
     ]
     # Path to hidden location of SQLite database
     hidden_index_db_path: str = '/var/lib/iib/_hidden/do.not.edit.db'
     # path where catalog resides in fbc_fragment
     # might need to be changed, currently based on test fbc-fragment
     fbc_fragment_catalog_path: str = '/configs'
+    # path where operator deprecations will be stored
+    operator_deprecations_dir: str = 'operator-deprecations-content'
```
yashvardhannanavati, line 95: Could you please add this to README?

iib/workers/tasks/build_add_deprecations.py
```diff
@@ -0,0 +1,198 @@
...
+    from_index_configs_deprecations_dir = os.path.join(
+        from_index_configs_dir, conf['operator_deprecations_dir']
+    )
+    operator_dir = os.path.join(from_index_configs_deprecations_dir, operator_package)
+    if not os.path.exists(operator_dir):
+        os.makedirs(operator_dir)
+    operator_deprecations_file = os.path.join(operator_dir, f'{operator_package}.json')
+    set_request_state(request_id, 'in_progress', 'Adding deprecations to from_index')
+    with open(operator_deprecations_file, 'w') as output_file:
+        json.dump(json.loads(deprecation_schema), output_file)
```
yashvardhannanavati, line 184: You probably don't need `json.loads` here, we have already established it's a valid json when we took the input from the user.

iib/workers/tasks/build_add_deprecations.py
```diff
@@ -0,0 +1,198 @@
...
+        from_index_resolved,
+        add_or_rm=True,
+    )
+    _cleanup()
+    set_request_state(
+        request_id, 'complete', 'The deprecation schema was successfully added to the index image'
+    )
+
+
+def add_deprecations_to_index(
```
yashvardhannanavati, line 163: Could you please add new lines and a few comments to make this function more readable?

yashvardhannanavati (changes requested):
Few minor comments, otherwise looks great!

iib/web/api_v1.py
```diff
@@ -1345,5 +1348,60 @@ def add_deprecations() -> Tuple[flask.Response, int]:
     db.session.add(request)
     db.session.commit()
 
-    flask.current_app.logger.debug('Successfully validated request %d', request.id)
-    return flask.jsonify({'msg': 'This API endpoint hasn not been implemented yet'}), 501
+    messaging.send_message_for_state_change(request, new_batch_msg=True)
+
+    overwrite_from_index = payload.get('overwrite_from_index', False)
+    from_index_pull_spec = request.from_index.pull_specification if request.from_index else None
```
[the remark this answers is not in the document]
lipoja: Do you think there might be case where from_index is just empty string and validation will pass?
I think it is possible - however I did not test it yet.

iib/workers/tasks/build_add_deprecations.py
```diff
@@ -0,0 +1,198 @@
...
+    set_request_state(request_id, 'in_progress', 'Getting all deprecations present in index image')
+    from_index_configs_dir = get_catalog_dir(from_index=from_index_resolved, base_dir=temp_dir)
+    conf = get_worker_config()
+    from_index_configs_deprecations_dir = os.path.join(
+        from_index_configs_dir, conf['operator_deprecations_dir']
+    )
+    operator_dir = os.path.join(from_index_configs_deprecations_dir, operator_package)
+    if not os.path.exists(operator_dir):
+        os.makedirs(operator_dir)
+    operator_deprecations_file = os.path.join(operator_dir, f'{operator_package}.json')
```
lipoja, line 181: Do you think there might be a case where we get .yml on the ingput?

iib/web/api_v1.py
```diff
@@ -1345,5 +1348,60 @@ def add_deprecations() -> Tuple[flask.Response, int]:
...
+    :raise NotFound: if the request is not found or there are no deprecation_schema for the request
+    :raise ValidationError: if the request is of invalid type or is not completed yet
+    """
+    request = Request.query.get_or_404(request_id)
+    if request.type != RequestTypeMapping.add_deprecations.value:
+        raise ValidationError(
+            f'The request {request_id} is of type {request.type_name}. '
+            'This endpoint is only valid for requests of type add_deprecations.'
+        )
+
```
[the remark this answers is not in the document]
lipoja: But in this case we do not need to wait for the request to be completed, right?
We should store deprecation schema to database in time of accepting the request. 
I would be OK for keeping it available even the task is not in final state.

iib/workers/config.py
lipoja: And is this the default path correct? 
Right now it defaults to: `/config/operator-deprecations-content`

iib/web/api_v1.py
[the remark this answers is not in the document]
yashvardhannanavati: The base class should verify it when `cls.from_json` is called on line 2333 in models.py. Am I missing something?

iib/web/api_v1.py
```diff
@@ -1345,5 +1348,60 @@ def add_deprecations() -> Tuple[flask.Response, int]:
...
+    messaging.send_message_for_state_change(request, new_batch_msg=True)
+
+    overwrite_from_index = payload.get('overwrite_from_index', False)
+    from_index_pull_spec = request.from_index.pull_specification if request.from_index else None
+    celery_queue = _get_user_queue(
+        serial=overwrite_from_index, from_index_pull_spec=from_index_pull_spec
+    )
+
+    args = [
+        request.id,
```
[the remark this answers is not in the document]
yashvardhannanavati: Well, the function definition is..
```
def _get_safe_args(
    args: List[Any],
    payload: PayloadTypesUnion,
) -> List[Union[str, List[str], bool, Dict[str, str]]]:
```

So we can indeed order it whatever sequence we want. I marked it as nit. I am willing to look the other way on this :)

iib/web/api_v1.py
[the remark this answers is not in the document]
yashvardhannanavati: hm, fair point. We can do without adding this.

However, I am thinking of what happens if `deprecation_schema` validation fails for a request and it's never added to the database. Should we instead add a `try/except` on line 1401?

iib/web/api_v1.py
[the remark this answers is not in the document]
chandwanitulsi: I have made from_index an non-optional argument, so  if condition is not valid anymore..will put update in next patch

iib/web/api_v1.py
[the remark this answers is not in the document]
chandwanitulsi: Exactly, deprecation_schema can be accesses even if the request is not complete.
try/except sounds fair to me

iib/workers/tasks/build_add_deprecations.py
chandwanitulsi: Hmm, I thought we generally keep a different shorter message in IIBError, I can put the same message

iib/workers/config.py
chandwanitulsi: yup, operator_deprecations_dir is just the name of dir, the path will default to /config/operator-deprecations-content/

iib/workers/tasks/build_add_deprecations.py
chandwanitulsi: IIRC it was adding content as string and opm validate was failing without json.loads()

iib/workers/tasks/build_add_deprecations.py
chandwanitulsi: sure I can new lines and comments. Is there anything specific that was hard to go through?

iib/web/api_v1.py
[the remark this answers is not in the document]
yashvardhannanavati: Do you mean optional? Could you please share the reason behind it?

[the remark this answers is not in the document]
chandwanitulsi: from-index will be mandatory. While adding deprecation in the gitlab repo, product teams will have to speficy which ocp index version they want to target. IIB should not have to decide which from_index to be used to add-deprecations, so we should get that info from input and via Pub

iib/web/api_v1.py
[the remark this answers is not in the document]
chandwanitulsi: hmm, I followed the pattern of previous endpoints, kinda putting most important agruments first

iib/web/api_v1.py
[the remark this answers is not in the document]
chandwanitulsi: exactly, that if condition will be removed in next patch, so request.from_index.pull_specification should have a valid input

iib/web/api_v1.py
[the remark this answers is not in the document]
lipoja: ok, maybe I get it wrong. 

What I see is: _get_safe_args takes the args do copy it and returns list in same order with hidden secrets. 

But I was referring to: 
```
 handle_add_deprecations_request.apply_async(
            args=args, link_error=error_callback, argsrepr=repr(safe_args), queue=celery_queue
        )
```
Where args should be in same order as defined in the handler  function
```
def handle_add_deprecations_request(
    request_id: int,
    operator_package: str,
    deprecation_schema: str,
    from_index: Optional[str] = None,
    binary_image: Optional[str] = None,
    overwrite_from_index: bool = False,
    overwrite_from_index_token: Optional[str] = None,
    build_tags: Optional[Set[str]] = None,
    binary_image_config: Optional[Dict[str, Dict[str, str]]] = None,
) -> None:
```

And I think both `safe_args` and `args` should have the same order.

iib/workers/config.py
lipoja: @chandwanitulsi I still have some questions about this directory - should not this be set by full path (similar to variables above)? 
Is there any agreement with other teams that the path for deprecation must be in /configs folder? 
Would not be better to have this set the path as image LABEL as well?

lipoja: as we agreed on standup - lets update comment that this folder is subfolder of `fbc_fragment_catalog_path`
and maybe consider changing the name to start with something else then letter. For example "_" 
`_operator-deprecations-content`

yashvardhannanavati (changes requested):

iib/web/api_v1.py
```diff
@@ -1345,5 +1348,65 @@ def add_deprecations() -> Tuple[flask.Response, int]:
...
+
+    try:
+        deprecation_schema_for_request = (
+            DeprecationSchema.query.join(RequestAddDeprecationsDeprecationSchema)
+            .filter(
+                RequestAddDeprecationsDeprecationSchema.request_add_deprecations_id == request_id
+            )
+            .first()
+        )
+    except Exception as e:
```
yashvardhannanavati, line 1409: ooh! I missed this earlier. We probably shouldn't catch a generic exception. Also, shouldn't it raise `IIBError`?

yashvardhannanavati (changes requested):
Just one comment. Apologies, I missed it earlier. Otherwise, LGTM!

iib/web/api_v1.py
chandwanitulsi: Can you suggest which exception should be caught? I raised the same exception as I thought raising the same could help us debug the error

chandwanitulsi: updated in next patch

yashvardhannanavati (approved):

yashvardhannanavati:
@chandwanitulsi could you please rebase your branch?

PR #716, opened 2024-09-05
Remove OPM usage for Dockerfile creation

yashvardhannanavati:
Refers to CLOUDDST-23891


iib/workers/tasks/opm_operations.py
```diff
@@ -671,33 +670,35 @@ def opm_generate_dockerfile(
...
+                    CMD ["serve", "/configs", "--cache-dir=/tmp/cache"]
+
+                    # Copy declarative config root and cache into image
+                    ADD catalog /configs
+                    COPY --chown=1001:0 cache /tmp/cache
+
+                    # Set DC-specific label for the location of the DC root directory
+                    # in the image
+                    LABEL operators.operatorframework.io.index.configs.v1=/configs
+                """
```
xDaile, line 692: Why not add also the remaining three commands? (I am okay with the current state, if there is a reason why not to add those).

Something like:
`ADD database/index.db /var/lib/iib/_hidden/do.not.edit.db
LABEL com.redhat.index.delivery.version=\"{delivery_version}\"
LABEL com.redhat.index.delivery.distribution_scope=\"{distribution_scope}\"
`

lipoja: I had a similar idea so we can remove all the functions which modify Dockerfile and have just one that generates it all in one call. 
However if it is a lot of work we can have different jira for that.

yashvardhannanavati: yes! what Jan said. It'll be major restructuring. Let's have a follow-up Jira to do that

xDaile (approved):

lipoja (approved):

PR #727, opened 2024-09-21
Fix fetching s3 bucket data

yashvardhannanavati:


yashvardhannanavati:
https://github.com/release-engineering/iib/pull/680 changed the way IIB fetched AWS S3 bucket data but it also introduced a bug by returning the byte data instead of streaming body. This commit fixes it

lipoja (approved):

PR #732, opened 2024-10-07
Remove deprecation data in fbc-operations, if present

yashvardhannanavati:
When an operator opts into the FBC fragment workflow, it no longer requires deprecations to be specified in configs/. Deprecations can be handled in fragments.

Refers to CLOUDDST-23449

### DIFF
--- a/iib/workers/tasks/opm_operations.py
+++ b/iib/workers/tasks/opm_operations.py
@@ -1068,6 +1068,9 @@ def opm_registry_add_fbc_fragment(
     )
 
     if operators_in_db:
+        remove_operator_deprecations(
+            from_index_configs_dir=from_index_configs_dir, operators=operators_in_db
+        )
         log.info('Removing %s from %s index.db ', operators_in_db, from_index)
         _opm_registry_rm(index_db_path=index_db_path, operators=operators_in_db, base_dir=temp_dir)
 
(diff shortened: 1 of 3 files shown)


iib/workers/tasks/opm_operations.py
```diff
@@ -31,6 +31,7 @@
 )
 
 log = logging.getLogger(__name__)
+worker_config = get_worker_config()
```
lipoja, line 34: Do we need it as global variable?

We usually have it like this:
```
    conf = get_worker_config()
    return _serve_cmd_at_port(
        serve_cmd, cwd, port, conf['iib_grpc_max_tries'], conf['iib_grpc_init_wait_time']
    )
```

yashvardhannanavati: changed it. Also, moved the code to a function to avoid duplication across endpoints.

iib/workers/tasks/opm_operations.py
```diff
@@ -1068,6 +1069,18 @@ def opm_registry_add_fbc_fragment(
     )
 
     if operators_in_db:
+        for operator in operators_in_db:
+            log.info('Removing operator deprecations for %s, if present', operator)
```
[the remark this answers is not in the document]
yashvardhannanavati: Done :)

tests/test_workers/test_tasks/test_opm_operations.py
```diff
@@ -897,9 +923,17 @@ def test_opm_registry_add_fbc_fragment(
...
         assert mock_cpt.call_count == 2
+        # Assert deprecations were removed correctly
+        assert not os.path.exists(deprecation_file)
+        assert not operator_deprecation_dir.check()
+        assert deprecations_dir.check(dir=True)
     else:
         assert mock_cpt.call_count == 1
         assert mock_orr.call_count == 0
+        # Assert deprecations are still present
+        assert os.path.exists(deprecation_file)
```
chandwanitulsi, line 934: For the test, this is fine.. but in the practical situation if the package is not present in the db, we assume that it's not the first time run of an opt in package, and its deprecation file should have been already removed in a previous request.

Wondering if we should remove the deprecation file for each fbc request of the package if the file is present, irrespective of it was present in db or not, wdyt?

yashvardhannanavati: > Wondering if we should remove the deprecation file for each fbc request of the package if the file is present, irrespective of it was present in db or not, wdyt?

Could you please elaborate on this? If it's not the first request of an fbc package and the package is missing in the db, the deprecations shouldn't exist in the first place. If it's still present, it's a bug in the workflow somewhere which should be fixed rather than covering it.

yashvardhannanavati: Discussed on a call and changes have been made to make the test more clear.

chandwanitulsi (approved):

PR #736, opened 2024-11-06
Drop support of Python 3.8

lipoja:
[CLOUDDST-24653]


xDaile (approved):

tox.ini
```diff
@@ -11,18 +11,15 @@ labels =
...
 usedevelop = true
-# 3.8 is the current suppported version
-# 3.9 is the latest version available on RHEL8
+# 3.12 is the current supported version on RHEL 8.10
 basepython =
-    py38: python3.8
-    py39: python3.9
-    migrate-db: python3.8
-    pip-compile: python3.8
+    py312: python3.12
```
yashvardhannanavati, line 20: should this also have 3.13?

tox.ini
```diff
@@ -63,15 +60,8 @@ deps =
...
-[testenv:py38]
-description = Python 3.8 unit tests [Mandatory]
-commands =
-    {[testenv]pytest_command}
-deps = 
-    -rrequirements-test.txt
-
-[testenv:py39]
-description = Python 3.9 unit tests [Mandatory]
+[testenv:py312]
```
yashvardhannanavati, line 63: same here? 

Reason: So it's easy to drop 3.12 when the time comes and we can rely on 3.13 in the meanwhile

lipoja: I've added tests for 3.13, please check.

tox.ini
lipoja: I've added tests for 3.13, please check.

yashvardhannanavati (approved):

PR #738, opened 2024-11-19
Fix Python3.12 dev_env

xDaile:


lipoja (approved):

docker/Dockerfile-api
```diff
@@ -22,6 +22,8 @@ RUN dnf -y install \
     && dnf update -y \
     && dnf clean all
 RUN update-alternatives --set python3 $(which python3.12)
+RUN pip3 install --upgrade pip setuptools
```
yashvardhannanavati, line 25: Instead of using pip, could you instead use dnf ? `python3.12-setuptools`

yashvardhannanavati (changes requested):

lipoja (approved):

PR #767, opened 2025-01-14
Allow "binary_image" to receive "scratch"

JAVGan:
This commit changes IIB to accept string "scratch" as `binary_image` in order to support the binaryless build in the future.

With this change IIB accept string "scratch" as binary image:

- defined in IIB_BINARY_IMAGE_CONFIG
- passed by binary_image parameter

Refers to CLOUDDST-23094

### DIFF
--- a/iib/workers/tasks/build.py
+++ b/iib/workers/tasks/build.py
@@ -807,6 +807,15 @@ def handle_add_request(
         ),
     )
     from_index_resolved = prebuild_info['from_index_resolved']
+
+    # FIXME: To be removed when the binaryless image build support is implemented
+    is_binaryless = prebuild_info['binary_image'] == "scratch"
+    if is_binaryless:
+        _cleanup()
+        log.warning("IIB is not yet able to process binaryless images.")
+        set_request_state(request_id, 'failed', 'IIB is not yet able to process binaryless images')
+        return
+
     Opm.set_opm_version(from_index_resolved)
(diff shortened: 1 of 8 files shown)


JAVGan:

iib/workers/tasks/build.py
```diff
@@ -807,6 +807,17 @@ def handle_add_request(
...
     )
     from_index_resolved = prebuild_info['from_index_resolved']
+
+    # FIXME: To be removed when the binaryless image build support is implemented
+    is_binaryless = prebuild_info['binary_image'] == "scratch"
+    if is_binaryless:
+        _cleanup()
+        log.warning("IIB is not yet able to process binaryless images.")
+        set_request_state(
+            request_id, 'complete', 'IIB is not yet able to process binaryless images'
```
xDaile, line 817: Is 'complete' the correct state here? I would prefer 'failed'.

chandwanitulsi (commented):
Overall looks good, but I agree that request should not complete if we are failing for binaryless image in worker

iib/workers/tasks/build.py
JAVGan: Ok!

I'll patch it accordingly as Tulsi and you agree on that

JAVGan:
@chandwanitulsi @xDaile I've patched it, PTAL again

xDaile (approved):

PR #784, opened 2025-01-23
IIB uses opm render

lipoja:
We agreed to drop `opm serve` command and move to `opm render` instead. 
This PR adds opm render and removes unused parts of code with opm serve.

[CLOUDDST-25207]


iib/workers/tasks/opm_operations.py
```diff
@@ -295,196 +295,132 @@ def inner(*args, **kwargs):
...
+    opm_render_output = run_cmd(
+        cmd, {'cwd': base_dir}, exc_msg=f'Failed to list bundles on {input_data}'
+    )
+
+    if not opm_render_output:
+        log.info("There are no bundles in %s", input_data)
+        return []
+
+    log.debug("Parsing data from opm render")
+    opm_data = [json.loads(package) for package in re.split(r'(?<=})\n(?={)', opm_render_output)]
```
JAVGan, line 410: I believe this piece of code is duplicated from `get_list_packages`. Wouldn't make sense to have a private or lower level function just to do this and share the code between `get_list_packages` and `get_list_bundles`?

iib/workers/tasks/opm_operations.py
```diff
@@ -295,196 +295,132 @@ def inner(*args, **kwargs):
...
 
-    :param int opm_port: OPM port number obtained from create_port_filelock decorator
-    :param int opm_pprof_port: Pprof opm port number obtained from create_port_filelock decorator
-    :param str catalog_dir: path to file-based catalog directory that should be served.
-    :return: tuple containing port number of the running service and the running Popen object.
-    :rtype: (int, Popen)
-    """
-    log.info('Serving data from file-based catalog %s', catalog_dir)
+    olm_packages = [
+        json.loads(package) for package in re.split(r'(?<=})\n(?={)', opm_render_output)
```
JAVGan, line 326: Just out of curiosity: Any specific reason we're splitting the content between brackets and then loading into `json`? Couldn't we simply load the JSON list and parse its dictionaries?

lipoja: The parser does not work that way. Since the opm output is not valid JSON. We get something like this:
```
{
...
data
...
}
{
....
data2
...
}
```
The whole output is not a list with multiple dictionaries. Those are just dictionaries so JSON package can't parse it correctly. And the easiest way to do it is to split it.

JAVGan: Got it, thanks!

iib/workers/tasks/opm_operations.py
lipoja: Good idea, changed the code - please check.

iib/workers/tasks/utils.py
```diff
@@ -37,7 +37,9 @@
 from iib.workers.config import get_worker_config
 from iib.workers.s3_utils import upload_file_to_s3_bucket
 from iib.workers.api_utils import set_request_state
-from iib.workers.tasks.opm_operations import opm_registry_serve, opm_serve_from_index
+from iib.workers.tasks.opm_operations import (
+    get_list_bundles,
+)
```
xDaile, line 42: Redundant space

xDaile:
Small comment, otherwise LGTM.

iib/workers/tasks/utils.py
xDaile, line 42: Is a one-line possible here?

iib/workers/tasks/opm_operations.py
```diff
@@ -487,6 +489,133 @@ def _serve_cmd_at_port(
     raise IIBError(f'Index registry has not been initialized after {max_tries} tries')
 
 
+def get_list_packages(
```
yashvardhannanavati, line 298: Consider renaming this to `get_operator_package_list` or something similar. `get_list_packages` sounds confusing to me

iib/workers/tasks/opm_operations.py
```diff
@@ -487,6 +489,133 @@ def _serve_cmd_at_port(
...
+    :param BundleImage olm_bundle: olm bundle dictionary
+    :return: OLM bundle version
+    :rtype: str
+    """
+    for property in olm_bundle['properties']:
+        if property['type'] == "olm.package":
+            return property['value']['version']
+
+    log.warning("No olm package version found for OLM bundle.")
+    return ""
```
yashvardhannanavati, line 335: Should this raise an error? I don't see the empty value being handled in the rest of the code.

iib/workers/tasks/opm_operations.py
xDaile: While doing the review, I was confused by this as well.

iib/workers/tasks/opm_operations.py
lipoja: Yeah, It might be better. I've updated the code.

iib/workers/tasks/opm_operations.py
lipoja: Changed - please review

iib/workers/tasks/opm_operations.py
```diff
@@ -10,7 +11,7 @@
 import tempfile
 import textwrap
 import time
-from typing import Callable, List, Optional, Set, Tuple, Union
+from typing import Callable, List, Optional, Set, Tuple, Union, Any
```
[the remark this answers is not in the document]
lipoja: I removed Any - it was not used

yashvardhannanavati (approved):
Didn't get a chance to test it locally. Code wise looks good!

xDaile (approved):

chandwanitulsi (commented):
Minor comments, otherwise looks good

JAVGan (dismissed):

iib/workers/tasks/opm_operations.py
```diff
@@ -295,196 +295,132 @@ def inner(*args, **kwargs):
     return decorator
 
 
-def opm_serve_from_index(base_dir: str, from_index: str) -> Tuple[int, subprocess.Popen]:
+def get_operator_package_list(
+    input_data: str,
```
chandwanitulsi, line 299: input_data seems very generic and doesn't show any meaning, how about we name it something like input_image_or_path

iib/workers/tasks/opm_operations.py
```diff
@@ -295,196 +295,132 @@ def inner(*args, **kwargs):
...
+    return package_names
 
 
-@create_port_filelocks(port_purposes=["opm_port", "opm_pprof_port"])
-def opm_serve(
-    opm_port: int,
-    catalog_dir: str,
-    opm_pprof_port: Optional[int] = None,
-) -> Tuple[int, subprocess.Popen]:
+def _get_olm_package_version(olm_bundle: dict) -> str:
```
chandwanitulsi, line 322: this seems more like bundle version than package version, can we rename this to_get_olm_bundle_version

iib/workers/tasks/opm_operations.py
```diff
@@ -295,196 +295,132 @@ def inner(*args, **kwargs):
...
+    log.warning(error_msg)
+    raise IIBError(error_msg)
 
 
-@create_port_filelocks(port_purposes=["opm_port"])
-def opm_registry_serve(
-    opm_port: int,
-    db_path: str,
-) -> Tuple[int, subprocess.Popen]:
+def _get_input_data_path(input_data: str, base_dir: str) -> str:
```
chandwanitulsi, line 339: same here, its confusing if input is image or location path, how about using something like input_image_or_path /input_image_or_dir?

iib/workers/tasks/opm_operations.py
```diff
@@ -295,196 +295,132 @@ def inner(*args, **kwargs):
...
     """
-    Locally start OPM registry service, which can be communicated with using gRPC queries.
+    Retrieve correct path to date which will be used when calling opm render.
```
chandwanitulsi, line 341: to date?

iib/workers/tasks/opm_operations.py
```diff
@@ -295,196 +295,132 @@ def inner(*args, **kwargs):
...
 
 
-def _serve_cmd_at_port_defaults(serve_cmd: List[str], cwd: str, port: int) -> subprocess.Popen:
+def get_list_bundles(
+    input_data: str,
+    base_dir: str,
+) -> List[BundleImage]:
     """
-    Call `_serve_cmd_at_port()` with default values from IIB config.
+    Run OPM render, which can be communicated with using gRPC queries.
```
chandwanitulsi, line 370: can you please explan how gRPC queries is relevant here? are we still using gPRC anywhere?

iib/workers/tasks/opm_operations.py
lipoja: data ;)

iib/workers/tasks/opm_operations.py
lipoja: Thank you for noticing.

chandwanitulsi (approved):

JAVGan (approved):

PR #808, opened 2025-02-04
CLOUDDST-24254 : Merge index image failing with FBC operator present …

ashwini3326:
This commit fix this bug by ignoring deprecation operators which are not available in database.
@release-engineering/exd-guild-hello-operator

### DIFF
--- a/iib/workers/tasks/build_merge_index_image.py
+++ b/iib/workers/tasks/build_merge_index_image.py
@@ -17,6 +17,7 @@ from iib.workers.tasks.opm_operations import (
     deprecate_bundles_fbc,
     opm_index_add,
     deprecate_bundles,
+    verify_operators_exists,
     Opm,
 )
 from packaging.version import Version
@@ -40,7 +41,6 @@ from iib.workers.tasks.fbc_utils import is_image_fbc
 from iib.workers.tasks.utils import (
     add_max_ocp_version_property,
     chmod_recursively,
-    filter_operators_present_in_db,
     get_bundles_from_deprecation_list,
     request_logger,
     set_registry_token,
@@ -326,32 +326,43 @@ def handle_merge_request(
         if missing_bundle_paths:
             add_max_ocp_version_property(missing_bundle_paths, temp_dir)
         set_request_state(request_id, 'in_progress', 'Deprecating bundles in the deprecation list')
+
         log.info('Deprecating bundles in the deprecation list')
         intermediate_bundles = missing_bundle_paths + source_index_bundles_pull_spec
         deprecation_bundles = get_bundles_from_deprecation_list(
             intermediate_bundles, deprecation_list
         )
+
         # We do not need to pass the invalid_version_bundles through the
         # get_bundles_from_deprecation_list function because we already know
         # they are present in the newly created index.
-        deprecation_bundles = deprecation_bundles + [
-            bundle['bundlePath'] for bundle in invalid_version_bundles
+
+        invalid_version_bundles_names = [
+            bundle["packageName"] for bundle in invalid_version_bundles
         ]
 
-        index_db_file = os.path.join(temp_dir, get_worker_config()['temp_index_db_path'])
-
-        # Operator passed in deprecation list should be available in operator database,
-        # filter_operators_present_in_db removes operaors which are passed in
-        # deprecation list and does not exists in database.
-        deprecation_bundles = filter_operators_present_in_db(
-            deprecation_bundles, index_db_file, temp_dir
+        intermediate_image_name = _get_external_arch_pull_spec(
+            request_id, arch, include_transport=False
         )
-
-        if deprecation_bundles:
-            intermediate_image_name = _get_external_arch_pull_spec(
-                request_id, arch, include_transport=False
+        if invalid_version_bundles_names:
+            filtered_invalid_version_bundles_names, _ = verify_operators_exists(
+                from_index=intermediate_image_name,
+                base_dir=temp_dir,
+                operator_packages=invalid_version_bundles_names,
+                overwrite_from_index_token=None,
             )
 
+            # Operator passed in deprecation list should be available in operator database,
+            invalid_version_bundles = [
+                bundle
+                for bundle in invalid_version_bundles
+                if bundle["packageName"] in list(filtered_invalid_version_bundles_names)
+            ]
+            deprecation_bundles = deprecation_bundles + [
+                bundle['bundlePath'] for bundle in invalid_version_bundles
+            ]
+
+        if deprecation_bundles:
             # we can check if source index is FBC or not because intermediate_image
             # will be always the same type because it is built
             # from source index image in _add_bundles_missing_in_source()
--- a/iib/workers/tasks/build_merge_index_image.py
+++ b/iib/workers/tasks/build_merge_index_image.py
@@ -40,6 +40,7 @@ from iib.workers.tasks.fbc_utils import is_image_fbc
 from iib.workers.tasks.utils import (
     add_max_ocp_version_property,
     chmod_recursively,
+    filter_operators_present_in_db,
     get_bundles_from_deprecation_list,
     request_logger,
     set_registry_token,
@@ -336,6 +337,15 @@ def handle_merge_request(
         deprecation_bundles = deprecation_bundles + [
             bundle['bundlePath'] for bundle in invalid_version_bundles
         ]
+
+        index_db_file = os.path.join(temp_dir, get_worker_config()['temp_index_db_path'])
+
+        # Operator passed in deprecation list should be available in operator database,
+        # filter_operators_present_in_db removes operaors which are passed in
+        # deprecation list and does not exists in database.
+        deprecation_bundles = filter_operators_present_in_db(
+            deprecation_bundles, index_db_file, temp_dir
+        )
 
         if deprecation_bundles:
             intermediate_image_name = _get_external_arch_pull_spec(
--- a/iib/workers/tasks/utils.py
+++ b/iib/workers/tasks/utils.py
@@ -1265,29 +1265,3 @@ def get_bundle_metadata(
         for pullspec in operator_csv.get_pullspecs():
             bundle_metadata['found_pullspecs'].add(pullspec)
     return bundle_metadata
-
-
-def filter_operators_present_in_db(
-    deprecation_bundles: List[str],
-    index_db_file: str,
-    temp_dir: str,
-) -> List[str]:
-    """
-    Check if operator marked for deprecation is available, if not drop them.
-
-    :param List[str] deprecation_bundles: List of bundles to be depricated
-    :param str index_db_file: Path of operator database
-    :param str temp_dir: temp directory where opm will be executed
-
-    :return:rtype: list(str) A List of operator bundle to be deprecated
-    """
-    bundles_list = get_list_bundles(index_db_file, temp_dir)
-
-    deprecation_bundles = [
-        bundle["bundlePath"]
-        for bundle in bundles_list
-        if bundle["bundlePath"] in deprecation_bundles
-    ]
-
-    log.info("Deprecation bundle after filter is : %s", deprecation_bundles)
-    return deprecation_bundles
--- a/iib/workers/tasks/utils.py
+++ b/iib/workers/tasks/utils.py
@@ -1265,3 +1265,29 @@ def get_bundle_metadata(
         for pullspec in operator_csv.get_pullspecs():
             bundle_metadata['found_pullspecs'].add(pullspec)
     return bundle_metadata
+
+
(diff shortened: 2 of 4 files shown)


iib/workers/tasks/utils.py
```diff
@@ -1293,3 +1293,41 @@ def get_bundle_metadata(
...
+    Check if operator marked for deprecation is available in operator db, if not drop them.
+
+    :param List[str] deprecation_bundles: List of bundles to be depricated
+    :param str index_db_file: Path of operator database
+
+    :return: A List of operator bundle to be deprecated
+    :raises:
+    """
+    try:
+        with sqlite3.connect(index_db_file) as conn:
```
lipoja, line 1311: By this line you are introducing new dependency - `sqlite3`.
I think, we should use `opm` to get relevant data from index image DB.
You can use `get_list_bundles` function - you do no need to implement anything new.

Please rebase and check the code if it is possible to use `get_list_bundles`

lipoja (changes requested):

iib/workers/tasks/utils.py
```diff
@@ -1265,3 +1265,31 @@ def get_bundle_metadata(
         for pullspec in operator_csv.get_pullspecs():
             bundle_metadata['found_pullspecs'].add(pullspec)
     return bundle_metadata
+
+
+def filter_deprecation_operator_from_db(
```
lipoja, line 1270: I am thinking about this name - if I did not know more I would think that the function will filter out the operators from DB (that it will remove them from DB) but we want to get only list of operators present in DB.

 what about something like `get_operators_present_in_db` or `filter_operators_present_in_db`   or just `filter_operators` ? 
@chandwanitulsi is good with naming :) maybe she could suggest something.

iib/workers/tasks/utils.py
```diff
@@ -1265,3 +1265,31 @@ def get_bundle_metadata(
...
+) -> List[str]:
+    """
+    Check if operator marked for deprecation is available, if not drop them.
+
+    :param List[str] deprecation_bundles: List of bundles to be depricated
+    :param str index_db_file: Path of operator database
+    :param str temp_dir: temp directory where opm will be executed
+
+    :return: A List of operator bundle to be deprecated
+    """
```
lipoja, line 1283: missing `:rtype: list(str)`

iib/workers/tasks/utils.py
```diff
@@ -1265,3 +1265,31 @@ def get_bundle_metadata(
...
+
+    available_bundlepaths = []
+    for bundle in bundles_list:
+        available_bundlepaths.append(bundle["bundlePath"])
+
+    deprecation_bundles = list(
+        filter(lambda bundlepath: bundlepath in available_bundlepaths, deprecation_bundles)
+    )
+
+    log.info(f"Deprecation bundle after filter is : {deprecation_bundles}")
```
lipoja, line 1294: For logging we should not use f-string formatting but `%s` notation because logging lib is optimized for it.
https://docs.python.org/3/howto/logging-cookbook.html#formatting-styles

iib/workers/tasks/utils.py
```diff
@@ -1265,3 +1265,31 @@ def get_bundle_metadata(
...
+    :param List[str] deprecation_bundles: List of bundles to be depricated
+    :param str index_db_file: Path of operator database
+    :param str temp_dir: temp directory where opm will be executed
+
+    :return: A List of operator bundle to be deprecated
+    """
+    bundles_list = get_list_bundles(index_db_file, temp_dir)
+
+    available_bundlepaths = []
+    for bundle in bundles_list:
```
lipoja, line 1287: Can we do the filtering in one cycle? Something like:

```python
available_bundlepaths = [
bundle["bundlePath"]  for bundle in bundles_list if bundle["bundlePath"] in deprecation_bundles
]

```

then just log and return available_bundlepaths

tests/test_workers/test_tasks/test_utils.py
```diff
@@ -1264,3 +1265,61 @@ def test_add_max_ocp_version_property_empty_index(mock_apti, mock_glb, mock_gbj,
...
+
+    operator_after_deprication = utils.filter_deprecation_operator_from_db(
+        deprecation_bundles, "index/db/file", "temp_dir"
+    )
+
+    assert operator_after_deprication == ["op1@sha"]
+
+
+@mock.patch('iib.workers.tasks.utils.get_list_bundles')
+def test_no_matching_filter_deprecation_operator_from_db_(mock_bundle_list):
```
lipoja, line 1300: extra `_` at the end of name `test_no_matching_filter_deprecation_operator_from_db_`

tests/test_workers/test_tasks/test_utils.py
```diff
@@ -1264,3 +1265,61 @@ def test_add_max_ocp_version_property_empty_index(mock_apti, mock_glb, mock_gbj,
...
+        {
+            'bundlePath': 'op2@sha',
+            'csvName': 'op2.v0.1.501-1',
+            'packageName': 'op2',
+            'version': '0.1.501-1',
+        },
+    ]
+
+    deprecation_bundles = [
+        "op1@sha",
```
lipoja, line 1289: Do you think you could add `"op3@sha",` here as well? 
The result should be the same

tests/test_workers/test_tasks/test_utils.py
```diff
@@ -1264,3 +1265,61 @@ def test_add_max_ocp_version_property_empty_index(mock_apti, mock_glb, mock_gbj,
...
+        {
+            'bundlePath': 'op2@sha',
+            'csvName': 'op2.v0.1.501-1',
+            'packageName': 'op2',
+            'version': '0.1.501-1',
+        },
+    ]
+
+    deprecation_bundles = [
+        "op3@sha",
```
lipoja, line 1318: Do we need this tests if we include this in the previous one?

lipoja (changes requested):

tests/test_workers/test_tasks/test_utils.py
ashwini3326: It makes unittest more specific, so I thought of not testing both cases in single unitest.

tests/test_workers/test_tasks/test_utils.py
ashwini3326: removed, will reflect in next push.

iib/workers/tasks/utils.py
ashwini3326: changed,  will reflect in next push.

iib/workers/tasks/utils.py
ashwini3326: fixed, will reflect in next push.

iib/workers/tasks/utils.py
ashwini3326: going with filter_operators_present_in_db

iib/workers/tasks/build_merge_index_image.py
```diff
@@ -326,6 +327,15 @@ def handle_merge_request(
             bundle['bundlePath'] for bundle in invalid_version_bundles
         ]
 
+        index_db_file = os.path.join(temp_dir, get_worker_config()['temp_index_db_path'])
+
+        # Operator passed in deprecation list should be available in operator database,
+        # filter_operators_present_in_db removes operaors which are passed in
+        # deprecation list and does not exists in database.
+        deprecation_bundles = filter_operators_present_in_db(
+            deprecation_bundles, index_db_file, temp_dir
```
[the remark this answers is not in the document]
ashwini3326: In the process of bundle deprecation we are fetching list of such bundles from two sources, one is from a "deprecation_list"  and second is from "invalid_version_bundles" .  As this "invalid_version_bundles" can come from target index too which won't be available in source index db, in this case deprecating from source index will fail, so using _get_present_bundles() will not resolve this issue ( as per what I have seen in codes). @lipoja could you share your thoughts.

ashwini3326:
@lipoja @chandwanitulsi  @xDaile  PTAL

iib/workers/tasks/build_merge_index_image.py
```diff
@@ -316,12 +317,32 @@ def handle_merge_request(
...
         # We do not need to pass the invalid_version_bundles through the
         # get_bundles_from_deprecation_list function because we already know
         # they are present in the newly created index.
+
+        invalid_version_bundles_names = [
+            bundle["packageName"] for bundle in invalid_version_bundles
+        ]
+
+        filtered_invalid_version_bundles_names, _ = verify_operators_exists(
+            from_index=source_from_index_resolved,
```
lipoja, line 334: Maybe I told it to you wrongly. Now when I look to the code more deeply I am thinning about checking `intermediate_image_name` instead of `source_from_index_resolved`.
Because we run deprecate function on intermediate_image_name and this image should already contain data from source+target images.

iib/workers/tasks/build_merge_index_image.py
```diff
@@ -316,21 +317,40 @@ def handle_merge_request(
...
         # they are present in the newly created index.
+
+        invalid_version_bundles_names = [
+            bundle["packageName"] for bundle in invalid_version_bundles
+        ]
+
+        intermediate_image_name = _get_external_arch_pull_spec(
+            request_id, arch, include_transport=False
+        )
+        filtered_invalid_version_bundles_names, _ = verify_operators_exists(
```
lipoja, line 336: I would suggest to add `if` statement here. 
```python
if invalid_version_bundles_names:
     filtered_invalid_version_bundles_names, _ = verify_operators_exists(
          from_index=intermediate_image_name,
          base_dir=temp_dir,
          operator_packages=invalid_version_bundles_names,
      )

      # Operator passed in deprecation list should be available in operator database,
      invalid_version_bundles = [
          bundle
          for bundle in invalid_version_bundles
          if bundle["packageName"] in list(filtered_invalid_version_bundles_names)
      ]

else:
   invalid_version_bundles = []
```
This way we do not call `verify_operators_exists` for empty list.

lipoja (changes requested):

iib/workers/tasks/build_merge_index_image.py
ashwini3326: @lipoja PTAL, the else statement is not required cause in that case we are not using invalid_version_bundles anywhere ahed.

lipoja (dismissed):

yashvardhannanavati (dismissed):

iib/workers/tasks/build_merge_index_image.py
```diff
@@ -314,23 +315,43 @@ def handle_merge_request(
...
 
+            # Operator passed in deprecation list should be available in operator database,
+            invalid_version_bundles = [
+                bundle
+                for bundle in invalid_version_bundles
+                if bundle["packageName"] in list(filtered_invalid_version_bundles_names)
+            ]
+            deprecation_bundles = deprecation_bundles + [
+                bundle['bundlePath'] for bundle in invalid_version_bundles
+            ]
```
xDaile, line 352: Why not assign `bundle['bundlePath']` right away to the variable `invalid_version_bundles`? (variable name, can be changed, I aim for the usage of the variable). That way it would look like:

```
            invalid_version_bundle_paths = [
                bundle['bundlePath']
                for bundle in invalid_version_bundles
                if bundle["packageName"] in list(filtered_invalid_version_bundles_names)
            ]
            deprecation_bundles = deprecation_bundles + invalid_version_bundle_paths
```

xDaile, line 352: Does the coverage show this in green? I would expect some test for the value of the deprecation_bundles/invalid_version_bundles.

xDaile (commented):
Small comments, otherwise LGTM

iib/workers/tasks/build_merge_index_image.py
ashwini3326: patched

xDaile: @lipoja @yashvardhannanavati @chandwanitulsi do we need a test for this `deprecation_bundles` content constructed with `invalid_version_bundles`? I would prefer to have this covered by the tests.

xDaile: Also, I checked and the coverage is showing this in green, but did not find a test, which would test particularly this part.

xDaile:
@ashwini3326 Do we need two commits here? Wouldn't one be enough?

ashwini3326:
> @ashwini3326 Do we need two commits here? Wouldn't one be enough?

can't we squash the commits before merge?

iib/workers/tasks/build_merge_index_image.py
ashwini3326: @xDaile PTAL, I have updated the tests to check  cases for different return from  verify_operators_exists functions.

JAVGan (approved):

PR #815, opened 2025-02-05
Bump opentelemetry packages

JAVGan:
Bump opentelemetry packages


JAVGan:
@chandwanitulsi @lipoja @xDaile @yashvardhannanavati PTAL.

I would like to include this in the upcoming release

lipoja (approved):

PR #845, opened 2025-02-19
WIP: Skip building multi-arch binary-less image

JAVGan:
This commit allows IIB to process a single build for binaryless images with a generic architecture labeled as "linux-amd64"

Refers to CLOUDDST-23383


JAVGan:
@chandwanitulsi @xDaile @lipoja @yashvardhannanavati PTAL

iib/workers/tasks/utils.py
```diff
@@ -1191,19 +1185,27 @@ def prepare_request_for_build(
 
     binary_image = build_request_config.binary_image(request_index, distribution_scope)
     if binary_image == "scratch":  # binaryless image mode
+        arches = set(["linux-amd64"])
```
lipoja, line 1188: Is the architecture correct, does it pass? 
I cant find any valid list of architectures, just this:


> Set the ARCH of the image to be built, and that of the base image to be pulled, if the build uses one, to the provided value instead of using the architecture of the host. (Examples: arm, arm64, 386, amd64, ppc64le, s390x)


https://github.com/containers/buildah/blob/main/docs/buildah-build.1.md

lipoja: OK, for our purposes I got this - we deliver just this list: 
```bash
podman manifest inspect registry.redhat.io/redhat/redhat-operator-index:v4.16 | grep architecture
                "architecture": "amd64",
                "architecture": "arm64",
                "architecture": "ppc64le",
                "architecture": "s390x",
```

Could not w use just the `amd64` instad of `linux-amd64` ?

lipoja (changes requested):

iib/workers/tasks/utils.py
```diff
@@ -1191,19 +1185,27 @@ def prepare_request_for_build(
...
+        raise IIBError('No arches were provided to build the index image')
+
     if not arches.issubset(binary_image_arches):
         raise IIBError(
             'The binary image is not available for the following arches: {}'.format(
                 ', '.join(sorted(arches - binary_image_arches))
             )
         )
 
+    arches_str = ', '.join(sorted(arches))
```
xDaile, line 1206: Why are we not returning this format (joined and sorted array) from the method `gather_index_image_arches` method? Also, this exception "No arches were provided to build the index image" is already thrown within the mentioned method, should we deduplicate that?

iib/workers/tasks/utils.py
JAVGan: Sure, I'll change it to `arm64` instead

JAVGan: Patched

chandwanitulsi (commented):
@JAVGan The code LGTM
While testing this change I noticed that if we build index image without binary image we see less labels for the index image.
I think its okay but I have reached out to stakeholders to confirm that..
Lets wait for the confirmation before merging this PR

iib/workers/tasks/utils.py
```diff
@@ -1191,19 +1185,27 @@ def prepare_request_for_build(
...
     if binary_image == "scratch":  # binaryless image mode
+        arches = set(["amd64"])
         binary_image_resolved = binary_image
         binary_image_arches = arches
     else:
+        arches = gather_index_image_arches(build_request_config, index_info)
         binary_image_resolved = get_resolved_image(binary_image)
         binary_image_arches = get_image_arches(binary_image_resolved)
 
+    if not arches:
```
lipoja, line 1196: This already duplicates the check in `get_image_arches` and when I look at the code I think it might not be necessary to have it here.

iib/workers/tasks/utils.py
lipoja: We use this joined format only for printing it out. We should not return it because we work with arches as set. This way you don't have to parse it from string when we work with arches later on.

lipoja (changes requested):

iib/workers/tasks/utils.py
JAVGan: Sorry, I don't get it..

The `get_image_arches` has only a check if the manifest is `V2` or not, if there are no arches it will return an empty set... Or am I missing something?

lipoja: Yeah, my bad - I meant `gather_index_image_arches` is raising the same. (as Michal mentioned belov)

JAVGan: Cool! I'll get rid of it then, tks!

JAVGan:
@lipoja PTAL again

lipoja (approved):

PR #873, opened 2025-03-04
CLOUDDST-26062 : Isolate temporary folders being used to fetch bundle from source and target index.

ashwini3326:
This merge request separates the temporary folders being used for fetching source and target index bundles. @lipoja @chandwanitulsi @xDaile @JAVGan @yashvardhannanavati  PTAL

### DIFF
--- a/iib/workers/tasks/build_merge_index_image.py
+++ b/iib/workers/tasks/build_merge_index_image.py
@@ -2,6 +2,7 @@
 import itertools
 import logging
 import os
+import shutil
 import stat
 import tempfile
 from typing import List, Optional, Tuple
@@ -279,9 +280,13 @@ def handle_merge_request(
         log.info('Getting bundles present in the source index image')
 
         with set_registry_token(overwrite_target_index_token, target_index, append=True):
+            source_bundles_dir = os.path.join(temp_dir, 'source_bundles')
+            os.makedirs(source_bundles_dir, exist_ok=True)
+
             source_index_bundles, source_index_bundles_pull_spec = _get_present_bundles(
-                source_from_index_resolved, temp_dir
+                source_from_index_resolved, source_bundles_dir
             )
+            shutil.rmtree(source_bundles_dir)
 
         target_index_bundles: List[BundleImage] = []
         if target_index:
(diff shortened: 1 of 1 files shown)


iib/workers/tasks/build_merge_index_image.py
```diff
@@ -279,8 +279,11 @@ def handle_merge_request(
         log.info('Getting bundles present in the source index image')
 
         with set_registry_token(overwrite_target_index_token, target_index, append=True):
+            source_bundles_dir = os.path.join(temp_dir, 'source_bundles')
```
lipoja, line 282: I am thinking, do we work with the database later on? 
Can we save some time/space when we keep this in directly in tmp folder? The think is we add bundles which are missing in source. So it means we would need to extract the source again? Was it previously done again - or  have we used the already extracted config?

I did not inspect the code that deeply yet. This is just an idea.

iib/workers/tasks/build_merge_index_image.py
```diff
@@ -289,7 +292,12 @@ def handle_merge_request(
             with set_registry_token(
                 overwrite_target_index_token, target_index_resolved, append=True
             ):
-                target_index_bundles, _ = _get_present_bundles(target_index_resolved, temp_dir)
+                target_bundles_dir = os.path.join(temp_dir, 'target_bundles')
+                os.makedirs(target_bundles_dir, exist_ok=True)
+
+                target_index_bundles, _ = _get_present_bundles(
```
lipoja, line 298: Will we work with the date in `target_bundles_dir` later on? If not I would suggest to remove this folder. So we keep our disk space consumption low.

ashwini3326: the whole temporary folder is removed after the operation completes, so this directory will automatically be removed.

lipoja: Yes, that is true. But on the other hand we had a lot of issue with disc space. Therefore I thought removing data when they are no longer needed is good idea.

JAVGan (approved):
It LGTM, please wait others reviews as well

lipoja (approved):

PR #938, opened 2025-03-28
Deprecate fbc bundles in chunks

xDaile:


iib/workers/config.py
```diff
@@ -44,6 +44,7 @@ class Config(object):
     iib_image_push_template: str = '{registry}/iib-build:{request_id}'
     iib_index_image_output_registry: Optional[str] = None
     iib_log_level: str = 'INFO'
+    iib_deprecate_bundles_limit = 200
```
lipoja, line 47: How many  bundles are usually deprecated?
How many bundles needs to be deprecated for community?

iib/workers/tasks/opm_operations.py
```diff
@@ -515,13 +515,20 @@ def deprecate_bundles_fbc(
...
-        index_db=index_db_file,
-        bundles=bundles,
-    )
+    # Break the bundles into chunks of at max iib_deprecate_bundles_limit bundles
+    for i in range(
+        0,
+        len(bundles),
+        conf.iib_deprecate_bundles_limit,
+    ):  # Determine position i in the bundles array
+        opm_registry_deprecatetruncate(
```
lipoja, line 527: Example: we split input to 4 chunks. What happens when last chunk fails? 
Will the `IIBError` from `run_cmd` terminate everything? 
I am asking this to verify that we do not end up in inconsistent deprecation (3 out of 4 chunks were deprecated and image was build with it)

tests/test_workers/test_tasks/test_opm_operations.py
```diff
@@ -680,11 +681,11 @@ def test_deprecate_bundles_fbc(
         binary_image="some:image",
         from_index=from_index,
     )
-
-    mock_ord.assert_called_once_with(base_dir=tmpdir, index_db=index_db_file, bundles=bundles)
+    if len(bundles) != 0:
```
lipoja, line 684: Nitpick: would not be simple `if bundles:`  enough?

lipoja (changes requested):

iib/workers/config.py
xDaile: For the community, I counted ~ 3200 in the logs. For the Prod v4.19, there were 45 deprecations.

iib/workers/tasks/opm_operations.py
xDaile: I suppose it would work as previously - the image would not be built if there would be a failure during `opm_registry_deprecate_truncate`.

iib/workers/config.py
xDaile: In most cases, we will not even come close to 200. The community index is very different from the other ones. In our case and setup, it would be safe to increase this limit to something like 500 (maybe even 1000), but the 200 will work with a smaller stack size, so it should prevent this problem in the future.

lipoja: The code looks OK. The question is if `opm registry deprecatetruncate` command has significant "loading" time (for example builds a cache). Because if we keep limit 200 then we increase the processing time by the "loading opm time" 16x. 

Example: I know `opm render` takes more then 1 minute. So it would be extra 16 minutes and I think it is significant amount. Maybe `opm registry deprecatetruncate`  is fast and we can merge this. 

Could you try it at post here what have you learned?

xDaile: The `opm deprecatetruncate` took at most ~4 seconds with around ~700 bundles for the deprecation.

PR #970, opened 2025-05-20
Add Index Catalog to GitLab push configuration

JAVGan:
This commit introduces a new API configuration named `IIB_INDEX_TO_GITLAB_PUSH_MAP`, which may contain a mapping between index images (keys) to push its catalogs into certain GitLab repositories (values).

The implementation of GitLab push will be added into a different PR.

Refers to CLOUDDST-26435

### DIFF
--- a/README.md
+++ b/README.md
@@ -199,6 +199,8 @@ The custom configuration options for the REST API are listed below:
   to another dictionary mapping ocp_version label to a binary image pull specification.
   This is useful in setting up customized binary image for different index image images thus
   reducing complexity for the end user. This defaults to `{}`.
+* `IIB_INDEX_TO_GITLAB_PUSH_MAP` - the mapping, `dict(<str>:<str>)`, to specify which index
+  images (keys) which should have its catalog pushed into a GitLab repository (value). This defaults to {}.
 * `IIB_GRAPH_MODE_INDEX_ALLOW_LIST` - the list of index image pull specs on which using the
   `graph_update_mode` parameter while submitting an IIB request is permitted. This defaults to `[]`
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -199,6 +199,8 @@ The custom configuration options for the REST API are listed below:
   to another dictionary mapping ocp_version label to a binary image pull specification.
   This is useful in setting up customized binary image for different index image images thus
   reducing complexity for the end user. This defaults to `{}`.
+* `IIB_INDEX_TO_GITLAB_PUSH_MAP` - the mapping, `dict(<str>:<str>)`, to specify which index
+  images (keys) which should have its catalog pushed into a GitLab repository (value).
```
yashvardhannanavati, line 203: We should add a line `This defaults to `{}`.` to be consistent with other descriptions :)

iib/web/api_v1.py
```diff
@@ -99,6 +99,7 @@ def _get_rm_args(
         request.distribution_scope,
         flask.current_app.config['IIB_BINARY_IMAGE_CONFIG'],
         payload.get('build_tags', []),
+        flask.current_app.config['IIB_INDEX_TO_GITLAB_PUSH_MAP'],
```
yashvardhannanavati, line 102: We probably also need to add this to `handle_fbc_operations`

yashvardhannanavati (changes requested):

README.md
JAVGan: I've updated it, PTAL again

iib/web/api_v1.py
JAVGan: I've added it to the `fbc_operations`, PTAL again

yashvardhannanavati (approved):

PR #1034, opened 2025-06-18
feat(CLOUDDST-27628): Load GitLab tokens to config

JAVGan:
This commit changes the IIB Web to load a map between GitLab repositories and the corresponding auth tokens into `IIB_INDEX_TO_GITLAB_TOKEN_MAP` by storing the JSON mapping in the enviroment variable with the very same name.

It then loads the `IIB_INDEX_TO_GITLAB_TOKEN_MAP` infor Flask config and sends it to the workers during the handle `add`, `rm` and `fbc` operations.

Refers to CLOUDDST-27628


JAVGan:
@yashvardhannanavati @nmars this is my WIP PR to load the Git auth tokens into IIB's API/worker.

Please let me know WDYT so I can properly adjust it.

Once everything is properly set and approved I'll patch the playbooks

JAVGan:
@yashvardhannanavati can you please let us know if this would be a good approach for the token introduction? Or would it be better to do like #1036 ?

PR #1061, opened 2025-07-23
CLOUDDST-28116: IIB skip processing of task in final state

ashwgit:
@release-engineering/exd-guild-hello-operator  These changes are expected to deny if someone is tries to update the final state of a build. PTAL


iib/workers/api_utils.py
```diff
@@ -168,6 +169,11 @@ def update_request(
             rv.status_code,
             rv.text,
         )
+        if rv.json().get("error") in [
```
nmars, line 172: Is it certain the "error" field will be in the JSON response body? (sorry this may be a newbie question)

iib/workers/api_utils.py
```diff
@@ -168,6 +169,11 @@ def update_request(
...
+            "A failed request cannot change states",
+            "A complete request cannot change states",
+        ]:
+            raise ValidationError(rv.json().get("error"))
```
yashvardhannanavati, line 176: Doing this will potentially trigger an infinite loop for `complete` requests.

So, ValidationError will trigger https://github.com/release-engineering/iib/blob/master/iib/workers/tasks/general.py#L20 which will try to set the state message to `An unknown error occurred. See logs for details ` by once again calling the `/patch` endpoint and trying to mark the request as `failed` even when it was `complete`. This will again raise the same error "A complete request cannot change states" and so on.

iib/workers/api_utils.py
ashwgit: yeah, that is expected for the case where response status in not 200 ( in this case -> rv.ok != True)

iib/workers/api_utils.py
ashwgit: I have implemented new changes to fix the callback issue. Thanks for pointing out. btw it does not triggers an infinite callback, but I have seen it to repeat it around 3 times. Fixing it in the new changes.

ashwgit: @yashvardhannanavati PTAL. btw, I will create a seperate MR for adding initial state change for rm request.

iib/workers/tasks/general.py
```diff
@@ -36,5 +36,9 @@ def failed_request_callback(
         msg = 'An unknown error occurred. See logs for details'
         log.error(msg, exc_info=exc)
 
+    if isinstance(exc, FinalStateOverwiteError):
+        log.info(f"Request {request_id} is in a final state,ignoring update.")
+        _cleanup()
+        return
```
xDaile, line 42: Still an log `An unknown error occurred. See logs for details` would be logged here. 
I would add elif under the first if.

iib/workers/tasks/general.py
```diff
@@ -36,5 +36,9 @@ def failed_request_callback(
...
     _cleanup()
     set_request_state(request_id, 'failed', msg)
```
xDaile, line 44: And as @yashvardhannanavati mentioned, it would loop here again.

iib/workers/tasks/general.py
ashwgit: fixed

iib/workers/tasks/general.py
ashwgit: @xDaile  could you try testing it locally, as while testing this one, I did found that it is breaking the loop.

iib/workers/api_utils.py
```diff
@@ -145,7 +145,8 @@ def update_request(
     :param str exc_msg: an optional custom exception that can be a template
     :return: the updated request
     :rtype: dict
-    :raises IIBError: if the request to the IIB API fails
+    :raises ValidationError: if the request fails trying changing final state (complete, failed)
```
yashvardhannanavati, line 148: nit: this needs to be changed to FinalStateOverwriteError.

yashvardhannanavati:
one minor comment. Otherwise, LGTM!

@ashwgit could you please rebase the PR too?

iib/workers/api_utils.py
ashwgit: done, thanks for having it noticed

tests/test_workers/test_tasks/test_general.py
```diff
@@ -12,6 +12,7 @@
     (
         (IIBError('Is it lunch time yet?'), 'Is it lunch time yet?'),
         (RuntimeError('I cannot run in the rain!'), 'An unknown error occurred'),
+        (FinalStateOverwriteError("can not overwite final state"), "Already in final state"),
```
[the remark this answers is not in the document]
ashwgit: as we are already testing for the error type, perhaps we can skip the log check.

PR #1107, opened 2025-08-20
fix(git_utils): get_git_token format adjustment

JAVGan:
This commit fixes a bug on `get_git_token` as the
`iib_index_configs_gitlab_tokens_map` will always be a `dict[str, str]` from HashiVault, never being able to be a Tuple.

In order to return a Tuple we need to split the token-name from token-value. For that, we're stablishing the standard as a column `:` to separate such values.

Refers to CLOUDDST-29117

## Summary by Sourcery

Standardize the Git token mapping to use colon-separated strings instead of tuples and add validation for the new format

Bug Fixes:
- Raise an error if a token string lacks the expected ':' separator

Enhancements:
- Parse token mappings as "token_name:token_value" strings in get_git_token

Documentation:
- Update README to document the new "token-name:token-value" mapping format

Tests:
- Update unit tests to use the new string format for tokens and add a test for missing token format

### DIFF
--- a/README.md
+++ b/README.md
@@ -340,6 +340,7 @@ The custom configuration options for the Celery workers are listed below:
   `{registry}/iib-build:{request_id}`.
 * `iib_index_configs_gitlab_tokens_map` - A map of index image addresses to GitLab tokens.
   These Gitlab repositories are intended to store image `/configs` directories.
+  Its format should be the full repository URL as keys and `token-name:token-value` as value.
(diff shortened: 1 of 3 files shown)


JAVGan:
@sourcery-ai review

yashvardhannanavati (dismissed):

yashvardhannanavati (approved):

PR #1112, opened 2025-08-21
Fix bug on failed_request_callback

JAVGan:
This commit fixes a bug on `failed_request_callback` which were causing the `IIBError` messages to being overwritten with `An unknown error occurred. See logs for details`.

It also fixes the unit-test that wasn't properly constructed and didn't catch this issue.

## Summary by Sourcery

Fix failed_request_callback to preserve exception-specific messages and use the generic error message only for unknown errors, and update its unit test to validate the new behavior

Bug Fixes:
- Prevent failed_request_callback from overwriting specific exception messages with a generic unknown error message
- Ensure FinalStateOverwriteError does not trigger a state update

Tests:
- Correct test_failed_request_callback to use the full default error message and properly assert set_request_state calls
- Add conditional assertion to skip set_request_state for FinalStateOverwriteError

### DIFF
--- a/iib/workers/tasks/general.py
+++ b/iib/workers/tasks/general.py
@@ -36,9 +36,9 @@ def failed_request_callback(
         log.info(f"Request {request_id} is in a final state,ignoring update.")
         _cleanup()
         return
-
-    msg = 'An unknown error occurred. See logs for details'
-    log.error(msg, exc_info=exc)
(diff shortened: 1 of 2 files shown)


JAVGan:
@sourcery-ai review

yashvardhannanavati (approved):

PR #1123, opened 2025-08-28
fix(CLOUDDST-29225): remove check for non-empty configs dir

nmars:
Removed the check for sub-directories under `/configs` when pushing to Git. It is valid to build an "empty" index image with no operators added to it.

## Summary by Sourcery

Remove requirement for non-empty configs directory to allow empty index images and streamline configs copying to the Git repo

Bug Fixes:
- Allow pushing empty configs directories by removing check for subdirectories in `/configs`

Enhancements:
- Validate the source configs path as a directory
- Simplify configs copying by cleaning the target directory and using `shutil.copytree` instead of per-package loops

### DIFF
--- a/iib/workers/tasks/git_utils.py
+++ b/iib/workers/tasks/git_utils.py
@@ -61,13 +61,11 @@ def push_configs_to_git(
         raise IIBError(f"Remote branch '{branch}' not found for repo {repo_url}")
 
     # Verify the path to existing /configs is correct
-    if not os.path.exists(src_configs_path):
+    if not os.path.isdir(src_configs_path):
         raise IIBError(f"Catalog configs directory does not exist: {src_configs_path}")
 
-    # Make sure there are subdirs under /configs for the operator package dirs
+    # List operators content if any
     operator_packages = os.listdir(src_configs_path)
-    if not operator_packages:
-        raise IIBError(f"No packages found in configs directory {src_configs_path}")
 
     # Clone/checkout remote repo in temp dir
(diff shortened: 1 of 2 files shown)


nmars:
@sourcery-ai review

iib/workers/tasks/git_utils.py
```diff
@@ -78,18 +73,17 @@ def push_configs_to_git(
...
                 "Copying content of %s to local Git repository %s",
                 src_configs_path,
                 repo_configs_dir,
             )
-            for operator_package in operator_packages:
-                src_pkg_dir = os.path.join(src_configs_path, operator_package)
-                dest_pkg_dir = os.path.join(repo_configs_dir, operator_package)
-                os.makedirs(dest_pkg_dir, exist_ok=True)
-                shutil.copytree(src_pkg_dir, dest_pkg_dir, dirs_exist_ok=True)
+            shutil.copytree(src_configs_path, repo_configs_dir)
```
JAVGan, line 86: Will this work if `configs` get deleted? It is, will it recreate this config dir?

JAVGan (dismissed):
LGTM

JAVGan:
@nmars please wait for @yashvardhannanavati ack before merging

iib/workers/tasks/git_utils.py
JAVGan, line 86: @yashvardhannanavati do you know if it's ok to remove and add the `/configs` from the index image?

Or should we continue merging it like before? Just supporting an empty dir?

PR #1126, opened 2025-08-29
git_utils: Fix behavior for RM request

JAVGan:
The `RM` request should cause the `git_utils` to remove the operators from the `git` repository and commit the changes, which was something that didn't happen in the previous versions.

This commit changes the `git_utils` module as by adding a new parameter to `push_configs_to_git` named `rm_operators` which should only receive a list of operators when a `RM` operation is handled.

It also updates the `build` module to send the `RM` operators list to `push_configs_to_git` by passing it over to the necessary functions.

Finally, it adds new unit-tests and improve some existing ones.

Signed-off-by: Jonathan Gangi <jgangi@redhat.com>

Assisted-by: Cursor/Gemini

## Summary by Sourcery

Add support for handling RM requests by introducing a new rm_operators parameter to push_configs_to_git, implementing removal logic for operators during Git updates, and propagating this parameter through the build workflow. Update and extend unit tests to cover removal scenarios and introduce a RegexMatcher helper for verifying directory deletions.

New Features:
- Add rm_operators parameter to push_configs_to_git to enable operator removal on RM requests
- Propagate rm_operators through build tasks (_update_index_image_pull_spec, _overwrite_from_index, handle_rm_request) to push_configs_to_git

Enhancements:
- Implement logic in git_utils to remove specified operator directories when rm_operators is provided
- Introduce RegexMatcher utility in tests for regex-based call assertions

Tests:
- Add comprehensive unit tests for push_configs_to_git removal scenarios (partial, full, none, and empty repo)
- Update existing tests to include rm_operators parameter and verify commit_and_push behavior

### DIFF
--- a/iib/workers/tasks/git_utils.py
+++ b/iib/workers/tasks/git_utils.py
@@ -4,7 +4,7 @@ import logging
 import os
 import tempfile
 import shutil
-from typing import Dict, Optional, Tuple
+from typing import Dict, Optional, Tuple, List
 
 from operator_manifest.operator import ImageName
 import requests
@@ -27,6 +27,7 @@ def push_configs_to_git(
     src_configs_path: str,
     index_repo_map: Dict[str, str],
     commit_message: Optional[str] = None,
+    rm_operators: Optional[List[str]] = None,
 ) -> None:
     """
     Pushes /configs subfolders to a Git repository.
@@ -39,6 +40,8 @@ def push_configs_to_git(
                            files reside.
     :param dict(str) index_repo_map: The repo mapping to resolve the git URL.
     :param str commit_message: Custom commit message. If None, a default message is used.
+    :param list(str) rm_operators: List of operators to remove from the index image. Only
+        required if ``add_or_rm`` is ``True`` and an RM operation is requested.
     :raises IIBError: If src_configs_path is not found, remote branch does not exist,
                       or a Git operation fails.
     """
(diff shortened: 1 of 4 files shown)


iib/workers/tasks/git_utils.py
```diff
@@ -84,14 +87,21 @@ def push_configs_to_git(
                 repo_configs_dir,
             )
 
-            if operator_packages:
-                for operator_package in operator_packages:
-                    src_pkg_dir = os.path.join(src_configs_path, operator_package)
-                    dest_pkg_dir = os.path.join(repo_configs_dir, operator_package)
-                    os.makedirs(dest_pkg_dir, exist_ok=True)
-                    shutil.copytree(src_pkg_dir, dest_pkg_dir, dirs_exist_ok=True)
+            if rm_operators:
```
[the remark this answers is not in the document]
JAVGan: Nice catch! Updating

iib/workers/tasks/git_utils.py
```diff
@@ -84,13 +87,21 @@ def push_configs_to_git(
                 repo_configs_dir,
             )
 
-            if operator_packages:
+            if rm_operators:
+                # Remove operators from the Git repo if an RM operation is requested
+                for operator_package in set(rm_operators):
+                    if operator_package:  # We don't want to remove the whole /configs directory
+                        operator_dir = os.path.join(repo_configs_dir, operator_package)
+                        shutil.rmtree(operator_dir)
```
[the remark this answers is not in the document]
JAVGan: @yashvardhannanavati @nmars WDYT? Should I do it?

nmars (approved):
Looks good!

iib/workers/tasks/git_utils.py
```diff
@@ -84,13 +87,24 @@ def push_configs_to_git(
                 repo_configs_dir,
             )
 
-            if operator_packages:
+            if rm_operators:
+                # Remove operators from the Git repo if an RM operation is requested
+                for operator_package in set(rm_operators):
+                    if operator_package:  # We don't want to remove the whole /configs directory
```
yashvardhannanavati, line 93: Consider changing lines 93-94 to 
```
operator_dir = os.path.join(repo_configs_dir, operator_package)
if os.path.isdir(operator_dir):
    try
    ...
```

JAVGan: But in this case if `operator_package` is an empty string the `os.path.isdir(operator_dir)` will return True and we'll remove the `/configs`, no? The idea here is to check if `operator_package` is not empty so it can be safely removed.

Or is there a better way to accomplish that?

JAVGan: Example:

```
repo_configs_dir = "/tmp/configs"
operator_package = ""

operator_dir = os.path.join(repo_configs_dir, operator_package)
os.path.isdir(operator_dir)  # will result in /tmp/configs
shutil.rmtree(operator_dir) # xxx
```

yashvardhannanavati: @JAVGan In case the `operator_package` dir exists but it empty, running shutil.rmtree on `/configs/operator_package` would only remove the `operator_package` directory and not the parent directory. 

In your changes, on line 91, you are checking if `rm_operators` is not empty. So the operator_package variable will never be an empty string, no?

JAVGan: Hmm ok, makes sense!

I'll patch it then as you suggested

JAVGan: Patched, PTAL again

PR #1304, opened 2026-04-02
Check for user in iib_no_ocp_label_allow_list

MichalZelenak:
## Summary by Sourcery

Allow bundle addition when the requesting user is configured to bypass missing OCP version labels.

Enhancements:
- Extend _add_bundles_missing_in_source to treat the requesting user as eligible for no-OCP-label allowances based on worker configuration.

Tests:
- Add tests covering behavior when the requesting user is and is not included in iib_no_ocp_label_allow_list with ignore_bundle_ocp_version enabled.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -9,15 +9,15 @@ from datetime import datetime
 import os
 import sys
 
-import pkg_resources
+from importlib.metadata import PackageNotFoundError, version as get_version
 
 # -- Path setup --------------------------------------------------------------
 sys.path.append(os.path.abspath('../'))
 
 # -- Project information -----------------------------------------------------
 try:
-    version = pkg_resources.get_distribution('iib').version
-except pkg_resources.DistributionNotFound:
+    version = get_version('iib')
+except PackageNotFoundError:
     version = 'unknown'
 project = 'IIB Image Builder Service'
 copyright = datetime.today().strftime('%Y') + ', Red Hat Inc.'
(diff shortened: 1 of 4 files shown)


iib/workers/tasks/build_merge_index_image.py
```diff
@@ -166,9 +166,12 @@ def _add_bundles_missing_in_source(
 
     if ignore_bundle_ocp_version:
         target_index_tmp = '' if target_index is None else target_index
+        user = get_request(request_id)['user']
         allow_no_ocp_version = any(
-            target_index_tmp.startswith(index) or source_from_index.startswith(index)
-            for index in get_worker_config()['iib_no_ocp_label_allow_list']
+            target_index_tmp.startswith(entry)
```
yashvardhannanavati, line 171: Can you also update README.md to specify this change?

yashvardhannanavati (changes requested):

docs/conf.py
```diff
@@ -44,7 +44,7 @@
 
 # -- Extension configuration -------------------------------------------------
 # This must be mocked because Read the Docs doesn't have krb5-devel installed
-autodoc_mock_imports = ["requests_kerberos"]
+autodoc_mock_imports = ["requests_kerberos", "operator_manifest"]
```
chandwanitulsi, line 47: can you share why do we have this change in the PR?

chandwanitulsi (commented):
Overall looks good, one question on unrelated change

lipoja:
I agree with the comments from Yash and Tulsi. Let's address it and then we can do final approval.

lipoja:
> [!IMPORTANT]
> Please propagate this change to main branch as well. 

@MichalZelenak

iib/workers/tasks/build_merge_index_image.py
MichalZelenak: Updated

docs/conf.py
MichalZelenak: I added another commit where I reverted this change - check the pipeline now (docs build fails because pkg_resources is not accessible). 

I will revert the commit once discussed. Should I split the change (this docs change being the second one), into two commits? Or open a separate PR for that?

chandwanitulsi (approved):

yashvardhannanavati (approved):

PR #1305, opened 2026-04-06
Allow using separate Docker config.json for ORAS

JAVGan:
This commit changes the `get_oras_artifact` and `push_oras_artifact` to allow using a default Docker config.json, containing only credentials for the ORAS operations, whenever the following conditions are met:

1. The worker variables `iib_index_db_artifact_registry` and `iib_index_db_artifact_registry` are properly set with the registry URL and its pull/push secret.
2. The methods `[pull|push]_oras_artifact` are called without any value being set to the parameter `registry_auths`, forcing it to use the default value.

It also adds unit-tests for this new change.

Refers to CLOUDDST-32383

Signed-off-by: Jonathan Gangi <jgangi@redhat.com>

Assisted-by: Cursor/Gemini

## Summary by Sourcery

Allow ORAS artifact pull/push operations to use a dedicated registry auth secret from worker configuration when no explicit auth is provided.

New Features:
- Introduce support for a worker-configured ORAS registry auth secret used as a dedicated dockerconfig for ORAS operations.

Enhancements:
- Default ORAS pull and push helpers to use worker-configured ORAS auth when available, falling back to the standard Docker config.json otherwise.
- Extend indexdb cache verification and refresh logic to consume worker configuration for artifact registry and template.

Documentation:
- Document the new `iib_index_db_oras_auth_secret` worker configuration option and its behavior in ORAS authentication.

Tests:
- Add unit tests covering ORAS auth resolution from worker configuration for pull and push operations, and update cache-sync tests to use the new configuration fields.


JAVGan:
@sourcery-ai review

iib/workers/tasks/oras_utils.py
```diff
@@ -121,7 +151,11 @@ def get_oras_artifact(
...
     temp_dir = tempfile.mkdtemp(prefix=temp_dir_prefix, dir=base_dir)
 
-    # Use namespace-specific registry authentication if provided
+    # If no registry_auths are provided, use the secret from the worker configuration
+    registry_auths = registry_auths or _get_oras_registry_auths()
+
+    # Use namespace-specific registry authentication if provided or the secret from the
+    # worker configuration by defaut.
     with set_registry_auths(registry_auths, use_empty_config=True):
         try:
```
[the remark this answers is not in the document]
yashvardhannanavati: @JAVGan thoughts on this? this would be a valid concern if we set `iib_index_db_oras_auth_secret` to only have access to Quay. 

How about extracting auth for `registry/namespace` of the artifact_ref from the default config.json? It should fallback to using the default config.json if auth isn't found for that specific registry/namespace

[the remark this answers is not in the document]
JAVGan: I think it makes sense. In this case, however, we shouldn't directly call the `set_registry_auths`, with `use_empty_config` right? Otherwise we would fall in the same scenario:

- Extract the creds from artifact_ref and build a config.json just with this creds
- Using `set_registry_auths` with `use_empty_config` == bug

Is this correct? Then should we directly pass the creds via ORAS arguments? I would like to hear your suggestion on this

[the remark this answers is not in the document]
yashvardhannanavati: here's what I am thinking. We can have `iib_index_db_oras_auth_secret ` as a config.json file which would have auths for all oras registries. Then, we can simply call oras with `--registry-config` and pass that file.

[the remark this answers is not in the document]
lipoja: I am inclined to the solution that @yashvardhannanavati mentioned.
We do not need to move around with configs, just force the right one to oras.
ORAS will be used only within IIB - it is part of its internal workflow.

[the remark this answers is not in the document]
JAVGan: I've updated it in a separate commit, so we can track the changes. If looks good I can squash it later on.

I initially wrote how I assumed it should work then asked AI to review and apply some fixes, please let me know if it makes sense or if it's overkill

JAVGan:
@yashvardhannanavati can you please take another look?

iib/workers/tasks/oras_utils.py
```diff
@@ -15,6 +17,77 @@
...
+        return 'docker.io'
+
+    first = parts[0]
+    if '.' in first or ':' in first or first == 'localhost':
+        return first
+    # e.g. ``foo/bar:latest`` — path without registry prefix
+    return 'docker.io'
+
+
+def _build_oras_registry_config_dict(
```
lipoja: Do we even need this?
I thought we will have the whole config for oras in hashivault.

lipoja: I thought the change in IIB would be:
 - check if docker config is on given path (set in config)
 - if it exists then use it as argument while calling oras

and in iib-playbooks you will have the deployment of the config to correct location and setting the config variable.

JAVGan: Yeah, I think the best solution is doing it in the simplest way. Let me get rid of the second commit and redo this in order that the docker config.json will be set in the hashivault directly.

JAVGan:
@lipoja @yashvardhannanavati PTAL again. I've changed the code and it's simpler now. Please let me know if the current approach solves our needs.

Thanks!

iib/workers/tasks/oras_utils.py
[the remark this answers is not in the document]
JAVGan: Patched, I'll resolve this and wait for more reviews, feel free to open new topics.

iib/workers/tasks/oras_utils.py
```diff
@@ -94,11 +95,85 @@ def get_indexdb_artifact_pullspec(from_index: str) -> str:
...
+        'auths': {
+            conf['iib_index_db_artifact_registry']: {
+                'auth': conf['iib_index_db_oras_auth_secret'],
+            },
+        },
+    }
+
+
+@contextmanager
+def _set_oras_registry_auths() -> Generator[Optional[str], None, None]:
```
lipoja: @JAVGan I am sorry, but I have the same question as before. Do we need this function?

`.docker/config.json` is picked up automagically when you do not specify `--registry-config path`

iib/workers/tasks/oras_utils.py
```diff
@@ -121,11 +195,13 @@ def get_oras_artifact(
     # Create a subdirectory within the provided base_dir
     temp_dir = tempfile.mkdtemp(prefix=temp_dir_prefix, dir=base_dir)
 
-    # Use namespace-specific registry authentication if provided
-    with set_registry_auths(registry_auths, use_empty_config=True):
+    with _set_oras_registry_auths() as exclusive_oras_registry_path:
```
lipoja: Why don't we keep it as it is and do just:

add `iib_docker_config_oras` as variable to config.py (similar to `iib_docker_config_template`)

and here we can just check if it is set and file exists.

```python
cmd_args = []
conf = get_worker_config()
if conf['iib_docker_config_oras'] and os.path.exists(conf['iib_docker_config_oras'] ):
    cmd_args = ['--registry-config', exclusive_oras_registry_path]
else:
 #   log warning or info that config for oras is not set and default will be used.
```

lipoja:
@JAVGan Thank you for working on it. Does the suggested changes make sense to you?
Or do you think it is better to do it the way you proposed? You can ping me on slack for discussion.

iib/workers/tasks/oras_utils.py
JAVGan: Yeah, that makes sense! I thought I had to do it differently, thanks for the suggestion! I'll apply it

iib/workers/tasks/oras_utils.py
JAVGan: Yeah, my mistake. I thought we would need it 😅 

I'll drop it in favor of your suggestion, thanks a ton!

iib/workers/tasks/oras_utils.py
```diff
@@ -94,6 +94,60 @@ def get_indexdb_artifact_pullspec(from_index: str) -> str:
...
+def _get_oras_registry_auths() -> Optional[Dict[str, Any]]:
+    """
+    Get the registry authentication for ORAS.
+
+    It reads the secret from the worker configuration and returns the authentication data.
+    :return: The registry authentication for ORAS in the format of a dictionary when available.
+    :rtype: dict
+    """
+    conf = get_worker_config()
+    if not (conf['iib_index_db_artifact_registry'] and conf['iib_index_db_oras_auth_secret']):
```
lipoja: For what (and when) will be the `iib_index_db_oras_auth_secret` used for?
Do we have use-case?

JAVGan:
@lipoja I got rid of the leftovers of my previous solutions, now it seems ok. Can you PTAL again? Sorry for all the previous mess 😓

README.md
```diff
@@ -321,10 +321,11 @@ The custom configuration options for the Celery workers are listed below:
    ImageStream is pushed. This is typically an internal OpenShift registry or another registry
    dedicated to hosting ImageStreams for `index.db` artifacts. If unset, caching of index.db
   artifacts will be disabled.
-* `iib_index_db_artifact_registry` - the container registry where `index.db` artifact images
-  (for example `index-db:<tag>`) are stored and from which they are distributed. This is often
-  a central or dedicated registry for artifacts generated by IIB. This value **must be set** in
-  order for `index.db` artifacts to be pushed and for configuration validation to succeed.
+* `iib_index_db_artifact_registry`. When `iib_index_db_oras_auth_path` is unset and this is set
```
yashvardhannanavati: I think this change deleted some of the information from README. Could you please add that back. 

``iib_index_db_artifact_registry`. When ` => ``iib_index_db_artifact_registry` - When `

yashvardhannanavati (dismissed):
one nit comment. Otherwise, LGTM

JAVGan:
@yashvardhannanavati PTAL again

yashvardhannanavati (approved):

PR #1310, opened 2026-04-17
Check for user in iib_no_ocp_label_allow_list main propagation

MichalZelenak:
Propagation of the changes made in this [PR](https://github.com/release-engineering/iib/pull/1307)

### DIFF
--- a/compose-files/docker-compose.yml
+++ b/compose-files/docker-compose.yml
@@ -68,7 +68,7 @@ services:
     container_name: jaeger
     ports:
       - 4318:4318
-      - 16686:16686 # web
+      - 16686:16686  # web
     environment:
       - COLLECTOR_OTLP_ENABLED=true
 
--- a/compose-files/podman-compose.yml
+++ b/compose-files/podman-compose.yml
@@ -67,7 +67,7 @@ services:
     container_name: jaeger
     ports:
       - 4318:4318
-      - 16686:16686 # web
+      - 16686:16686  # web
     environment:
       - COLLECTOR_OTLP_ENABLED=true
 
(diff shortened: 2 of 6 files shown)


docs/conf.py
```diff
@@ -9,15 +9,15 @@
 import os
 import sys
 
-import pkg_resources
+from importlib.metadata import PackageNotFoundError, version as get_version
```
lipoja, line 12: For now - we should try to rebase main on tom of master branch. 
We want to keep these two branches in sync before we switch them. 

That means if you rebase it then this changes will be already present.

MichalZelenak: There would be a merge conflict with the two commits related to this PR ([commit from this PR](https://github.com/release-engineering/iib/pull/1310/changes/dec0694f3c97a8bed219e3db42f4a45d669721f6)), [commit merged to master](https://github.com/release-engineering/iib/pull/1304/changes/993fef1ab05f3269dfd1c8f00a83938ce0ec1568). 

We could cherry-pick certain commits from the master and get them into main.

MichalZelenak: Another point: PR checks would fail if I remove the docs commit.

lipoja: "PR checks would fail if I remove the docs commit." - this will be solved by rebase because you've already fixed that in master, right?

I am not sure about the other commits (if they are causing conflicts), but we should `git rebase master main`  to keep it in sync.

MichalZelenak: So, to close this PR, we should just remove the docs commit?

lipoja: Remove docs commit, and `git rebase master main`

lipoja (approved):

PR #1313, opened 2026-04-27
Remove podman usage

lipoja:
Adding extraction of directory using oc command

We can't use podman inside container, and we need to extract data from index image. The best way for our purposes was to use `oc image extract`. 

This commit adds extract_directory_from_image_non_privileged used to extract data from image.

## Summary by Sourcery

Replace podman-based image directory extraction with a non-privileged `oc image extract` helper and wire it into FBC fragment and related image inspection workflows.

New Features:
- Add `extract_directory_from_image_non_privileged` helper to extract directories from container images using `oc image extract`.

Enhancements:
- Refactor FBC fragment extraction and related image inspection to use the new non-privileged image directory extraction helper instead of `_copy_files_from_image`.
- Set a default ORAS auth config path for index.db artifacts in the worker configuration.

Tests:
- Extend and update tests to cover the new extraction helper behavior and its integration into FBC fragment and related image inspection logic.

### DIFF
--- a/docs/requirements.txt
+++ b/docs/requirements.txt
@@ -26,7 +26,7 @@ requests
 ruamel.yaml
 semver
 setuptools
-sphinx
+sphinx==8.2.3
 sphinx_rtd_theme
 tenacity
 typing-extensions
--- a/iib/workers/tasks/build.py
+++ b/iib/workers/tasks/build.py
@@ -24,7 +24,12 @@ from iib.workers.api_utils import set_request_state, update_request
 from iib.workers.config import get_worker_config
 from iib.workers.tasks.celery import app
 from iib.workers.greenwave import gate_bundles
-from iib.workers.tasks.fbc_utils import is_image_fbc, get_catalog_dir, merge_catalogs_dirs
+from iib.workers.tasks.fbc_utils import (
+    is_image_fbc,
+    get_catalog_dir,
+    merge_catalogs_dirs,
+    extract_directory_from_image_non_privileged,
+)
 from iib.workers.tasks.git_utils import push_configs_to_git, revert_last_commit
 from iib.workers.tasks.opm_operations import (
     opm_registry_add_fbc,
@@ -733,7 +738,9 @@ def inspect_related_images(
             bundle, "operators.operatorframework.io.bundle.manifests.v1"
         )
         with tempfile.TemporaryDirectory(prefix=f'iib-{request_id}-') as temp_dir:
-            _copy_files_from_image(bundle, manifest_location, temp_dir)
+            extract_directory_from_image_non_privileged(
+                image=bundle, src_path=manifest_location, dest_path=temp_dir
+            )
             manifest_path = os.path.join(temp_dir, manifest_location)
             try:
                 operator_manifest = OperatorManifest.from_directory(manifest_path)
--- a/iib/workers/tasks/fbc_utils.py
+++ b/iib/workers/tasks/fbc_utils.py
@@ -102,6 +102,83 @@ def merge_catalogs_dirs(src_config: str, dest_config: str):
     opm_validate(conf_dir)
 
 
+def extract_directory_from_image_non_privileged(image: str, src_path: str, dest_path: str) -> None:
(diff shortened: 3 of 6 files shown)


iib/workers/tasks/fbc_utils.py
```diff
@@ -102,6 +102,73 @@ def merge_catalogs_dirs(src_config: str, dest_config: str):
...
+    original_src_path = src_path
+
+    # ``oc`` glob is ``<dir>/*``; strip trailing slashes so we do not end up with ``//`` in paths.
+    stripped = src_path.rstrip('/')
+    if not stripped:
+        log.error(f'src_path must name a directory under /, got {original_src_path!r}')
+        raise IIBError(f'src_path must name a directory under /, got {original_src_path!r}')
+
+    # Collapse ``.``, ``..``, and repeated slashes; log when the effective path differs from input.
+    normalized = os.path.normpath(stripped)
```
yashvardhannanavati: @lipoja please consider using `pathlib` instead of `os.path` like we discussed a while ago.

yashvardhannanavati (changes requested):

iib/workers/tasks/fbc_utils.py
lipoja: Updated please check

docs/requirements.txt
```diff
@@ -26,7 +26,7 @@ requests
 ruamel.yaml
 semver
 setuptools
-sphinx
+sphinx==8.2.3
```
MichalZelenak, line 29: Why are we pinning Sphinx?

iib/workers/tasks/build.py
```diff
@@ -733,7 +738,7 @@ def inspect_related_images(
             bundle, "operators.operatorframework.io.bundle.manifests.v1"
         )
         with tempfile.TemporaryDirectory(prefix=f'iib-{request_id}-') as temp_dir:
-            _copy_files_from_image(bundle, manifest_location, temp_dir)
+            extract_directory_from_image_non_privileged(bundle, manifest_location, temp_dir)
```
MichalZelenak: Is there some standard/rule, about how we are calling functions? I would prefer the format with named variables, e.q.
```
extract_directory_from_image_non_privileged(
    image=bundle,
    src_path=manifest_location,
    dest_path=temp_dir,
)
```

IMO the code is more readable that way.
I get that, it is not suitable for every call, but here, for me, it makes sense.

iib/workers/tasks/fbc_utils.py
```diff
@@ -115,15 +192,15 @@ def extract_fbc_fragment(
...
     """
-    from iib.workers.tasks.build import _copy_files_from_image
-
     log.info("Extracting the fbc_fragment's catalog from  %s", fbc_fragment)
     # store the fbc_fragment at /tmp/iib-**/fbc-fragment-{index} to prevent
     # cross-contamination
     conf = get_worker_config()
     fbc_fragment_path = os.path.join(temp_dir, f"{conf['temp_fbc_fragment_path']}-{fragment_index}")
     # Copy fbc_fragment's catalog to /tmp/iib-**/fbc-fragment-{index}
-    _copy_files_from_image(fbc_fragment, conf['fbc_fragment_catalog_path'], fbc_fragment_path)
```
MichalZelenak, line 126: Would it make sense to annotate/document, that we can remove this method in the future?

docs/requirements.txt
lipoja: Because docs is failing to build. It is in separate commit.
It was pinned before but after the rebase it dissapeared.

iib/workers/tasks/fbc_utils.py
lipoja: Hmm I it is still part of the old IIB, and we did not annotate the rest ether. 
So for me it does not make sens for now. Once we know IIB is fully working on OpenShift, we will remove the old code from repo.

MichalZelenak (approved):

yashvardhannanavati (approved):

PR #1331, opened 2026-05-14
Agentready quick fixes

yashvardhannanavati:


.github/workflows/agentready.yaml
```diff
@@ -0,0 +1,26 @@
+---
+name: AgentReady Assessment
+
+on:
+  pull_request:
```
chandwanitulsi, line 5: Should we not have this on main branch only?

yashvardhannanavati: yes, it should be. Fixed it

chandwanitulsi (approved):

PR #1334, opened 2026-05-26
Fixing extract_fbc_fragment

lipoja:
At this moment IIB creates PR where `/config/operator` is copied to incorrect location `/config/config/operator`

This PR is introducing a fix.
Extract data correctly and return path to config directory.


lipoja:
Targeted incorrect branch. Closing.

PR #1338, opened 2026-06-04
Fix _overwrite_from_index for containerized IIB

lipoja:
We should not push to Git when using containerized IIB inside _overwrite_from_index because the push is already done earlier in the code.


yashvardhannanavati:
@lipoja consider adding unit tests for this?

yashvardhannanavati (approved):

PR #1343, opened 2026-06-10
Refactor set_registry_token to support repository-specific auth

chandwanitulsi:
Updated the set_registry_token function to allow for appending tokens specific to a repository while leaving registry-level credentials intact. Adjusted related tests to verify the new behavior, ensuring proper handling of authentication entries for both repository and registry levels.

### DIFF
--- a/iib/workers/tasks/utils.py
+++ b/iib/workers/tasks/utils.py
@@ -574,23 +574,65 @@ def reset_docker_config() -> None:
         os.symlink(conf.iib_docker_config_template, docker_config_path)
 
 
+def _docker_auth_key_for_image(container_image: str) -> str:
+    """
+    Return the docker config ``auths`` key for ``container_image``.
+
+    Uses ``registry/namespace/repo`` when a namespace is present, since that is the most
+    specific key container runtimes reliably match. Without a namespace, only registry-level
+    keys are used because ``registry/repo`` matching is not consistent across runtimes.
+
+    :param str container_image: the pull specification of the container image
+    :return: the docker config ``auths`` key for ``container_image``
+    :rtype: str
+    :raises IIBError: if the pull specification does not contain a resolvable registry.
+    """
+    image_name = ImageName.parse(container_image)
+    registry = image_name.registry
+    if image_name.namespace and registry:
+        return f'{registry}/{image_name.namespace}/{image_name.repo}'
+    if registry:
+        return registry
+    raise IIBError(
+        f'Unable to determine the registry from the pull specification {container_image!r} '
+        'for Docker authentication'
+    )
+
+
 @contextmanager
 def set_registry_token(
     token: Optional[str], container_image: Optional[str], append: bool = False
 ) -> Generator:
     """
-    Configure authentication to the registry that ``container_image`` is from.
+    Configure authentication for the image identified by ``container_image``.
 
-    This context manager will reset the authentication to the way it was after it exits. If
-    ``token`` is falsy, this context manager will do nothing.
+    The token is written to ``~/.docker/config.json`` under the most specific ``auths`` key that
+    container runtimes reliably match for that pull specification:
+
+    * ``registry/namespace/repo`` when ``container_image`` includes a namespace
+    * ``registry`` when the image has no namespace (for example, ``localhost:5000/myimage:tag``)
+
+    If the pull specification does not contain a resolvable registry, :exc:`IIBError` is raised.
+
+    Broader credentials already present in the Docker configuration, such as registry- or
+    namespace-level entries for the same host, are not modified. Only the scoped ``auths`` key
+    derived from ``container_image`` is set or overwritten.
+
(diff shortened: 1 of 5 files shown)


chandwanitulsi:
Fixed both the low and [architectural-coherence] medium findings by full-send, the merge-index-image finding is a partially valid case as we don't use overwrite_from_index token for source_image and that seems to be separate issue

iib/workers/tasks/utils.py
```diff
@@ -621,8 +662,9 @@ def set_registry_token(
 
                 log.debug('Docker config will be updated')
 
-    log.debug('Setting the override token for the registry %s ', registry)
-    registry_auths['auths'].update({registry: {'auth': encoded_token}})
+    log.debug('Setting the override token for the image %s', auth_key)
+    registry_auths.setdefault('auths', {})
```
lipoja: Is this `setdefault` necessary?
I am asking because it should be already defined on line 652 `registry_auths: Dict[str, Any] = {'auths': {}}`
Or am I wrong?

lipoja: otherwise LGTM. Can be merged if fixed or if I was wrong :)

iib/workers/tasks/utils.py
```diff
@@ -574,21 +574,60 @@ def reset_docker_config() -> None:
...
+    specific key container runtimes reliably match. Without a namespace, only registry-level
+    keys are used because ``registry/repo`` matching is not consistent across runtimes.
+    """
+    image_name = ImageName.parse(container_image)
+    registry = image_name.registry
+    if image_name.namespace:
+        return f'{registry}/{image_name.namespace}/{image_name.repo}'
+    if registry:
+        return registry
+    if image_name.repo:
```
yashvardhannanavati: I think we should drop support for this case as it seems highly unlikely

iib/workers/tasks/utils.py
yashvardhannanavati: @chandwanitulsi +1, we are defining it on line 652, we can get rid of this line

yashvardhannanavati (changes requested):

iib/workers/tasks/utils.py
chandwanitulsi: So setdefault('auths', {}) is a guard for the append+load path, where the loaded docker config might not include auths.

Since we always start from config.json.template I removed this case assuming the template would always have right format

chandwanitulsi: Updated in next path

iib/workers/tasks/utils.py
chandwanitulsi: Updated in next patch

yashvardhannanavati:
@chandwanitulsi the CI seems to be failing

iib/workers/tasks/utils.py
```diff
@@ -574,23 +574,62 @@ def reset_docker_config() -> None:
...
 
+def _docker_auth_key_for_image(container_image: str) -> str:
+    """
+    Return the docker config ``auths`` key for ``container_image``.
+
+    Uses ``registry/namespace/repo`` when a namespace is present, since that is the most
+    specific key container runtimes reliably match. Without a namespace, only registry-level
+    keys are used because ``registry/repo`` matching is not consistent across runtimes.
+
+    :raises IIBError: if the pull specification does not contain a resolvable registry.
```
yashvardhannanavati, line 588: nit: consider adding the following to the docstring

```
    :param str pull_spec: the pull specification of the container image to resolve
    :return: the resolved auth key
    :rtype: str
```

yashvardhannanavati (commented):
one minor comment. otherwise, LGTM

iib/workers/tasks/utils.py
lipoja: +1 - just use correct variable names

chandwanitulsi: Done

tests/test_workers/test_tasks/test_utils.py
```diff
@@ -96,19 +96,31 @@ def test_set_registry_token(
...
 
+@pytest.mark.parametrize(
+    'container_image, expected_auth_key',
+    (
+        ('registry.redhat.io/ns/repo:latest', 'registry.redhat.io/ns/repo'),
+        ('localhost:5000/myimage:tag', 'localhost:5000'),
+        ('myregistry.io:5000/', 'myregistry.io:5000'),
+    ),
+)
+def test_docker_auth_key_for_image(container_image, expected_auth_key):
```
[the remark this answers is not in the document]
chandwanitulsi: that's quite unlikely

iib/workers/tasks/utils.py
```diff
@@ -621,8 +655,8 @@ def set_registry_token(
```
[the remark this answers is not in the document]
chandwanitulsi: verified with skopeo

yashvardhannanavati (approved):

PR #1345, opened 2026-06-17
Fix containerized regenerate-bundle handler and related_images function for add

yashvardhannanavati:
- commit 1: The tests use keyword arguments but the call uses positional args. Hence, the bug wasn't caught.
- commit 2: Bundle images set the manifests.v1 label to 'manifests/' (relative),
but extract_directory_from_image_non_privileged requires an absolute
container path. Normalize the label value to absolute and use basename
for the local extraction path to avoid os.path.join discarding temp_dir


lipoja:
I rebased master to main. Can you pull and rebase on main, thanks you.

lipoja (approved):

PR #1365, opened 2026-07-27
Fix Git FBC catalog divergence after bundle deprecation in containerized handlers

lipoja:
## Summary

Fixes Git FBC catalog divergence after bundle deprecation in containerized **add** and **merge** handlers.

When `deprecate_bundles_db` updates `index.db`, `merge_catalogs_dirs` only copies files from the migrated catalog (`copytree` with `dirs_exist_ok=True`) and does not remove stale `configs/<package>/` directories left in the Git clone. Deprecated operators removed from `index.db` could therefore remain in the Git-backed FBC committed to Konflux.

This PR extracts Git cleanup into a shared utility and invokes it from both handlers before `merge_catalogs_dirs`.

## Problem

Containerized workflow:

1. `deprecate_bundles_db` mutates `index.db`
2. `opm_migrate` produces `catalog_from_db` without deprecated operators
3. `merge_catalogs_dirs(catalog_from_db, localized_git_catalog_path)` merges into the Git catalog

Step 3 overwrites matching files but **does not delete** extra files/directories already present under `configs/<package>/` in Git. The final committed FBC could still contain deprecated operator content even though `index.db` was updated correctly.

The containerized **add** handler already worked around this with inline cleanup; the **merge** handler did not.

## Solution

- Add `remove_deprecated_operators_from_git_catalog()` in `iib/workers/tasks/utils.py`
- Call it from containerized **add** and **merge** handlers after `opm_migrate` and before `merge_catalogs_dirs`
- Drive Git removal from the same `deprecation_bundles` list passed to `deprecate_bundles_db` (including `invalid_bundles` in merge)
- Deduplicate `deprecation_bundles` in the merge handler before DB deprecation and Git cleanup

### Implementation notes

- Removes the entire `configs/<package>/` directory per deprecated bundle’s operator package (not individual bundle versions). This matches prior add-handler behavior and is required because `merge_catalogs_dirs` cannot prune stale version/channel files on its own. If the operator remains in `index.db`, `catalog_from_db` restores it on merge.
- Package names are validated against OLM DNS subdomain rules with path traversal checks before `shutil.rmtree`
- Unsafe or missing package labels are skipped with warning logs (fail-safe)

## Test plan

- [x] Unit tests for `remove_deprecated_operators_from_git_catalog` and package name validation (`test_utils.py`)
- [x] Updated containerized add handler tests
- [x] New merge handler tests: deprecation flow, deduplication with `invalid_bundles`, skip when no bundles deprecated
- [x] `tox -e py312`
- [x] `tox -m static`
- [x] Manual: containerized add/merge request with `deprecation_list` and verify Git catalog no longer contains removed operators

### DIFF
--- a/iib/workers/tasks/build_containerized_add.py
+++ b/iib/workers/tasks/build_containerized_add.py
@@ -45,11 +45,11 @@ from iib.workers.tasks.utils import (
     chmod_recursively,
     get_bundles_from_deprecation_list,
     get_resolved_bundles,
+    remove_deprecated_operators_from_git_catalog,
     request_logger,
     reset_docker_config,
     set_registry_token,
     RequestConfigAddRm,
-    get_image_label,
     verify_labels,
     prepare_request_for_build,
 )
@@ -243,22 +243,14 @@ def handle_containerized_add_request(
             generate_cache=False,
         )
 
-        # we have to remove all `deprecation_bundles` from `localized_git_catalog_path`
-        # before merging catalogs otherwise if catalog was deprecated and
-        # removed from `index.db` it stays on FBC (from_index)
-        # Therefore we have to remove the directory before merging
-        for deprecate_bundle_pull_spec in deprecation_bundles:
-            # remove deprecated operators from FBC stored in index image
-            deprecate_bundle_package = get_image_label(
-                deprecate_bundle_pull_spec, 'operators.operatorframework.io.bundle.package.v1'
+        # Remove stale operator directories from the from_index Git catalog before merging.
+        # merge_catalogs_dirs only copies from catalog_from_db; without this step, operators
+        # removed from index.db by deprecate_bundles_db would remain under configs/<package>/.
+        # Use the same bundle list as index.db deprecation.
+        if deprecation_bundles:
+            remove_deprecated_operators_from_git_catalog(
+                localized_git_catalog_path, deprecation_bundles
             )
-            bundle_from_index = Path(localized_git_catalog_path) / deprecate_bundle_package
-            if bundle_from_index.is_dir():
-                log.debug(
-                    "Removing deprecated bundle from catalog before merging: %s",
-                    deprecate_bundle_package,
-                )
-                shutil.rmtree(bundle_from_index)
         # overwrite data in `localized_git_catalog_path` by data from `catalog_from_db`
         # this adds changes on not opted in operators to final
         merge_catalogs_dirs(catalog_from_db, localized_git_catalog_path)
--- a/iib/workers/tasks/build_containerized_merge.py
+++ b/iib/workers/tasks/build_containerized_merge.py
@@ -46,6 +46,7 @@ from iib.workers.tasks.utils import (
     RequestConfigMerge,
     set_registry_token,
     get_bundles_from_deprecation_list,
+    remove_deprecated_operators_from_git_catalog,
 )
 from iib.workers.tasks.fbc_utils import merge_catalogs_dirs
 from iib.workers.tasks.iib_static_types import BundleImage
@@ -243,6 +244,7 @@ def handle_containerized_merge_request(
         deprecation_bundles = deprecation_bundles + [
             bundle['bundlePath'] for bundle in invalid_bundles
         ]
+        deprecation_bundles = list(dict.fromkeys(deprecation_bundles))
 
         # process the deprecation list into the intermediary index.db file
         if deprecation_bundles:
@@ -273,6 +275,16 @@ def handle_containerized_merge_request(
         # final destination of catalog (defined in Dockerfile)
         catalog_from_db = os.path.join(temp_dir, 'from_db')
         os.rename(fbc_dir, catalog_from_db)
+
+        # Remove stale operator directories from the target Git catalog before merging.
+        # This complements deprecate_bundles_db: merge_catalogs_dirs only copies from
+        # catalog_from_db and would leave existing configs/<package>/ dirs untouched, or
+        # re-add packages that were removed from index.db only when deprecate_bundles_db ran.
+        # Use the same bundle list as index.db deprecation (includes invalid_bundles).
+        if deprecation_bundles:
+            remove_deprecated_operators_from_git_catalog(
+                localized_git_catalog_path, deprecation_bundles
+            )
 
         # Merge migrated FBC with existing FBC in Git repo
         # overwrite data in `catalog_from_index` by data from `catalog_from_db`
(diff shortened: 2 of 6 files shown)


iib/workers/tasks/utils.py
```diff
@@ -455,6 +556,49 @@ def get_bundles_from_deprecation_list(bundles: List[str], deprecation_list: List
...
+    """
+    for operator_package in get_operator_packages_from_deprecation_list(deprecation_list):
+        operator_catalog_dir = Path(localized_git_catalog_path) / operator_package
+        if operator_catalog_dir.is_dir():
+            log.debug(
+                'Removing deprecated operator from catalog before merging: %s',
+                operator_package,
+            )
+            shutil.rmtree(operator_catalog_dir)
+
```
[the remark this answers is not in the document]
lipoja: Fixed

iib/workers/tasks/utils.py
```diff
@@ -455,6 +556,49 @@ def get_bundles_from_deprecation_list(bundles: List[str], deprecation_list: List
...
+        )
+        if operator_package:
+            operator_packages.append(operator_package)
+        else:
+            log.warning(
+                'Could not determine operator package name for deprecated bundle %s. '
+                'The operator directory will not be removed from the Git catalog.',
+                pull_spec,
+            )
+    return list(dict.fromkeys(operator_packages))
```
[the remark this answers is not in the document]
lipoja: I think I am fine whit IIB failing on this step.

iib/workers/tasks/utils.py
```diff
@@ -430,6 +431,92 @@ class RequestConfigAddDeprecations(RequestConfig):
...
+    :rtype: bool
+    """
+    if bundle in resolved_deprecation_pull_specs:
+        return True
+
+    if '@sha256:' not in bundle:
+        return False
+
+    bundle_digests = _get_bundle_image_digests(bundle)
+    return bool(bundle_digests & deprecation_digests)
```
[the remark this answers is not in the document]
lipoja: Partially fixed

iib/workers/tasks/utils.py
```diff
@@ -430,6 +431,109 @@ class RequestConfigAddDeprecations(RequestConfig):
...
+    :rtype: str | None
+    """
+    if '@sha256:' not in pull_spec:
+        return None
+    return _normalize_digest(pull_spec.rsplit('@sha256:', 1)[1])
+
+
+def _get_bundle_image_digests(pull_spec: str) -> Set[str]:
+    """
+    Get all digests that identify the same bundle image.
```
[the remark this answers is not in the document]
lipoja: Fixed

iib/workers/tasks/utils.py
```diff
@@ -430,6 +431,109 @@ class RequestConfigAddDeprecations(RequestConfig):
...
+    :param str pull_spec: the pull specification of the container image
+    :return: the normalized digest or ``None`` if the pull spec is not digest-pinned
+    :rtype: str | None
+    """
+    if '@sha256:' not in pull_spec:
+        return None
+    return _normalize_digest(pull_spec.rsplit('@sha256:', 1)[1])
+
+
+def _get_bundle_image_digests(pull_spec: str) -> Set[str]:
```
yashvardhannanavati: I don't think we need this commit at all. All bundles stored in the index.db go through the exact same get_resolved_bundles function, so matching would always work.

yashvardhannanavati: I think the fix goes way beyond the intended scope of CLOUDDST-32845. Consider sticking to the original scope.

lipoja: This is present because otherwise our iib-api-tests were failing.

lipoja: It was because of incorrect testing data.
I removed the functions from the code.

iib/workers/tasks/build_containerized_add.py
```diff
@@ -243,22 +243,13 @@ def handle_containerized_add_request(
...
-        # removed from `index.db` it stays on FBC (from_index)
-        # Therefore we have to remove the directory before merging
-        for deprecate_bundle_pull_spec in deprecation_bundles:
-            # remove deprecated operators from FBC stored in index image
-            deprecate_bundle_package = get_image_label(
-                deprecate_bundle_pull_spec, 'operators.operatorframework.io.bundle.package.v1'
+        # Remove stale operator directories from the from_index Git catalog before merging.
+        # merge_catalogs_dirs only copies from catalog_from_db; without this step, operators
+        # removed from index.db by deprecate_bundles_db would remain under configs/<package>/.
+        if deprecation_list:
```
yashvardhannanavati: should this check `if deprecation_bundles` instead?

iib/workers/tasks/build_containerized_merge.py
```diff
@@ -274,6 +276,15 @@ def handle_containerized_merge_request(
         catalog_from_db = os.path.join(temp_dir, 'from_db')
         os.rename(fbc_dir, catalog_from_db)
 
+        # Remove stale operator directories from the target Git catalog before merging.
+        # This complements deprecate_bundles_db: merge_catalogs_dirs only copies from
+        # catalog_from_db and would leave existing configs/<package>/ dirs untouched, or
+        # re-add packages that were removed from index.db only when deprecate_bundles_db ran.
+        if deprecation_list:
```
yashvardhannanavati: same question as above

iib/workers/tasks/build_containerized_add.py
lipoja: thanks, fixed

iib/workers/tasks/build_containerized_merge.py
lipoja: thanks, fixed

tests/test_workers/test_tasks/test_utils.py
```diff
@@ -992,6 +993,94 @@ def testget_bundles_from_deprecation_list(mock_grb):
...
+
+    assert packages == []
+    mock_log.warning.assert_called_once()
+    assert pull_spec in mock_log.warning.call_args[0][1]
+
+
+@mock.patch('iib.workers.tasks.utils.get_operator_packages_from_deprecation_list')
+@mock.patch('iib.workers.tasks.utils.shutil.rmtree')
+def test_remove_deprecated_operators_from_git_catalog(mock_rmtree, mock_get_packages, tmpdir):
+    catalog_path = tmpdir.mkdir('configs')
```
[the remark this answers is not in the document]
lipoja: test added

yashvardhannanavati (approved):

PR #1370, opened 2026-08-03
Fix set_registry_token call for token usa

chandwanitulsi:
Assisted-by: Cursor/Gemini

### DIFF
--- a/iib/workers/tasks/build_fbc_operations.py
+++ b/iib/workers/tasks/build_fbc_operations.py
@@ -18,6 +18,7 @@ from iib.workers.tasks.build import (
 from iib.workers.tasks.celery import app
 from iib.workers.tasks.opm_operations import opm_registry_add_fbc_fragment, Opm
 from iib.workers.tasks.utils import (
+    get_images_needing_overwrite_token,
     get_resolved_image,
     prepare_request_for_build,
     request_logger,
@@ -69,12 +70,23 @@ def handle_fbc_operation_request(
     _cleanup()
     set_request_state(request_id, 'in_progress', 'Resolving the fbc fragments')
 
-    # Resolve all fbc fragments
+    # Apply overwrite_from_index_token only to same-namespace fragments that are not already
+    # covered by worker Docker config credentials. from_index is not pulled here; prepare_request
+    # and later from_index accessors apply the overwrite token for the index itself.
+    fragments_needing_token = get_images_needing_overwrite_token(from_index, fbc_fragments)
     resolved_fbc_fragments = []
-    for fbc_fragment in fbc_fragments:
-        with set_registry_token(overwrite_from_index_token, fbc_fragment, append=True):
-            resolved_fbc_fragment = get_resolved_image(fbc_fragment)
-            resolved_fbc_fragments.append(resolved_fbc_fragment)
+
+    def _resolve_fbc_fragments() -> None:
+        for fbc_fragment in fbc_fragments:
+            resolved_fbc_fragments.append(get_resolved_image(fbc_fragment))
+
+    if fragments_needing_token:
+        with set_registry_token(
+            overwrite_from_index_token, fragments_needing_token, append=True
+        ):
+            _resolve_fbc_fragments()
+    else:
+        _resolve_fbc_fragments()
 
     prebuild_info = prepare_request_for_build(
         request_id,
--- a/iib/workers/tasks/build_fbc_operations.py
+++ b/iib/workers/tasks/build_fbc_operations.py
@@ -18,6 +18,7 @@ from iib.workers.tasks.build import (
 from iib.workers.tasks.celery import app
 from iib.workers.tasks.opm_operations import opm_registry_add_fbc_fragment, Opm
 from iib.workers.tasks.utils import (
+    get_images_needing_overwrite_token,
     get_resolved_image,
     prepare_request_for_build,
     request_logger,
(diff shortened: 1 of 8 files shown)


iib/workers/tasks/build_fbc_operations.py
```diff
@@ -69,12 +78,17 @@ def handle_fbc_operation_request(
...
     set_request_state(request_id, 'in_progress', 'Resolving the fbc fragments')
 
-    # Resolve all fbc fragments
+    # Resolve all fbc fragments. Only apply overwrite_from_index_token when the fragment
+    # lives in the same repository as from_index; otherwise the token is for a different
+    # repo and must not be used for fragment auth.
     resolved_fbc_fragments = []
     for fbc_fragment in fbc_fragments:
-        with set_registry_token(overwrite_from_index_token, fbc_fragment, append=True):
+        if from_index and _images_in_same_repository(fbc_fragment, from_index):
```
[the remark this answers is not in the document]
lipoja: Yeah documenting it might be nice.

lipoja (dismissed):
Once you address the [medium] logic-error you can merge it.
LGTM.

iib/workers/tasks/utils.py
```diff
@@ -599,40 +599,170 @@ def _docker_auth_key_for_image(container_image: str) -> str:
...
+            continue
+        try:
+            with open(path, 'r') as f:
+                return json.load(f).get('auths', {}) or {}
+        except (OSError, json.JSONDecodeError) as e:
+            log.warning('Failed to read Docker config auths from %s: %s', path, e)
+    return {}
+
+
+def docker_config_has_auth_for_image(
```
lipoja: This one is never used in IIB code - only in tests, can we remove it if it is not needed?

lipoja (changes requested):

lipoja:
LGTM, lets fix the linting errors and remove the unused code.

MichalZelenak:
Apart from the comments above LGTM. 

Once we migrate to the Pods, would it make sense, to craft the credentials file (docker config) per command? That way, we would have better control over the credentials.

jfrancin:
Hello - wondering how long before this gets merged? I've been waiting for this to be done so I can once again do 'manual' IIB job submissions (I have the relevent credentials) to remove or add bundles in the Certified, Marketplace, or Community bundle indexes.

iib/workers/tasks/opm_operations.py
```diff
@@ -1020,22 +1038,42 @@ def opm_registry_add_fbc_fragment(
...
-        )
-        fragment_data.append((fragment_path, fragment_operators))
-        all_fragment_operators.extend(fragment_operators)
+    fragments_needing_token = get_images_needing_overwrite_token(from_index, fbc_fragments)
+
+    if fragments_needing_token:
+        with set_registry_token(
+            overwrite_from_index_token, fragments_needing_token, append=True
+        ):
+            for i, fbc_fragment in enumerate(fbc_fragments):
```
yashvardhannanavati: nit: repeated code. Consider creating a helper function.

iib/workers/tasks/utils.py
```diff
@@ -599,40 +599,170 @@ def _docker_auth_key_for_image(container_image: str) -> str:
...
+    * ``registry/namespace/repo`` when the image includes a namespace
     * ``registry`` when the image has no namespace (for example, ``localhost:5000/myimage:tag``)
 
-    If the pull specification does not contain a resolvable registry, :exc:`IIBError` is raised.
+    If a pull specification does not contain a resolvable registry, :exc:`IIBError` is raised.
 
     Broader credentials already present in the Docker configuration, such as registry- or
     namespace-level entries for the same host, are not modified. Only the scoped ``auths`` key
-    derived from ``container_image`` is set or overwritten.
+    derived from each image is set or overwritten. When the worker config has no
```
yashvardhannanavati, line 791: The namespace-level fallback in set_registry_token applies to all callers, not just the resolve/opm paths this PR targets. When the overwrite token is repo-scoped (not namespace-wide), stamping the shared registry/namespace key can cause same-namespace images that were previously pulled anonymously or via a registry-level template cred to fail with 401/403. Is this an acceptable risk?

iib/workers/tasks/utils.py
```diff
@@ -599,40 +599,170 @@ def _docker_auth_key_for_image(container_image: str) -> str:
...
+
+    keys: List[str] = []
+    if image_name.namespace:
+        keys.append(f'{registry}/{image_name.namespace}/{image_name.repo}')
+        keys.append(f'{registry}/{image_name.namespace}')
+    keys.append(registry)
+    return keys
+
+
+def _load_docker_config_auths() -> Dict[str, Any]:
```
yashvardhannanavati, line 678: The capability is needed, but it introduces a third, differently-behaving config reader alongside the append block (reads ~/.docker/config.json, raises on bad JSON) and set_registry_auths (reads the template only):

- It prefers ~/.docker/config.json over the template, but the credentials actually applied come from the template. These agree only when config is the post-reset_docker_config symlink; with a leftover/nested real config, the "already covered" decision is made against a different source than what gets merged. For a "does the worker have baseline creds" check, the template is the source of truth.
- It swallows OSError/JSONDecodeError and returns {}, which makes token stamping broader on a read error (empty auths ⇒ everything looks uncovered) — the opposite of this PR's intent, and inconsistent with the append block that raises on the same error.

yashvardhannanavati (changes requested):
@chandwanitulsi could you please fix the tests too?

iib/workers/tasks/utils.py
chandwanitulsi: Always reads the reset template: _cleanup() runs before the first check, and the deeper checks run outside any token context (which restores the template in its finally).

Fixed. _load_docker_config_auths now raises on a corrupt config instead of returning {} (which would have broadened stamping). Tests added.

iib/workers/tasks/utils.py
chandwanitulsi: some pull paths (e.g. opm) don't match repo-scoped keys, so the repo key alone fails. Bounded to same-namespace, no existing path-auth, namespace-key-only. The anonymous same-namespace 401/403 case is a known, accepted trade-off for the index namespace.

iib/workers/tasks/utils.py
chandwanitulsi: Fixed

iib/workers/tasks/utils.py
```diff
@@ -1097,7 +1257,7 @@ def get_index_image_info(
     if not from_index:
```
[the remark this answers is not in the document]
chandwanitulsi: Fixed

iib/workers/tasks/utils.py
```diff
@@ -721,8 +841,39 @@ def set_registry_token(
 
                 log.debug('Docker config will be updated')
 
-    log.debug('Setting the override token for the image %s', auth_key)
-    registry_auths['auths'].update({auth_key: auth_entry})
+    # Snapshot worker/template auths before we add overwrite keys. Used to decide whether a
+    # namespace-level fallback is needed when no path-scoped credentials exist yet.
+    existing_auths = _load_docker_config_auths()
```
yashvardhannanavati: @chandwanitulsi  This change makes lines 831-840 irrelevant because irrespective of the `append` flag value, existing auth is loaded from the template. This will be a very invasive change. I am ok with letting it slide for the sake of this PR but it should definitely be looked at and fixed

yashvardhannanavati (approved):

yashvardhannanavati (approved):

lipoja (approved):

